Issue #271, opened 2014-12-05
Custom colors for scatters


JosteinTopland:
Hi,

I'd like to set custom colors to scatters. For my case, the scatters are on top of a heat map. When I set value to a scatter, the color is set from the value of the heat map palette color. If value is not set on the scatter, the scatter is not shown.

How can I set custom colors to scatters?

The following code reproduces the issue:

	Plot.Model = new PlotModel();

	var colorAxis = new LinearColorAxis
	{
		Position = AxisPosition.Right,
		Palette = OxyPalettes.Hot(100)
	};
	Plot.Model.Axes.Add(colorAxis);

	var heatMap = new HeatMapSeries
	{
		X0 = -1,
		X1 = 1,
		Y0 = -1,
		Y1 = 1,
		Data = new double[,] {
			{0, 1, 2},
			{3, 4, 5},
			{6, 7, 8}
		}
	};
	Plot.Model.Series.Add(heatMap);

	var scatter = new ScatterSeries
	{
		MarkerStroke = OxyColors.Red,
		MarkerStrokeThickness = 10,

		// ISSUE: Scatter does not turn blue. <-----
		MarkerFill = OxyColors.Blue
	};

	// ISSUE: If 4th argument is removed, the scatter is not shown. <-----
	scatter.Points.Add(new ScatterPoint(0, 0, 20, 0));
	Plot.Model.Series.Add(scatter);

Screenshot:

![scatter](https://cloud.githubusercontent.com/assets/5829968/5313591/8c7b0184-7c6a-11e4-9038-f37822c005c2.png)

harika-strand:
How to give different colors to the scatter points (Colors to be custom configurable).
Suppose if i mention lowcolor and high color i can able to color the plot in two colors. if i want to show the plot points in 3 colors. then wt will i do ??

PR #307, opened 2014-12-30
Custom colors for scatters

JosteinTopland:
If MarkerFill color is set (i.e. not default OxyColors.Automatic), then fill scatters with MarkerFill.

https://github.com/oxyplot/oxyplot/issues/271


objorke:
Is this covered in an example? I was searching for #271 in `ExampleLibrary.Issues` but could not find any.

PR #316, opened 2015-01-05
Fix for possible NullReferenceException

GeertvanHorrik:


GeertvanHorrik:
Please discuss the code in the PR (Files changed tab) instead of the direct code tab, then it's easier to maintain.

Source/OxyPlot/Annotations/PathAnnotation.cs
```diff
@@ -386,11 +386,14 @@ public override void Render(IRenderContext rc, PlotModel model)
         /// </returns>
         protected override HitTestResult HitTestOverride(HitTestArguments args)
         {
-            var nearestPoint = ScreenPointHelper.FindNearestPointOnPolyline(args.Point, this.screenPoints);
-            double dist = (args.Point - nearestPoint).Length;
-            if (dist < args.Tolerance)
+            if (this.screenPoints != null)
```
objorke: invert `if` statement to reduce nesting (Resharper recommends this, and I agree)

GeertvanHorrik: That means 2 return null, but ok

Source/OxyPlot/Rendering/ScreenPointHelper.cs
```diff
@@ -22,11 +24,17 @@ public static class ScreenPointHelper
         /// <param name="point">The point.</param>
         /// <param name="points">The points.</param>
         /// <returns>The nearest point.</returns>
+        /// <exception cref="ArgumentNullException">The <paramref name="points"/> is <c>null</c>.</exception>
         public static ScreenPoint FindNearestPointOnPolyline(ScreenPoint point, IList<ScreenPoint> points)
         {
             double minimumDistance = double.MaxValue;
             var nearestPoint = default(ScreenPoint);
 
+            if (points == null)
```
objorke: isn't it normal to do the argument checks first? e.g. move this block to the top of the method?

objorke: sorry for commenting on the wrong place, I am still learning... thanks for the correction! :-)

Source/OxyPlot/Annotations/PathAnnotation.cs
GeertvanHorrik: Resharper says this. is redundant, and I agree ;-)

objorke: right, but isn't nesting worse?

Issue #121, opened 2014-08-14
Create two-color AreaSeries


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10150](http://oxyplot.codeplex.com/workitem/10150)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-03-07 15:23
See 

<a href="https://oxyplot.codeplex.com/discussions/538255" rel="nofollow">https://oxyplot.codeplex.com/discussions/538255</a>

PR #318, opened 2015-01-05
Feature two color area series wpf wrapper

piotrwarzocha:
So as I said, I've added a wpf wrapper for Two-color area series and an example of that wpf wrapper in WpfExamles.

Issue #121


objorke:
Great! But `OxyPlot.Wpf.csproj`(NET45) is not updated? Please keep all in sync!

piotrwarzocha:
Ok I've updated those files.

Issue #325, opened 2015-01-13
The specified Visual is not an ancestor of this Visual.


kevkon3:
I am trying out some WPF docking code (see http://www.actiprosoftware.com/products/controls/wpf/docking) and I am attempting to put a PlotView into a docking container. Everything works fine until I try to move the docking window to a new place.

I get a message of "The specified Visual is not an ancestor of this Visual." and stacktrace (copied below).

I took a look at IsUserVisible(FrameworkElement element, FrameworkElement container) in PlotBase. It checks to see if the element (I assume the plot) is visible, but not the container (I assume the docking window). My guess is that the docking windows briefly becomes invisible as it is moved to a new location. Perhaps adding a check to see if the container is visible would fix the issue?
   
   at System.Windows.Media.Visual.TrySimpleTransformToAncestor(Visual ancestor, Boolean inverse, GeneralTransform& generalTransform, Matrix& simpleTransform)
   at System.Windows.Media.Visual.InternalTransformToAncestor(Visual ancestor, Boolean inverse)
   at OxyPlot.Wpf.PlotBase.IsUserVisible(FrameworkElement element, FrameworkElement container) in c:\projects\oxyplot\Source\OxyPlot.Wpf\PlotBase.cs:line 450
   at OxyPlot.Wpf.PlotBase.IsVisibleToUser() in c:\projects\oxyplot\Source\OxyPlot.Wpf\PlotBase.cs:line 427
   at OxyPlot.Wpf.PlotBase.OnLayoutUpdated(Object sender, EventArgs e) in c:\projects\oxyplot\Source\OxyPlot.Wpf\PlotBase.cs:line 493
   at System.Windows.ContextLayoutManager.fireLayoutUpdateEvent()
   at System.Windows.ContextLayoutManager.UpdateLayout()
   at System.Windows.Interop.HwndSource.SetLayoutSize()
   at System.Windows.Interop.HwndSource.set_RootVisualInternal(Visual value)
   at System.Windows.Window.SetRootVisualAndUpdateSTC()
   at System.Windows.Window.SetupInitialState(Double requestedTop, Double requestedLeft, Double requestedWidth, Double requestedHeight)
   at System.Windows.Window.CreateSourceWindow(Boolean duringShow)
   at System.Windows.Window.ShowHelper(Object booleanBox)
[...]

PR #326, opened 2015-01-13
Issue #325 When checking to see if an element is visible, also check to see if the container is visible

kevkon3:


objorke:
Good! Thanks for the contribution!

objorke:
#325

Issue #331, opened 2015-01-19
Invalid TwoColorArea series rendering when the number of points is 1.


piotrwarzocha:
This is a small bug. 

When I plotted a TwoColorArea series with the number of points > 1 and althought I changed series points, which quantity was 1, the limit line wasn't updated.

PR #332, opened 2015-01-19
Fixed updating points for the limit line when actualPoints count is 1.

piotrwarzocha:
Issue #331 

Clearing limit line points list also when its number is 1.


objorke:
ps - can you add an example so we can test that the issue is fixed! :-)

PR #334, opened 2015-01-21
Xwt Tracker rewrite + various fixes

sevoku:
The Popover widget I've used for the tracker does not really fit the requirements. Now the tracker is simply drawn like everything else with a major performance increase (when enabled) and more tracker settings.

Additionally some smaller fixes are included.


objorke:
Good! My only comment is that it would be nice to have the tracker implementation in a separate object (including settings) instead of in a virtual method in the `PlotView`. I prefer composition over inheritance. The trackers in Silverlight and WPF are implemented this way!

sevoku:
Sure, this will be the next step :+1:

Issue #336, opened 2015-01-22
iOS multitouch pan/zoom is glitchy and slow


TheAlmightyBob:
1. The current touch handling logic can mix up fingers during a zoom gesture, leading to some unexpected jumpy behavior. For example, it will switch back and forth between single-touch and multitouch modes.
2. The KeepFixedAspectRatio mode is a bit unstable if trying to only zoom along one axis.
3. The second finger (i.e., starting a zoom gesture) creates an unnecessary second TouchManipulator, which, among other possible side-effects, means that the graph is being invalidated twice as often as it needs to be.

objorke:
Can this be closed?

TheAlmightyBob:
Yep! I only created it to accompany the pull request. :)

PR #337, opened 2015-01-22
Issue #336: Fixed multitouch instability

TheAlmightyBob:
Was mixing up fingers, switching modes mid-gesture, and creating a duplicate touch manipulator.

Also added a threshold to make it easier to zoom in only the X or Y direction without the graph going crazy (because it thought you were trying to zoom to/from near zero along the other axis, instead of trying not to zoom at all). Threshold is currently a constant... it could be a configurable property, but then might need a better name.

(also updated the Xamarin.iOS\ExampleBrowser deployment target as required by Xamarin Unified?)

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -18,6 +18,7 @@ Benoit Blanchon <>
 br
 brantheman
 Brannon King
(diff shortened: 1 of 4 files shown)


objorke:
Great! Thanks for contributing on iOS!

Issue #338, opened 2015-01-23
PlotModel.Invalidate is not hooked up on iOS/Android (including Forms)


TheAlmightyBob:
The other platforms have OnModelChanged logic that gets executed when the model is set on the PlotView (in particular, AttachPlotView), but this is missing from iOS/Android. Thus, calling Invalidate on the model has no effect. This can be observed in some of the examples that take advantage of this (e.g. Touch).

(this is intended to replace issue #288, which just describes one symptom of this issue)

objorke:
can this be closed now?

TheAlmightyBob:
It might be wise for somebody else to verify the fix on Android first, since I don't have a license for that.

tibel:
Please verify with latest OxyPlot version. Thanks.

According to our guidelines https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTING.md we will close this issue. Feel free to re-open, when you have new information.

Issue #288, opened 2014-12-11
PlotView not exposing Invalidate


david-laundav:
As mentioned in [this issue on Stackoverflow](http://stackoverflow.com/questions/27426584/fill-color-of-a-single-point-in-a-scatterseries), when I create a `ScatterSeries` and listen to the `TouchStarted` event to highlight the selected item, it is not updated immediately in the `PlotView`. I have to drag the canvas before the `SelectionColor` is applied.

First, in my ViewModel, I do like so:
```C#
var model = new PlotModel()
{  
SelectionColor = OxyColors.Red,
// ...
};

var scatterSeries = new ScatterSeries()
{
    SelectionMode = SelectionMode.Single,
    // ...
};

var scatterPoints = new List<ScatterPoint>
{
    new ScatterPoint(0, 4),
    // ...
};
```

Thereafter, I listen to the TouchStarted event of the ScatterSeries:
```
scatterSeries.TouchStarted += (sender, e) =>
{
    var xCoordinate = e.Position.X;
    var yCoordinate = e.Position.Y;

    var screenPoint = new ScreenPoint(xCoordinate, yCoordinate);

    var point = lineSeries.GetNearestPoint(screenPoint, false);
    var index = (int)point.Index;
    scatterSeries.SelectItem(index);

    // I expect it to refresh and update the PlotModel here:
    model.InvalidatePlot(true); 

    e.Handled = true;
};
```
In my ViewModel I have also created a PlotModel Property which my PlotView (in my View) is bound to (i.e. PlotView.ModelProperty). 

As mentioned, this seems to work fine, except for that the `PlotModel` is not updated immediately with the `SelectionColor` - I have to drag the canvas before it updates.

I think this may have something to do with the fact, that my PlotView is not invalidated after data has been updated. 

[As Cheesebaron mentions](http://stackoverflow.com/a/27427305/2399772), I am unable to Invalidate the PlotView since it is not exposed in Xamarin.Forms.

Is this something you could possibly include, or do you have any other solution?

objorke:
is this solved now?

david-laundav:
Tested on Android which works. Should be tested on iOS as well, as soon as #348 is fixed.

objorke:
I think this is a duplicate of #338

david-laundav:
It is a duplicate of #338 indeed. Closing.

PR #339, opened 2015-01-23
Issue #338: Attaching/detaching PlotView/Model in Model setter.

TheAlmightyBob:
Added to iOS/Android the same logic as other platforms have in OnModelChanged.
(kept to setters since the separate OnModelChanged isn't required here like it is on the DependencyProperty-based implementations, but it could easily be refactored for consistency)

Should also address issue #288.

Warning: I do not have an Android license, so am unable to test that. I am merely trusting that it works there since the same logic is tested on all the other platforms and isn't really platform-specific. (and I have seen this issue discussed in Xamarin's Android forums)


objorke:
Thanks!

PR #352, opened 2015-01-29
Update PlotView namespace on Android Layout due to namespace modification 

ze-pequeno:
Android samples do not work anymore due to namespace change


objorke:
thanks!

Issue #345, opened 2015-01-25
crash on touch event in windows phone 8.1 


jiongyang:
The SetCursorType method in PlotView.cs throws "not implemented exception" upon touch event. Can be fixed by the following code: 

```csharp
        public void SetCursorType(CursorType cursor)
        {

            // this is not needed for windows phone apps

#if WINDOWS_APP

            var type = CoreCursorType.Arrow;
            switch (cursor)
            {
                case CursorType.Default:
                    type = CoreCursorType.Arrow;
                    break;
                case CursorType.Pan:
                    type = CoreCursorType.Hand;
                    break;
                case CursorType.ZoomHorizontal:
                    type = CoreCursorType.SizeWestEast;
                    break;
                case CursorType.ZoomVertical:
                    type = CoreCursorType.SizeNorthSouth;
                    break;
                case CursorType.ZoomRectangle:
                    type = CoreCursorType.SizeNorthwestSoutheast;
                    break;
            }

            Window.Current.CoreWindow.PointerCursor = new CoreCursor(type, 1);

#endif

        }
```

objorke:
@jiongyang I don't think we can use compiler directives in this library that is used for both windows store apps and windows phone apps. Is there a way to determine if the `PointerCursor` can be set? Can move it into a try-catch block, but is there a better way?

jiongyang:
Hi @objorke, thanks for your reply. I haven't find a principled way to determine if the CoreCursor can be created at run time. A temporal workaround may be to use "Window.Current.CoreWindow.PointerPosition" to determine if a cursor is available. In my case, It is set to large negative value and does not point to an actual position.

objorke:
I found that the `PointerPosition` getter threw a "catastrophic exception" in some cases. Wrapping the `CoreCursor` constructor in a try-catch block seemed to work, but I have marked it as a "workaround". It would be great to see a nicer solution!

Currently there seems to be a WP8 build problem on AppVeyor, I probably need to create a new topic in their support forum.

objorke:
Can this be closed?

jiongyang:
Sure. Thanks!

objorke:
Waiting for updated NuGet packages, this fix should be included in the next successful build.

http://help.appveyor.com/discussions/questions/921-support-for-silverlight-4-available

PR #353, opened 2015-02-01
Handle Windows Phone set cursor exception

objorke:
This solves #345, but I guess there is a better solution...


objorke:
created new PR

Issue #356, opened 2015-02-03
Draw legend line with custom pattern


gotoh:
It seems that legend draws line with LineStyle although Dashes is set.
The following patch fix it.

```diff
diff --git a/Source/OxyPlot/Series/LineSeries.cs b/Source/OxyPlot/Series/LineSeries.cs
index 296ee98..dfed3fa 100644
--- a/Source/OxyPlot/Series/LineSeries.cs
+++ b/Source/OxyPlot/Series/LineSeries.cs
@@ -365,7 +365,7 @@ public override void RenderLegend(IRenderContext rc, OxyRect legendBox)
                 pts,
                 this.GetSelectableColor(this.ActualColor),
                 this.StrokeThickness,
-                this.ActualLineStyle.GetDashArray());
+                this.ActualDashArray);
             var midpt = new ScreenPoint(xmid, ymid);
             rc.DrawMarker(
                 legendBox,
```

objorke:
Great! That's an easy fix. But we need a pull request! (please include or refer to an example, do the fix and update the CHANGELOG.md, and possibly the CONTRIBUTORS/AUTHORS file!)

PR #357, opened 2015-02-03
Draw legend line with custom pattern #356

gotoh:
By this fix, LineSeries draw the line in legend with specified pattern in Dashes.
Before this, Dashes is not used in legend at all.

A sample is also added to see the change.


objorke:
This looks good!

objorke:
The build fails due to an error at AppVeyor (http://help.appveyor.com/discussions/questions/921-support-for-silverlight-4-available). I'll merge this PR, but we need to wait for new NuGet packages to be published until the AppVeyor has been fixed.

Issue #340, opened 2015-01-23
Add "tap" support for touch-enabled platforms


TheAlmightyBob:
[this time, I do not have a corresponding pull request ready to go. sorry.]

The current TouchStarted/TouchDelta/TouchCompleted pattern is a good start that works for pan/zoom and some other scenarios. However, it's difficult for an app to enable pan/zoom *and* try to handle those events to detect simple taps (e.g. to select lines/annotations). I believe the platforms each already have support for capturing taps separately, so this should mainly be a matter of piping that through the core library.
- Update: Those events actually could not be used to detect taps on Windows, and can no longer be used to detect taps on iOS. They really mean "pan/zoom manipulation started," not touch in general.

~~The current situation is particularly complicated on iOS (where I've spent the most time). Even after my recent fixes, the event sequence can be a little.... *unexpected*... to app developers. In handling the scenario of the user changing fingers mid-gesture, the logic may "restart" the touch handling.~~

~~For example:
Put one finger down
Put a second finger down
Lift the first finger~~

~~Results in:
TouchStarted (first finger)
TouchCompleted(first finger)
TouchStarted (second finger)~~

~~Possibly that is an issue in itself and could use work, but the point in this context is that it makes it that much more difficult for an app developer to try to detect a tap.~~

~~In the iOS scenario, this may benefit from refactoring the current logic into a custom gesture recognizer (to then be used in conjunction with a TapGestureRecognizer).~~

[above issues were resolved in #359. Pan/Zoom currently uses an iOS gesture recognizer and no longer generates extra Started/Completed events]

P.S. It might also be nice to have the ability to forward touches to the mouse events, particularly for Windows, but maybe there's a reason that hasn't been done before...

objorke:
Good ideas here! The original idea was to implement support for the 'low level' tap events (the current exposed api) in a similar way on all the touch-capable platforms, then hopefully create a more developer friendly gesture recognition system on top of it. I think such a system could be done in portable code!

It would be great if someone could look into this - do some refactoring of the current code or come up with some better ideas!

TheAlmightyBob:
If that's the goal, then it seems like the current API would need to be even lower-level... e.g., information about specific touches instead of just pan/scale properties. Which may mean abandoning some of the platform-specific helpers (such as the Windows manipulation events).

I certainly understand the value of doing as much as possible in PCL, and I do think it's possible, but it seems like a lot of refactoring and extra work vs. wrapping the existing platform-specific gesture support.

[...]

PR #359, opened 2015-02-08
iOS gesture recognizer and zoom options

TheAlmightyBob:
Related to, but not an actual fix for, issue #340.

- Fixes the issue I mentioned with extra Started/Completed events being generated mid-gesture. This involves a change to TouchManipulator that affects other platforms, so I again must warn that I am unable to test with Android.
- Refactors the iOS touch processing out of PlotView and into a custom gesture recognizer (this was the original intent of this change, the above issue just got fixed along the way).
- Adds properties to set the zoom threshold and "AllowPinchPastZero" (open to ideas for a better name there) behavior. These only apply if KeepAspectRatioWhenPinching is false, and they each default to the previous behavior. The intent of both options is to make it harder for users to accidentally zoom.

### DIFF
--- a/Source/OxyPlot.MonoTouch/PanZoomGestureRecognizer.cs
+++ b/Source/OxyPlot.MonoTouch/PanZoomGestureRecognizer.cs
@@ -1,0 +1,192 @@
+// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="PanZoomGestureRecognizer.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Recognizes drag/pinch multitouch gestures and translates them into pan/zoom information.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.MonoTouch
+{
+    using global::MonoTouch.Foundation;
+    using global::MonoTouch.UIKit;
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+
+    /// <summary>
+    /// Recognizes drag/pinch multitouch gestures and translates them into pan/zoom information.
+    /// </summary>
+    public class PanZoomGestureRecognizer : UIGestureRecognizer
+    {
+        /// <summary>
+        /// Up to 2 touches being currently tracked in a pan/zoom.
+        /// </summary>
+        private List<UITouch> activeTouches = new List<UITouch>();
+
+        /// <summary>
+        /// How far apart touch points must be on a certain axis to enable scaling that axis.
+        /// (only applies if KeepAspectRatioWhenPinching == false)
+        /// </summary>
+        public double ZoomThreshold { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether this <see cref="OxyPlot.Xamarin.iOS.PlotView"/> keeps the aspect ratio when pinching.
+        /// </summary>
+        /// <value><c>true</c> if keep aspect ratio when pinching; otherwise, <c>false</c>.</value>
+        public bool KeepAspectRatioWhenPinching { get; set; }
+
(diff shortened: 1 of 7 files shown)


objorke:
Looks good. I'll closer at this tomorrow! Could the MonoTouch and Unified API gesture recognizers share code (with some conditional #ifs) or is it better to keep it as it is now?

TheAlmightyBob:
Seems like that's a general question about the MonoTouch/Unified projects. I don't think there's anything particularly special about the gesture recognizers.

objorke:
Please also update the CHANGELOG.md with some info about the changes :-)

TheAlmightyBob:
I just realized that in my desire to make this a proper Pan/Zoom-dedicated gesture recognizer and support a future dedicated "Tap," this actually removes the current ability to fake tap detection via the OxyPlot touch events (it doesn't call Started until fingers move to indicate a pan/zoom).

That's probably a regression that should be dealt with. I'll think more about it later...

TheAlmightyBob:
Took a look at Windows, and the new behavior is more consistent with that, which also does not generate TouchStarted/Completed events for a tap. So maybe it's fine?

If somebody was relying on the previous behavior, they could instead attach their own TapGestureRecognizer to the PlotView and call relevant methods.. which would likely work better anyhow.

objorke:
I think this regression is ok, but it would be great to implement the tap gesture recognizers too!

Source/OxyPlot.MonoTouch/PanZoomGestureRecognizer.cs
```diff
@@ -0,0 +1,252 @@
...
+        /// </summary>
+        public bool AllowPinchPastZero { get; set; }
+
+        /// <summary>
+        /// The current calculated pan/zoom changes
+        /// </summary>
+        public OxyTouchEventArgs TouchEventArgs { get; set; }
+
+
+        public PanZoomGestureRecognizer()
```
objorke: missing XML comment - build warning!

Source/OxyPlot.MonoTouch/PanZoomGestureRecognizer.cs
```diff
@@ -0,0 +1,252 @@
...
+    {
+        /// <summary>
+        /// Up to 2 touches being currently tracked in a pan/zoom.
+        /// </summary>
+        private List<UITouch> activeTouches = new List<UITouch>();
+
+        // Distance between touchpoints when the second touchpoint begins. Used to determine
+        // whether the touchpoints cross along a given axis during the zoom gesture.
+        //
+        private ScreenVector startingDistance = default(ScreenVector);
```
objorke: missing XML comment [SA 1600]

Source/OxyPlot.MonoTouch/PanZoomGestureRecognizer.cs
```diff
@@ -0,0 +1,252 @@
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether this <see cref="OxyPlot.Xamarin.iOS.PanZoomGestureRecognizer"/> keeps the aspect ratio when pinching.
```
objorke: invalid cref (namespace should be OxyPlot.MonoTouch...)

TheAlmightyBob:
Whoops, didn't realize the XML comments were required, sorry.

Out of curiosity, what uses them? Where would the cref be supported?

objorke:
I have fixed the build errors. I also linked the `PanZoomGestureRecognizer` to avoid doing it twice.
I have recommended to follow the StyleCop rules in the [coding style](http://oxyplot.org/documentation/coding-style) page, but I think it is only the core and WPF projects where violations will actually break the build.

I don't think intellisense supports the crefs, but at some point I would like to generate API documentation!

TheAlmightyBob:
Thanks! I had assumed StyleCop was being run automatically when I built.

Issue #358, opened 2015-02-06
Change NuGet packages to use PCL bait and switch trick


objorke:
See http://log.paulbetts.org/the-bait-and-switch-pcl-trick/
How should the projects be set up? Should this only be applied to the Xamarin.Forms packages?

mattleibow:
Based on th structure of the Xamarin.Forms NuGet, this will happen automatically. This cannot be done with the "normal" platform specific NuGets as the interfaces are a bit different.

tibel:
OK, so bait-and-switch is only for the `OxyPlot.Xamarin.Forms` nuget package.
Still open is how to build the Xamarin.Forms package (see #274) with the right dependency versions.

PR #362, opened 2015-02-11
Update Xamarin NuGet

mattleibow:
Basically:

 - created an OxyPlot.Mobile.nuspec
 - created an OxyPlot.Mobile.sln
 - updated the Xamarin solution file versions
 - set the Android target version to 2.3 as the lower platforms are gong to be dropped
 - updated the OxyPlot.Xamarin.Forms.nuspec to use the new OxyPlot.Mobile NuGet


objorke:
This looks good! Do you think the OxyPlot.Xamarin.iOS, OxyPlot.Xamarin.Android, OxyPlot.WP8, OxyPlot.WindowsUniversal NuGet packages can be deleted now?

objorke:
And maybe the XF, iOS, Android, WU and WP8 .sln files could be deleted?

mattleibow:
Yes, that should be OK. Do you know when the current develop version of OxyPlot will go stable?

mattleibow:
I was thinking that if it is not going to be like this week, then I'll create the NuGets for the current stable version as well and get them up so that they can be used by developers right away...

objorke:
What about #358 - should the XF package use bait & switch?

mattleibow:
With the two nuspecs that are in this repo now, it will automatically switch them out depending on the platform the NuGet is installed to. So it should be working now.

objorke:
Great! Will this trick also solve #363?

objorke:
Forgot that the CHANGELOG.md file shold also be updated about changing Android target and the OxyPlot.Xamarin.Forms NuGet package change. I prefer the change log to be updated in each PR :-)

mattleibow:
@objorke I was just thinking, we may break quite a few builds if we start removing the current NuGets - there are several thousand downloads already.

What we can do to avoid this is to actually create an empty "Mobile" NuGet and then just reference the various platform specific ones. This way, we can avoid the breaks, but still provide a single NuGet that can be installed everywhere.

You can do the same for the "OxyPlot" NuGet and just create an empty NuGet with many dependencies. This will remove any possible conflicts with duplicate assemblies as everyone is referenceing the same packages in the end.

objorke:
I was thinking about adding dependencies the other way: e.g. OxyPlot.iOS depends on Oxyplot.Mobile. Mark OxyPlot.iOS and the others obsolete and unlist them after some time.

PR #364, opened 2015-02-11
Updated the changelog for my changes and made the Mobile NuGet empty

mattleibow:
Based on the changes in #362 :

 - Added the OxyPlot.Mobile NuGet
 - OxyPlot.Xamarin.Forms now depends on OxyPlot.Mobile

The change in the android target platform to 2.3 was only for the ExampleBrowser app, not the library - which is already done.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,14 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 ### Added
+- Created a new OxyPlot.Mobile NuGet to combine the mobile platforms into a single NuGet. (#362)
 - Support for XWT (#295)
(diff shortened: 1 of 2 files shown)


mattleibow:
Rather than adding the assemblies to a new NuGet, create an empty NuGet to:

 - avoid duplications of the various platforms
 - avoid breaking existing apps

objorke:
To reduce the number of packages, I think it would be better to reverse the dependencies (e.g. OxyPlot.iOS depends on OxyPlot.Mobile) and mark the old packages obsolete. Then after some time, unlist all the obsolete packages.

objorke:
I committed the modifications to the change log

mattleibow:
OK, that seems valid.

PR #373, opened 2015-02-16
Xamarin.Forms NuGet PCL fixes

mattleibow:
Specifying the Mobile NuGet only on platforms 
 - not supported in PCL
 - specify version number for Xamarin.Forms to prevent the IDE from picking the latest pre-release

Restore the NuGets before building OxyPlot.Mobile

### DIFF
--- a/Build/build-mobile.cmd
+++ b/Build/build-mobile.cmd
@@ -5,6 +5,7 @@ del /S /Q Output\*.*
 
 UpdateVersionNumbers.exe /VersionFromNuGet=OxyPlot.Core /Dependency=OxyPlot.Core /Dependency=OxyPlot.Mobile /ExtractReleaseNotes=CHANGELOG.md /Directory=.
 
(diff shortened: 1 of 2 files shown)


mattleibow:
@objorke I have specified a version number for Xamarin.Forms to prevent the IDE from downloading the pre-release versions. Do you think that there will be any issues? It does mean that the version will have to be specified when the project is updated...

The reason for specifying the Mobile NuGet for each platform is that the Mobile cannot be installed into a PCL as there is no assemblies to install. My first thought was to make it so that the Mobile NuGet could be installed into a PCL, but just not install anything. This seems a bit pointless, so I made the Xamarin.Forms NuGet rather only install the Mobile NuGet on the platforms that Xamarin.Forms actually supports.

Do you have any other suggestions?

objorke:
@mattleibow Both solutions sounds good to me. I think we should update to the latest stable Xamarin.Forms versions as soon as they are released.

Issue #383, opened 2015-02-19
proposal: small change to Axis to reveal "natural" dimensions


tr8dr:
## Problem
I have a series of PlotModels (Views) where I want to align 1 axis.  For example I may have a series of plots aligned vertically with the same X axis, but with different Y axes.

A problem to solve is that the render-width of each of these Y axes may be different (and in fact may change as a given plot is panned).   I can set __<model>.PlotMargins__ left margin to a fixed value, but this may clip the rendering for an axis.  The __ActualPlotMargin__ property will not adjust to the natural sizing if the __PlotMargin__ dimension is overridden.  Hence, I have a circular problem in adjusting the margins across panes in that if adjust cannot remeasure the natural sizing.

The rendering calls:
```csharp
 axis.Measure(IRenderContext);
``` 
on each axis and adjust to the max extent across the set.   This information is thrown-away if the PlotMargin dimension is set and not accessible at the application level.

## Proposal
Add an __Update(IRenderContext)__ method to Axis that either calls Measure(rc) or replaces it.  The method updates a property:  

```csharp
Thickness ActualDimensions { get; protected set; }
```
The internal renderer and application, alike, may now read this property.

objorke:
I don't think the layout "engine" should be exposed. To make it possible to maintain/refactor this engine, we should not expose any of the internal steps. Can we find a different solution to this issue?

Adding a property that contains the actual location and size of the axis sounds good. But it should use start with "Actual" for consistency.

tr8dr:
The layout methods can be protected.  I don't need to see these as a user of the library.   What I do need is a property which exposes the "desired" rendering width / height of the axis.   The desired height/width is distinct from the Actual height or width, as actual may reflect a user size override.

objorke:
I think I get it now. `DesiredSize` as in WPF [`UIElement.DesiredSize`](https://msdn.microsoft.com/en-us/library/system.windows.uielement.desiredsize%28v=vs.110%29.aspx) is a better name than `NaturalSize` in my opinion.

tr8dr:
I agree: DesiredSize sounds good.  Ideally this makes its way across other UI elements.   How would you propose unit testing?

objorke:
The WPF example in 6b10059 should be updated to demonstrate the functionality described in this issue.

tr8dr:
I can take a look at this when time permits (I would need to install a VM to work through it, as I run OS X exclusively).   With regard to making this work, there is a feature that would be useful:

[...]

PR #384, opened 2015-02-19
Added Axis.NaturalSize (issue #383)

tr8dr:
"Simple" change to expose Axis "natural size" so that application can make adjustments to align panes.


objorke:
#383 (link to issue should be in PR description, not title)

objorke:
It should not be 197 commits in this PR. Please rebase to latest before making a PR

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -543,6 +543,11 @@ public double Scale
         public bool UseSuperExponentialFormat { get; set; }
 
         /// <summary>
+        /// Gets or sets the natural dimensions of the axis, as of most recent rendering.
+        /// </summary>
+        public OxySize NaturalSize { get; protected set; }
```
objorke: I suggest "ActualSize" for consistency

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -835,85 +840,13 @@ public bool IsVertical()
...
 
         /// <summary>
-        /// Measures the size of the axis (maximum axis label width/height).
+        /// Updates the natural dimensions of axis with respect to graphics context and
+        /// any other graphical housekeeping.
         /// </summary>
         /// <param name="rc">The render context.</param>
-        /// <returns>The size of the axis.</returns>
-        public virtual OxySize Measure(IRenderContext rc)
+        public void Update(IRenderContext rc)
```
objorke: do not expose this - find other solution

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -1353,6 +1286,88 @@ protected internal virtual void ResetCurrentValues()
         }
 
         /// <summary>
+        /// Measures the size of the axis (maximum axis label width/height).
+        /// </summary>
+        /// <param name="rc">The render context.</param>
+        /// <returns>The size of the axis.</returns>
+        protected virtual OxySize Measure(IRenderContext rc)
```
objorke: yes, I think it is a good idea to change to protected

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
```diff
@@ -179,7 +179,10 @@ private static double MaxSizeOfPositionTier(IRenderContext rc, IEnumerable<Axis>
             double maxSizeOfPositionTier = 0;
             foreach (var axis in axesOfPositionTier)
             {
-                var size = axis.Measure(rc);
+                // update axis with latest render context, adjusting dimensions, etc.
+                axis.Update(rc);
```
objorke: can we find some other solution?

Source/OxyPlot/Axes/Axis.cs
tr8dr: Isn't ActualSize used to indicate the size that was chosen for rendering?  NaturalSize indicates the "natural" size of the element if involved no clipping / size override.   Correct me if I am wrong, if size is overriden by the user then Actual variables will reflect the override.

Source/OxyPlot/Axes/Axis.cs
tr8dr: protected internal?   (don't know if you need internal or not)?

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
tr8dr: What would you suggest?  I just need a property on Axes which reflects the "desired" or "natural" size (as opposed to the rendered size which may reflect user size overrides).

Source/OxyPlot/Axes/Axis.cs
objorke: Right, the "ActualX" properties should indicate the actual size that was chosen for rendering.
Sorry, I don't get the purpose of this new property yet...

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
objorke: or minimum/maximum size?

objorke:
Need an example that shows the necessity of the new property.

tr8dr:
I put an example in the issue, where explained why I need it.  Do you need more clarification?

objorke:
The example should be runnable and included in the PR

tr8dr:
Given that aligning 2 different panes is specific to a particular rendering library (Gtk, WPF, etc) I cannot provide a visual example for this.  The important thing is that the property exists and has proper values.  The best I can do is have the property value render in the legend or something like that?  What would you like to see?

Source/OxyPlot/Axes/Axis.cs
tr8dr: re: rebase, i tried that.  When I submitted this saw that included intermediate commits (sorry).  I do not know why the intermediate changes in the main repo show up here.  Obviously I did something wrong.  I followed the following guide:  

http://robots.thoughtbot.com/keeping-a-github-fork-updated

objorke:
If we have no example where we can test that this actually works, this feature may easily be broken in the future. A unit test that asserts on values is ok, I think that can be done in portable code.
An platform specific example in Gtk or any of the other platforms would also be great to debug and review the solution!

tr8dr:
I can make the property visible as text in a plot.  That would work in the ExamplesLibrary without issue (indeed this is how I tested it initially).

objorke:
Text in the plot is not very helpful, it will be very difficult to see regressions on the number in the future.

Sorry for being a bit difficult on this issue, but I am concerned about the maintainability of the library. The layout engine is one part I would really like to see being refactored and improved.

tr8dr:
I agree with the point.  How do you test rendering, which involves a graphics context and event loop though?  Do you have this rigged for testing?   I took a look at OxyPlot.Tests but did not notice that sort of unit test.

tr8dr:
If you can point me to a unit test example that would work here, greatly appreciated.   I like to unit test my libraries as well.  I have never done one for UIs though.

Source/OxyPlot/Axes/Axis.cs
objorke: I think the `DesiredSize` property can be set during the call to `Measure` - I understand this is how it is done in WPF. Revert (remove) the Update method.

objorke:
Rendering a `PlotModel` on a Svg or Pdf context should be possible in the OxyPlot.Tests project. After the `PlotModel` has been updated/rendered, it should be possible to assert the `Axis.DesiredSize` property.

Personally, I would have created a example in the [WpfExamples](https://github.com/oxyplot/oxyplot/tree/develop/Source/Examples/WPF/WpfExamples/Examples) project showing exactly how you would use the new property in your application. This is probably not much more work than writing a unit test, and can be a good demo on how to use the property. It also means that the use case you suggest is more likely to be supported in the future! :-)

tr8dr:
This branch is a mess because of my failed rebase.  Let me close this pull-request and see whether I can make proper and re-issue.   I am still not sure how to rebase the fork properly.  What I did last time apparently did not work ...

objorke:
See #386 for unit tests and wpf example

tr8dr:
I can't do a WPF (since I run on OS X), but I should be able to do something with pdf ...

PR #387, opened 2015-02-21
Addition of Axis.DesiredSize

tr8dr:
This is the "clean" resubmission for issue #383. See issue for full details.  Includes a unit test checking proper behavior of the property on initial render + subsequent render with an axis range change (using objorke's null renderer).

Note that I added OxyPlot.Tests.csproj reference to the OxyPlot.GtkSharp.sln solution so GTK developers can run both visual tests and unit tests.   If you prefer not to have this referenced in the sln, can revert.


objorke:
A little comment: Please rebase before pushing and making a pull request. I think this is one change and should be one commit and not five. :-)

Another minor comment: noone else is starting the commit messages with "- ", consider following the common style :-)

https://wiki.openstack.org/wiki/GitCommitMessages

tr8dr:
fine, can use a diff comment style for this project.  re: commits, I needed to commit my changes and then merge from your branch to add the unit tests.   Also style copy on the build flags style "errors" that are not found by the style cop in xamarin, so takes more than one iteration.

tr8dr:
btw, wanted to say, OxyPlot is a great project & has made it possible for me to develop an application which would otherwise be difficult to do on OS X, given the lack of libraries like this.   Thanks for putting this together in the open source community.

Issue #389, opened 2015-02-21
fixes for CandleStickAndVolumeSeries + VolumeSeries


tr8dr:
1. missing UpdateMaxMin() for default axis sizing
    * I added UpdateMaxMin() and some behavior around this to allow the use of these series without explicitly setting Minimum and Maximum on the axes.
2. CandleStickAndVolume series would throw an exception if created with a Y axis without the "Bars" key (normally used to distinguish the volume and bars axis)
    * now will find a default axis if Axis.Key = "Bars" is not available

I have implemented tests for the above.

objorke:
two different bugs -> should be two issues! :-)

PR #390, opened 2015-02-21
fixes for CandleStickAndVolumeSeries + VolumeSeries

tr8dr:
This addresses issue #389:

1. missing UpdateMaxMin() for default axis sizing
    * added UpdateMaxMin() and some behavior around this to allow the use of these series without explicitly setting Minimum and Maximum on the axes.
2. CandleStickAndVolume series would throw an exception if created with a Y axis without the "Bars" key (normally used to distinguish the volume and bars axis)
    * now will find a default axis if Axis.Key = "Bars" is not available

I have implemented tests for the above.

### DIFF
--- a/Source/Examples/ExampleLibrary/Series/FinancialSeries/CandleStickAndVolumeSeriesExamples.cs
+++ b/Source/Examples/ExampleLibrary/Series/FinancialSeries/CandleStickAndVolumeSeriesExamples.cs
@@ -18,7 +18,7 @@ namespace ExampleLibrary
     public static class CandleStickAndVolumeSeriesExamples
     {
         [Example("Candles + Volume (combined volume), adjusting Y-axis")]
-        public static Example CombinedVolume_Adj()
+        public static Example CombinedVolume_Adjusting()
         {
             return CreateCandleStickAndVolumeSeriesExample(
                 "Candles + Volume (combined volume)", 
@@ -28,7 +28,7 @@ namespace ExampleLibrary
         }
 
         [Example("Candles + Volume (combined volume), natural Y-axis")]
-        public static Example CombinedVolume_Nat()
+        public static Example CombinedVolume_Natural()
         {
             return CreateCandleStickAndVolumeSeriesExample(
                 "Candles + Volume (combined volume)", 
@@ -38,7 +38,7 @@ namespace ExampleLibrary
         }
 
         [Example("Candles + Volume (stacked volume), adjusting Y-axis")]
-        public static Example StackedVolume_Adj()
+        public static Example StackedVolume_Adjusting()
         {
             return CreateCandleStickAndVolumeSeriesExample(
                 "Candles + Volume (stacked volume)", 
@@ -48,7 +48,7 @@ namespace ExampleLibrary
         }
 
         [Example("Candles + Volume (stacked volume), natural Y-axis")]
-        public static Example StackedVolume_Nat()
+        public static Example StackedVolume_Natural()
         {
             return CreateCandleStickAndVolumeSeriesExample(
                 "Candles + Volume (stacked volume)", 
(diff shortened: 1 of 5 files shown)


Source/Examples/ExampleLibrary/Series/FinancialSeries/CandleStickAndVolumeSeriesExamples.cs
```diff
@@ -17,42 +17,101 @@ namespace ExampleLibrary
     [Tags("Series")]
     public static class CandleStickAndVolumeSeriesExamples
     {
-        [Example("Candles + Volume (combined volume)")]
-        public static Example CombinedVolume()
+        [Example("Candles + Volume (combined volume), adjusting Y-axis")]
+        public static Example CombinedVolume_Adj()
```
objorke: do not use abbreviations in .NET: https://msdn.microsoft.com/en-us/library/141e06ef%28v=vs.71%29.aspx
I also try to follow this rule on unit tests

Source/OxyPlot/Series/FinancialSeries/CandleStickAndVolumeSeries.cs
```diff
@@ -101,7 +106,25 @@ public LinearAxis BarAxis
...
+        { 
+            get
+            {
+                return this.volumeStyle;
+            }
+
+            set
+            {
+                // if style has changed adjust axis min/max
+                if (value != this.volumeStyle && this.VolumeAxis != null)
```
objorke: other parts of the library put this responsibility on the client. Should be consistent + keep it simple

Source/OxyPlot/Series/FinancialSeries/CandleStickAndVolumeSeries.cs
```diff
@@ -517,8 +555,122 @@ protected internal override void UpdateData()
         /// </summary>
         protected internal override void EnsureAxes()
         {
-            base.EnsureAxes();
+            // find volume axis
             this.VolumeAxis = (LinearAxis)this.PlotModel.Axes.FirstOrDefault(a => a.Key == "Volume");
```
objorke: such magic key names are not used in other series [consistency]

Source/OxyPlot/Series/FinancialSeries/CandleStickAndVolumeSeries.cs
```diff
@@ -517,8 +555,122 @@ protected internal override void UpdateData()
...
+
+            // assign the bar axis if not found
+            var barAxis = this.PlotModel.Axes.FirstOrDefault(a => a.Key == "Bars");
+            if (barAxis == null)
+            {
+                // determine a bar axis fallback
+                var defAxis = this.PlotModel.Axes.FirstOrDefault(a => a.IsVertical() && a.Key != "Volume");
+               
+                // would be nice to avoid having to change the key & override base method
+                defAxis.Key = "Bars";
```
objorke: NO! Do not change the key

Source/OxyPlot/Series/FinancialSeries/VolumeSeries.cs
```diff
@@ -79,7 +85,28 @@ public List<OhlcvItem> Items
...
         /// </summary>
-        public VolumeStyle VolumeStyle { get; set; }
+        public VolumeStyle VolumeStyle 
+        { 
+            get
+            {
+                return this.volumeStyle;
+            }
+        
+            set
```
objorke: see comment on CandleStickAndVolumeSeries

Source/OxyPlot/Series/FinancialSeries/CandleStickAndVolumeSeries.cs
tr8dr: My preference would be not to do this, but to be able to override the axis search.  If volume is not to be displayed, would like to have it default to the 1st YAxis.   CandleStickAndVolumeSeries can be used to just display candlesticks.  

Would you suggest making the keying mandatory for just bar display (as would be the default without this) or finding another way to assign default?    I wanted to override the base method where these are assigned but could not write to the YAxis property.

Source/OxyPlot/Series/FinancialSeries/CandleStickAndVolumeSeries.cs
tr8dr: agree.

objorke:
looks good!

Issue #136, opened 2014-08-14
Automatic invalidation of the plot view when the model changes


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10193](http://oxyplot.codeplex.com/workitem/10193)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-05-08 18:27
Currently this library is designed to not support automatic invalidation when properties of the
<code>PlotModel</code> changes. Observing all changes to the model (or to bound data) is a pretty big task... But we can change it, so please vote if you think should be prioritized!

**[tibel](http://www.codeplex.com/site/users/view/tibel)** 2014-05-09 07:19
Also you have to add something like <code>BeginUpdate()</code> and <code>EndUpdate()</code> to disable the invalidation while changing multiple properties in the PlotModel. Otherwise this would kill performance totally.

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-05-09 09:07
Yes, there should be a way to suspend the invalidation. This should be easy. The challenge is to get notifications from all property and collection changes. I think INotifyPropertyChanged and INotifyCollectionChanged can be used. But I would like to avoid
 weaving or AOP in this library, so this means manually editing a lot of properties. An alternative could be to use automatic code generation of the property setters and getters...

Also, some properties may require an update of the data, others only affects the visuals.

PR #398, opened 2015-02-27
Listen to CollectionChanged events in Series (WPF)

MendelMonteiro:
when Series.ItemsSource supports it listen to CollectionChanged events
and invalidate the plot when the event is raised.


objorke:
1. A pull request should be linked to an issue. I guess this is related to #136 but we could create a new issue for this enhancement (not for the portable model, but for the WPF implementation only)
2. Are you sure this change will not leak memory?
3. Should changes on properties in the collection items also be observed? (INotifyPropertyChanged)
4. CHANGELOG.md should be updated (including issue #). Please also update the CONTRIBUTORS file

MendelMonteiro:
1. Done

2. Well spotted we can indeed leak references of Series, I've changed it to use CollectionChangedEventManager.

3. There are 3 uses cases that I can see, fixed axis points that change value, watching a slice of a moving series and watching a series that grows.  Subscribing to the CollectionChanged events will handle the latter two but not the first.  I've created a separate issue for this.

4. Done

objorke:
Is there a problem with the .NET 4.0 or .NET 3.5 builds?

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ All notable changes to this project will be documented in this file.
 - Separated WPF Plot and PlotView (#252,#239)
 - Current CandleStickSeries renamed to OldCandleStickSeries, replaced by a faster implementation (#369)
 - Fixed axis min/max calc and axis assignment for CandleStick + VolumeSeries (#389)
+- Invalidate plot when ItemsSource contents change (INotifyCollectionChagned) on WPF only (#406)
```
objorke: INotifyCollectionChanged

objorke:
Can you please rebase and create a new PR? I think the 11 commits should be 1. :-)
And what about an example? Can we test this new feature in one of the existing examples?

objorke:
I have tested the example
1. Loading the example is very slow when the observable collection is not disconnected when adding points.. I think adding the 500 points means the model is updated 500 (!) times before anything is rendered. I think this can be fixed in the OxyPlot.Wpf.PlotBase class, but then we are into architecture changes that should be aligned on all the different implementations...
2. Pressing "Load" again gives a strange result. I think it would be better to change the example to "Add 500 points" and not reset the x-coordinate every time.
3. I think there is a panning bug in both the binding demos, I have not seen this before. Will add a bug issue about this.

MendelMonteiro:
1. The fact that loading is not batched up is a limitation that can be worked around by creating a disconnectable ObservableCollection which should be trivial.

2.  There is no added-value to being able to click the Load button more than once - the objective of the demo is to show that the collection is being observed.  I can maybe disable the Load button after the first click.

3. I noticed this today too, it seems to only impact the Plot and not PlotView and I was able to reproduce it using the current alpha version on nuget.

PR #423, opened 2015-03-27
Wpf: Override FrameworkElement.LogicalChildren in Plot

### LINKED ISSUE #40: Plot control in DataTemplate does not bind to correct data
**Imported from [http://oxyplot.codeplex.com/workitem/10000](http://oxyplot.codeplex.com/workitem/10000)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2012-09-02 18:33
The error is reproduced in WpfExamples 

See <a href="http://oxyplot.codeplex.com/discussions/355389" rel="nofollow">http://oxyplot.codeplex.com/discussions/355389</a>

draagyn:
Fixes #40 

Needed to override FrameworkElement.LogicalChildren in order to pass on DataContext and other inherited properties.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -41,6 +41,7 @@ Jonathan Shore <jonathan.shore@gmail.com>
 julien.bataille
 Just Slon <just.slon@gmail.com>
 kenny_evoleap
+Kenny Nygaard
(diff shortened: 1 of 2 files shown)


objorke:
Great! The change log should also be updated (preferably in the same commit!)

Issue #274, opened 2014-12-08
CI for the Xamarin projects


objorke:
The following projects should also be continuously integrated:
- OxyPlot.Xamarin.Android
- OxyPlot.Xamarin.iOS
- ~~OxyPlot.Xamarin.Mac (must be built on a Mac)~~
- OxyPlot.Xamarin.Forms (must be built on a Windows machine due to the Windows app renderer)

objorke:
There is a thread on the appveyor forum:
http://help.appveyor.com/discussions/suggestions/167-need-support-for-xamarinandroid-and-xamarinios

See also
http://xamarin.com/faq#q33

david-laundav:
There is further information in the following article:
http://developer.xamarin.com/guides/cross-platform/ci/

tibel:
I think there will be problems with resolving the dependencies here.
Especially `OxyPlot.Mobile` and `OxyPlot.Xamarin.Forms` will have problems,
because `OxyPlot.WP8` and `OxyPlot.WindowsUniversal` are build by AppVeyor.

Best would be that the Xamarin build depends on `OxyPlot.Core`, `OxyPlot.WP8` and `OxyPlot.WindowsUniversal` nuget packages only.

@mattleibow Do you have any suggestions also how this relates to https://github.com/oxyplot/xamarin-component?

tibel:
For me the versions of the published nuget packages seems to be a little random.

Package | Version
------- | -------
OxyPlot.Xamarin.iOS | 2015.1.721-alpha
OxyPlot.Xamarin.Mac | 2015.1.781-alpha
OxyPlot.Xamarin.Android | 2015.1.721-alpha
OxyPlot.Mobile | 2015.1.804-alpha
OxyPlot.Xamarin.Forms | 2015.1.804-alpha

Correct me if I'm wrong:
- the build currently uses the Xamarin.Mobile solution
- which rebuilds all the dependent (native) projects (e.g. OxyPlot.WP8)
- but OxyPlot.WP8 was build by AppVeyor already (duplicate?)
- the nuget packages for (e.g OxyPlot.Xamarin.iOS) are not updated

For me it looks like if it would be easier to maintain when:
- OxyPlot.Xamarin.Forms is build last
- does not rebuild the native projects
- only depend on the published nuget packages (no problems with different versions of the packages)
- so it would make sense to have it in a separate repo because of the above

@objorke What do you think on the Xamarin.Forms build?

objorke:
[...]

Issue #242, opened 2014-11-07
Reorganize NuGet packages


objorke:
Reorganize/clean up the NuGet packages to make it easier to understand which package to use.

List of current NuGet packages: http://www.nuget.org/packages?q=oxyplot

Packages built by AppVeyor:

- OxyPlot.Core - PCL, NET35, NET40, NET45 libraries
- OxyPlot.Wpf (NET4 WPF+XPS)
- OxyPlot.Windows (Windows store/WinRT and Windows mobile apps) NEW!
- OxyPlot.WP8 (Windows Phone 8 Silverlight)
- OxyPlot.Silverlight (SL5)
- OxyPlot.WindowsForms (NET4)
- OxyPlot.GtkSharp (NET4)
- OxyPlot.Xwt (NET4, based on Xwt)
- OxyPlot.Pdf (NET4, adds dependency to PdfSharp)
- OxyPlot.OpenXML (NET 4, adds dependency to OpenXML)

Packages built on machines with a Xamarin subscription

- OxyPlot.Xamarin.Android
- OxyPlot.Xamarin.iOS
- OxyPlot.Xamarin.Mac (must be built on a Mac...)
- OxyPlot.Xamarin.Forms (must be built on a Windows machine for the Windows app renderer...)

Remove / mark obsolete
- OxyPlot
- OxyPlot.Mobile
- OxyPlot.WindowsUniversal

tibel:
Also issue #242 describes these build issues.

objorke:
There is currently 19 packages, I think this can still be improved. 
I think #274 is about CI of the Xamarin packages

tibel:
A year ago I thought similar: Many packages are confusing. But changed my mind when seeing CoreFX the first time. They have much more packages than OxyPlot might ever have.

Thus said:
OxyPlot supports many UI (or rendering) technologies and there is more to come (e.g. DirectX, Eto?).
NuGet has the concept of platforms, but this is not enough as there might be multiple UI stacks on the platform.

For .NET 4.0 OxyPlot supports:
- WPF
- WinForms
- GTK#
- Xwt

As an OxyPlot user I will (probably) NOT use multiple UI stacks in one application. When I create a WPF app, all I need is OxyPlot.Wpf. I don't want some WinForms assembly references in my project.

[...]

PR #428, opened 2015-03-28
Nuget package cleanup

tibel:
- correct the package dependencies
- changed `OxyPlot` and `OxyPlot.Mobile` to be meta-packages that only reference other packages
- added `LICENSE` and `README.md` to all packages

### DIFF
--- a/Source/OxyPlot.GtkSharp/OxyPlot.GtkSharp.nuspec
+++ b/Source/OxyPlot.GtkSharp/OxyPlot.GtkSharp.nuspec
@@ -18,6 +18,11 @@
   </metadata>
   <files>
     <file src="..\..\Output\NET40\OxyPlot.GtkSharp.???" target="lib\net40" />
-    <file src="..\..\README.md" />
+
+    <file src="..\LICENSE" />
+    <file src="..\AUTHORS" />
+    <file src="..\CONTRIBUTORS" />
+    <file src="..\README.md" />
+    <file src="..\CHANGELOG.md" />
   </files>
 </package>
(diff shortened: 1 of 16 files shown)


objorke:
Thanks for cleaning up the nuspecs!

objorke:
It seems AppVeyor did not create the packages. I think there are some missing `"..\"` in the nuspecs

objorke:
I'll try to fix this.

objorke:
The relative paths of the Xamarin nuspecs are different, I think.

tibel:
Sorry for that. I did some copy-paste and have not verified the paths.

mattleibow:
@objorke @tibel Just to confirm. Are we going to be using separate packages for each platform?

tibel:
@mattleibow The packages were already existing. The only change I did was to remove duplicates. So one assembly is only in one package. When you verify the changes you will see that some assemblies where part of multiple packages.

For example `OxyPlot.WP8.dll` was part of the packages:
- OxyPlot.WP8
- OxyPlot
- OxyPlot.Mobile

objorke:
Is it a problem that the packages are not in the same build? (the Xamarin packages are built by Xamarin, not by AppVeyor)

mattleibow:
@tibel I know about those... just wanted to check to see as @objorke was wanting to eliminate the large number of packages. Either way, doesn't really matter, other than I have to update my publish scripts.

objorke:
@mattleibow @tibel Yes, I would think it would be nice to reduce the number of packages - it is confusing to see 19 packages in NuGet. I created #242 to clean up (#274 is about CI). Can we re-open #242 and discuss the best solution there?

objorke:
1. Is it possible to create a package that includes both winforms, wpf, gtk and xwt assemblies (all for NET4x) and automatically adds only the relevant references? I think this was the reason why I created some of these packages.
2. Is package download size important? I guess having meta packages is good to reduce the size of the downloads.

tibel:
Yes, you can put all in one package, BUT then you have to code install/remove of each assembly by hand in PowerShell (I would not try this; you loose nearly all nuget features). The package size is not so relevant (at least for me).

mattleibow:
@tibel I agree with you. Separate the UI platforms and leave it up to the dev to pick what should be installed.

Issue #414, opened 2015-03-17
Tracker format strings should use labels instead of indices


objorke:
Current default tracker format string for `HeatMapSeries`: 
`"{0}\n{1}: {2}\n{3}: {4}\n{5}: {6}"`

This could be changed to
`"{Title}\n{XAxisTitle}: {XAxisValue:0.000}\n{YAxisTitle}: {YAxisValue:0.000}\n{ColorAxisTitle}: {Value}"`

This should make it easier to extend with new parameters without creating breaking changes

johnguidone:
Has there been any movement on this? Currently attempting to display X, Y, Z values on tracker format string for a heat map, along with custom calculations (dz/dx and dz/dy derivatives).

PR #440, opened 2015-04-07
Added capability to display mean value to BoxPlot

stefan-schweiger:
Added Mean value (and constructor) to BoxPlot item
Changed BoxPlotSeries do display Mean value
Added BoxPlot Demo


Source/OxyPlot/Series/BoxPlotSeries.cs
```diff
@@ -94,6 +97,18 @@ public BoxPlotSeries()
         public double MedianThickness { get; set; }
 
         /// <summary>
+        /// Gets or sets the size of the mean point.
+        /// </summary>
+        /// <remarks>This property is only used when ShowMeanAsDot = true.</remarks>
+        public double MeanPointSize { get; set; }
```
objorke: or `MeanDotSize`?

Source/OxyPlot/Series/BoxPlotSeries.cs
```diff
@@ -130,6 +145,11 @@ public BoxPlotSeries()
         public bool ShowMedianAsDot { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to show the mean as a dot.
+        /// </summary>
+        public bool ShowMeanAsDot { get; set; }
```
objorke: wouldn´t it be better to use an enum here? (also for the ShowMedianAsDot)

Source/OxyPlot/Series/BoxPlotSeries.cs
```diff
@@ -383,6 +404,35 @@ public override void Render(IRenderContext rc, PlotModel model)
...
+                        clippingRect,
+                        new[] { meanLeft, meanRight },
+                        0,
+                        strokeColor,
+                        this.StrokeThickness * this.MeanThickness,
+                        LineStyle.Dash.GetDashArray(),
+                        LineJoin.Miter,
+                        true);
+                }
+                else if (!double.IsNaN(item.Mean))
```
objorke: duplicate check of NaN - isn´t it better to nest the ifs here?

Source/OxyPlot/Series/BoxPlotSeries.cs
stefan-schweiger: It's also called `MedianDotSize` so I just stuck with the name

Source/OxyPlot/Series/BoxPlotSeries.cs
stefan-schweiger: Yeah I thought about that, but I didn't want to introduce breaking changes for this, but if you think it's better to change this I can implement a enum as well ;)

Source/OxyPlot/Series/BoxPlotItem.cs
```diff
@@ -36,12 +36,39 @@ public struct BoxPlotItem
...
+        /// <param name="x">The x.</param>
+        /// <param name="lowerWhisker">The lower whisker.</param>
+        /// <param name="boxBottom">The box bottom.</param>
+        /// <param name="median">The median.</param>
+        /// <param name="mean">The mean.</param>
+        /// <param name="boxTop">The box top.</param>
+        /// <param name="upperWhisker">The upper whisker.</param>
+        /// <param name="outliers">The outliers.</param>
+        /// <param name="tag">The tag.</param>
+        public BoxPlotItem(
```
objorke: it is difficult to maintain ctors with arguments - cannot change order or add/remove without making a breaking change. I prefer to use parameterless ctors and use initializers instead!

Source/OxyPlot/Series/BoxPlotSeries.cs
```diff
@@ -22,7 +22,7 @@ public class BoxPlotSeries : XYAxisSeries
         /// <summary>
         /// The default tracker format string
         /// </summary>
-        public new const string DefaultTrackerFormatString = "{0}\n{1}: {2}\nUpper Whisker: {3:0.00}\nThird Quartil: {4:0.00}\nMedian: {5:0.00}\nFirst Quartil: {6:0.00}\nLower Whisker: {7:0.00}";
+        public new const string DefaultTrackerFormatString = "{0}\n{1}: {2}\nUpper Whisker: {3:N2}\nThird Quartil: {4:N2}\nMedian: {5:N2}\nMean: {6:N2}\nFirst Quartil: {7:N2}\nLower Whisker: {8:N2}";
```
objorke: I think this requires a "changed" note in the changelog

Source/Examples/WPF/WpfExamples/WpfExamples.csproj
```diff
@@ -73,6 +73,9 @@
     <Compile Include="Examples\BindingObservableCollectionDemo\MainWindow.xaml.cs">
       <DependentUpon>MainWindow.xaml</DependentUpon>
     </Compile>
+    <Compile Include="Examples\BoxPlotSeriesDemo\MainWindow.xaml.cs">
```
objorke: _NET 40 project not updated?

Source/OxyPlot/Series/BoxPlotItem.cs
stefan-schweiger: You can't have a parameterless ctors for a struc, so there is no real solution for this. But it doesn't break anything, as the old constructor is still in there, it just now calls the new constructor with a default value for mean.

objorke:
It would be great to see examples in the ExampleLibrary (very easy to add examples there!) with the different settings!

stefan-schweiger:
The example I've added to the WPF example project actually shows boxplots with and without mean value.

Source/OxyPlot/Series/BoxPlotSeries.cs
stefan-schweiger: I'm not a big fan of "over"-nesting everything, but it's your project so whatever codestyle you like better is fine to me. so should I change this?

Source/Examples/WPF/WpfExamples/WpfExamples.csproj
stefan-schweiger: Oh I didn't see the NET 4.0 project file as it isn't loaded by default in VS. I will also update it.

Source/OxyPlot/Series/BoxPlotSeries.cs
stefan-schweiger: Is the TrackerFormatString really so important to include it in the changelog isn't it enough to just include "Added Mean to BoxPlot" (which I also forgot to do haha)?

princeoffoods: Please add it to the changelog, All custom TrackFormatStrings will have to be updated.

Source/OxyPlot/Series/BoxPlotSeries.cs
objorke: I think this should be consistent witht the marker type, size, color of the LineSeries/ScatterSeries etc.

Source/OxyPlot/Series/BoxPlotSeries.cs
objorke: I think duplicate code is a worse anti-pattern than some nesting. To make it more maintainable (and allow more styles than dot/line) I think the drawing code here could be moved to a new private method.

Source/OxyPlot/Series/BoxPlotItem.cs
objorke: thanks, you are right. what about only one constructor?

Source/OxyPlot/Series/BoxPlotSeries.cs
objorke: I agree with @princeoffoods. Note that #414 would mitigate this problem.

stefan-schweiger: Oh ok now I see the problem. Yeah I will add it to the changelog.

Source/OxyPlot/Series/BoxPlotItem.cs
stefan-schweiger: so what would be your suggestion? just use the constructor with mean (which is a breaking change) or use the the one without mean and just use the setter for mean?

With most data types I would be on your side about not being able to change/add/remove the parameters. But a BoxPlot is defined by it's 5 base values (Lower Whisker + Box, Median, Upper Box + Whisker) and the 2 often optional values (Mean and Outliers) so if we would have to settle on one constructor (two constructors are still fine for me), I would probably go with something like this:

```c#
public BoxPlotItem(
    double x,
    double lowerWhisker,
    double boxBottom,
    double median,
    double boxTop,
    double upperWhisker)
    : this()
{
    ...
}
```

as those are the only 5 values really needed for displaying a boxplot and initialize Mean, Outliers and Tag (<-- which should definitely not be in the constructor for this data type) with the public setters.

What's your opinion on that?

Source/OxyPlot/Series/BoxPlotSeries.cs
stefan-schweiger: So if we are doing breaking changes I would propose something like this (and the same also for mean)

```c#
public MarkerStyle MedianMarkerStyle { get; set } // default None
public LineStyle MedianLineStyle { get; set } // default Line
public double MedianLineThickness { get; set } // default 2.0
public double MedianMarkerThickness { get; set; } // default 2.0
```

But this would allow to display a Dot and a Line mixed for Mean/Median, so maybe a seperate enum (Line,  Point, Dash, Circle, Square, Star) would also be suitable.

Source/OxyPlot/Series/BoxPlotSeries.cs
stefan-schweiger: Yeah you are right, but it makes no sense to change this before we haven't decided what to do about the Line/Maker style stuff overall.

Source/OxyPlot/Series/BoxPlotSeries.cs
objorke: the property can be kept, but marked obsolete. Change the getter/setter to use the enum property.

Source/OxyPlot/Series/BoxPlotItem.cs
objorke: I was thinking that the required values should be first arguments of the ctor, and the optional ones (mean, outliers, tag) should be at the end.

Source/Examples/WPF/WpfExamples/WpfExamples.csproj
objorke: good. I use a merge tool (e.g. WinMerge) to keep the project files in sync.

objorke:
The advantage of the ExampleLibrary is that the examples can be shown on all platforms.

Source/OxyPlot/Series/BoxPlotSeries.cs
objorke: Are we feature bloating this class now? 
The next feature requests will be custom median/mean marker shapes, fill colors etc.

Is line style really needed? Do we need other marker styles than line and dot? 
Shouldn't MedianMarkerThickness be renamed to MedianMarkerSize? 
Can all markers (outliers, median, mean) have the same marker sizes?

stefan-schweiger: Yeah, that's what I was also thinking about, but I guess for now the possibility to set some different styles and sizes should be enough.

It was just a suggestion, as I'm myself not 100% sure it is clear enough, but this would be more or less one enum to represent the following things:

```c#
public enum WhatEverStyle {
    None, // Pretty clear I guess
    Line, // draw a simply line
    Point, // draw a dotted line
    Dash, // draw a dashed line
    Circle, // draw a circle marker
    Square, // draw a square marker
    Star // draw a star marker
}
```

this would be the clearest way to prevent mean and median from being drawn as a line and a marker at the same time. and you don't have to set two properties every time you want to display it as a circle

from what I did until now with BoxPlot charts I can tell you that outliers should definitively be separated style/size wise from the rest. median and mean could probably be the same size but should always have different styles (for clear reasons).

Source/OxyPlot/Series/BoxPlotItem.cs
stefan-schweiger: this would still be a breaking change, but it's also ok for me.

stefan-schweiger:
aah ok now I get what you meant, no problem I will look into it tomorrow ;)

stefan-schweiger:
I've changed/added the mean stuff in the ExampleLibrary and also changed to constructor.
Are you ok with those changes?

**PS:** I will to the squash commits once we are finished with all issues of this PR

objorke:
Most of it looks good! But I still think we could clean up the size/style properties. The next feature request may be to add L-estimators like interquartile range, midhinge, range, mid-range, and trimean (http://en.wikipedia.org/wiki/Box_plot), and then we will end up with even more size properties if we follow the current pattern.

The wikipedia page is not mentioning the mean - how common is it to show this value? If it is less common than  outliers, then I think it should be behind outliers in the BoxPlotItem ctor.

stefan-schweiger:
It's not that uncommon to display a mean value in a boxplot. It's actually even one of the first points in "Types of boxplots" in the Wikipedia article.
Also if you do a google image search for BoxPlot about 1/3 to 1/2 of the results contain a mean (mostly displayed as a cross or circle)

screenshot 2015-04-09 00 00 46

But your point about the style properties is still valid. So what do you want to do about that? I've proposed some ways above, any thoughts on them?

objorke:
OK, I thought the mean was only used for the end of the whiskers.
I think we can go for your current solution since it is consistent with the current implementation Let us consider to refactor this later, if needed.
Please squash the commits, and I'll merge!

stefan-schweiger:
So you just want to stick with the `MedianPointSize` naming and not using a enum and if necessary change this later on?

Also as a reminder the `BoxPlotItem` constructor now contains a breaking change, is this ok for you?

objorke:
Yes, let us just make it consistent with the existing code for now, then refactor it later.
Isn´t it possible to avoid making a breaking change if you add the mean parameter (with a default value) last in the constructor?

stefan-schweiger:
it's possible to put Mean at the end, but putting it behind Tag is a very weird place for this parameter (as Tag is probably the less used one and as I've stated before Tag shouldn't be in the constructor at all in my opinion).

So this would be my preferred alternative. What do you thing about that?

```
[System.Obsolete("Use the other extended constructor")]
public BoxPlotItem(
    double x,
    double lowerWhisker,
    double boxBottom,
    double median,
    double boxTop,
    double upperWhisker,
    // Revert back to not having a default value, as I introduced it
    // with my last PR nad not many people should be using it by now
    IList<double> outliers, 
    object tag = null)
    : this(x, lowerWhisker, boxBottom, median, boxTop, upperWhisker, double.NaN, outliers, tag)
{
}

public BoxPlotItem(
    double x,
    double lowerWhisker,
    double boxBottom,
    double median,
    double boxTop,
    double upperWhisker,
    double mean = double.NaN,
    IList<double> outliers = null,
    object tag = null)
    : this()
{
    this.X = x;
    this.LowerWhisker = lowerWhisker;
    this.BoxBottom = boxBottom;
    this.Median = median;
    this.BoxTop = boxTop;
    this.UpperWhisker = upperWhisker;
    this.Mean = mean;
    this.Outliers = outliers ?? new List<double>();
    this.Tag = tag;
}
```

objorke:
I can think of the following "rules" to avoid a lot of constructors, breaking changes and future maintenance problems:
- all required parameters first, optional parameters last
- do not change the order of the required parameters (breaking change)
- always add new optional parameters at the end even if the new parameters are considered more important (avoid breaking changes)

We could require to use [named parameters](https://msdn.microsoft.com/en-us/library/dd264739.aspx), then it could be allowed to change the order of the optional parameters.

An alternative is to remove all the optional parameters (I like this better, but this is a breaking change...)

objorke:
I think it would be better if the `Mean` would have the last index in the tracker format string - avoiding breaking changes.

stefan-schweiger:
I know that breaking changes suck, but sometimes they should be done if the alternative is an illogical structured code.

objorke:
I will not block the PR because of this. But I think we have an anti-pattern going if there will be breaking changes every time we want to add a new feature...
Can you squash the commits before I merge?

stefan-schweiger:
Hmm I would suggest to avoid this in the future, we should mark the current constructor as deprecated and the new constructor should use only non optional parameters.

So you would have to use this in the future to create a BoxPlotItem.

var boxPlotItem = new BoxPlotItem(1, 12.5, 13.0, 14.5, 15.5, 16.0)
{
    Mean = 15.0,
    Outliers = new List<double> { 19.0, 19.5, 20.0 }
};

About the TrackerFormatString I see no other option besides a breaking change and wait for #414

objorke:
Yes, the BoxPlotItem should be changed to a reference type. And I agree, the optional parameters should be removed. I will approve it if you do it in this PR!

I don't agree that the parameters of the `TrackerFormatString` must be changed. The mean value could be added at the end.

Source/OxyPlot/Series/BoxPlotSeries.cs
```diff
@@ -242,6 +262,7 @@ public override TrackerHitResult GetNearestPoint(ScreenPoint point, bool interpo
                                 this.YAxis.GetValue(item.UpperWhisker),
                                 this.YAxis.GetValue(item.BoxTop),
                                 this.YAxis.GetValue(item.Median),
+                                this.YAxis.GetValue(item.Mean),
```
objorke: I think this should be added to the end, to avoid a breaking change.

stefan-schweiger:
OK I've changed the constructor accordingly, and squashed the commits.

I still think that mean should be in the middle of the TracerFormatString (as it's also where it will be displayed in the BoxPlot) but for now I've added it to the end. If #414 ever gets done this should be changed in my opinion.

objorke:
ok, thanks for this contribution!!

Issue #439, opened 2015-04-07
Update to Xamarin.Forms 1.4


objorke:

objorke:
See comment on #421

PR #441, opened 2015-04-07
Update Xamarin.Forms 1.4.1

tibel:
pull for issue #439


objorke:
don't know why the "in progress" label was added/deleted. I did nothing but look at the files changed...

objorke:
I think the change log should also be updated.

tibel:
@objorke Thanks. Changelog is updated.

PR #463, opened 2015-04-24
Code Analysis clean up

shoelzer:
This eliminates several Code Analysis warnings. I did it by changing code to satisfy the warning, but I can imagine reasonable arguments for leaving the code as-is and suppressing these warnings. Let me know what you think and I can update this PR.


objorke:
This looks OK to me. Thanks for fixing these issues!

PR #468, opened 2015-04-28
Enable checked math. This is a test. Will close shortly.

shoelzer:
I'm doing this so I can submit a PR on Github and have AppVeyor run all the build tests. Let's see if anything breaks.

### DIFF
--- a/Source/Examples/ExampleLibrary/ExampleLibrary.csproj
+++ b/Source/Examples/ExampleLibrary/ExampleLibrary.csproj
@@ -25,6 +25,7 @@
     <DefineConstants>TRACE;DEBUG;PCL</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
+    <CheckForOverflowUnderflow>true</CheckForOverflowUnderflow>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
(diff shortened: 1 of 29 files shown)


tibel:
Hope you don't want to enable these checks in release mode. This will kill performance totally.

shoelzer:
I was not planning to enable these checks in any build. I'm just curious to see if anything breaks.

Why do you think they would kill performance? I'm not saying you're wrong, but in my (limited) experience these checks are negligible except for tight loops doing math.

tibel:
Thanks for clarification.
Saying that the checks are negligible means we have to profile if this is really the case.

OxyPlot as the name implies is a plotting library so nearly every arithmetic calculation is part of rendering or some user interaction. Said that the design goal if FPS (frames per second) and responsiveness to user actions.

I personally know no library that has the checks enabled.

shoelzer:
Oh yes, I agree. I would definitely profile before proposing that these checks get turned on by default. Like I said, I'm just curious and poking around for bugs.

objorke:
isn't it easier to just run the tests locally? ;-)

shoelzer:
Perhaps. But then I would need to install NUnit and maybe other dependencies. I just haven't bothered doing that yet. I had noticed that that the tests run automatically when I push a PR so I just kept going with it.

I think I'm done with this experiment anyway. There are only three remaining test failures in this branch and I didn't find any real bugs. Kudos to both of you and the other OxyPlot devs!

Issue #501, opened 2015-06-12
Clear RangeColorAxis ranges


icerocker:
After adding ranges to a RangeColorAxis, there is no way of clearing the ranges. The solution is to create a new axis, then add the new ranges. It would be nice to have:
ClearRanges()

PR #502, opened 2015-06-16
Added method to clear RangeColorAxis ranges.

icerocker:


icerocker:
Ref. issue: Clear RangeColorAxis ranges #501

Issue #503, opened 2015-06-20
PlotView leak on iOS


TheAlmightyBob:
PlotViews don't appear to get garbage collected, due to the new-ish gesture recognizer never being detached.

The cause is a bit muddled in the example browser because the GraphViewController is also not getting GC'd, for other reasons.

PR #504, opened 2015-06-20
Fixed issue where the PanZoomGestureRecognizer was keeping the PlotVi…

TheAlmightyBob:
…ew from being garbage collected (#503). (also fixed similar issue with event handler in GraphViewController, so that the PlotView change could be observed using the example browser with HeapShot)


tibel:
Thanks

Issue #219, opened 2014-10-14
Tooltips no longer working in version 2014.1.319.1


GeertvanHorrik:
We are using a custom tooltip on the plot title. We accomplish this by using a custom Annotation and using src.SetToolTip on the IRenderContext.

When updating from 2014.1.277.1 => 2014.1.319.1 (or any new release), this seems to be broken.

tibel:
According to our guidelines https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTING.md we will close this issue. Feel free to re-open, when you have new information.

GeertvanHorrik:
Thanks. It's still on my todo list, but time and priorities are hard. Will re-open when I start on this.

objorke:
We could also add a `TitleTooltip` property on the `PlotModel`. Should be very easy to implement...

@GeertvanHorrik I don't like the implementation in `PlotTitleTooltipAnnotation` - does it triy to set the tooltip of all following "plot elements" (axes/annotations/series)? I think the current implementation where the tool tip is set for each plot element and cleared at the end is correct!

GeertvanHorrik:
Sorry for my late reply. This is just to let you know it's still on my list, just need to find some time.

objorke:
No problem, we have not marked this as high priority or part of the v2015.1 milestone.

GeertvanHorrik:
Going to try and fix this today. First updating to the latest oxyplot, then will check if the issue still exists.

GeertvanHorrik:
@objorke what I need is actually just the title to have a tooltip on a plot. Let me check if I can implement with what's currently available inside oxyplot.

GeertvanHorrik:
Using the ToolTip property on the Plot control seems to work. Then I can use a converter to set the right tooltip (not sure why I didn't use it before, probably wasn't available a long time ago when we started using OxyPlot).

The only disadvantage is that the location of the ToolTip sometimes is a bit off (but that's "ok" for me).

GeertvanHorrik:
The popup is positioned directly at the location you enter the control. So in a list of plots, when I move from 1 to another, the tooltip is there where I get into the control:

![image](https://cloud.githubusercontent.com/assets/1246444/8335504/5d88b15a-1a9e-11e5-9531-015206551c64.png)

GeertvanHorrik:
Created PR #508 as you can see, this issue can stay closed.

objorke:
[...]

PR #508, opened 2015-06-24
Added tooltips to the Performance Example for WPF

GeertvanHorrik:
Still working on this PR, but this is the result of #219


GeertvanHorrik:
See the image, the problem is that the tooltips are visible on the location where you enter the control (which looks really weird):

image

GeertvanHorrik:
When using ToolTipService.Placement="center", the tooltip is correctly aligned. I think using a custom callback is the solution.

GeertvanHorrik:
The readibility constraints are a bit too strict imho, I must force this:

```
            ToolTipService.AddToolTipOpeningHandler(this, (sender, args) =>
            {
                var toolTip = args.Source as ToolTip;
                if (toolTip != null)
                {
                    var mousePosition = Mouse.GetPosition(this);

                    toolTip.Placement = PlacementMode.RelativePoint;
                    toolTip.PlacementTarget = this;
                    toolTip.HorizontalOffset = mousePosition.X;
                    toolTip.VerticalOffset = mousePosition.Y;
                }
            });
```

into this:

```
            ToolTipService.AddToolTipOpeningHandler(
                this,
                (sender, args) =>
            {
                var toolTip = args.Source as ToolTip;
                if (toolTip != null)
                {
                    var mousePosition = Mouse.GetPosition(this);

                    toolTip.Placement = PlacementMode.RelativePoint;
                    toolTip.PlacementTarget = this;
                    toolTip.HorizontalOffset = mousePosition.X;
                    toolTip.VerticalOffset = mousePosition.Y;
                }
            });
```

objorke:
Why did you add this to the performance example? Wouldn't it be better to create a new example just demonstrating the tooltip capability?

GeertvanHorrik:
To check if I can reproduce. Yes, it might be better, but as with everyone, my time is limited but trying to do the right thing. Let me first finish the issue, then we can check if we need a separate example.

GeertvanHorrik:
@objorke Just added TitleToolTip. I need some help with rendering though. I see in other places (annotations, series, etc) that you set tooltip, render, then clear it again. However, I don't see how they are actually rendered / set to an element.

GeertvanHorrik:
@objorke never mind, found it. I need to set the tooltip before rendering, then it gets applied automatically.

GeertvanHorrik:
Also added the demo. Please review this PR.

GeertvanHorrik:
"Write-Error : Package 'gtksharp v2.12.26' did not install successfully: [ERROR]
 Running msiexec with /i "C:\Users\appveyor\AppData\Local\Temp\1\chocolatey\gtk
sharp\gtksharpInstall.msi" /quiet /qn /norestart  was not successful. Exit code
 was '1603'."

Is that something I caused?

GeertvanHorrik:
No, it seems this is for all users: #516

GeertvanHorrik:
Any updates on this?

Source/OxyPlot.Wpf/Plot.cs
```diff
@@ -153,6 +153,24 @@ private static void AppearanceChanged(DependencyObject d, DependencyPropertyChan
...
+        /// <param name="d">The d.</param>
+        /// <param name="e">The <see cref="System.Windows.DependencyPropertyChangedEventArgs" /> instance containing the event data.</param>
+        private static void TitleToolTipChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
+        {
+            var plot = (Plot)d;
+
+            var model = plot.ActualModel;
+            if (model != null)
+            {
+                model.TitleToolTip = e.NewValue as string;
```
tibel: For other properties this is handled in `OnAppearanceChanged`. Why the different handling?

GeertvanHorrik: I overlooked the SynchronizeProperties. Will update.

GeertvanHorrik:
Done.

tibel:
And update CHANGELOG.md, please.

GeertvanHorrik:
Changelog is updated.

PR #510, opened 2015-06-25
Fixes for the NuSpec so that they can be packaged on Mac OS

mattleibow:


mattleibow:
Temporarily closed as I am going to extend this to include the build script for the Xamarin CI server

mattleibow:
This script was run on CI and pushed to NuGet successfully.

objorke:
@mattleibow can the iOS/Android/Mac projects now be built on AppVeyor? Then we need to revert the changes to the nuspecs, right?

mattleibow:
Yes and no. AppVeyor can only build iOS and Android libraries - but not the Mac ones yet. Also, this script allows for building the projects and NuGet on Mac OS (at least the bits that can). Windows can handle both separator types, so there is no problem with the forward slashes.

objorke:
Thanks, then this change is OK.

objorke:
Thanks for fixing this!

Issue #115, opened 2014-08-14
Remove SL4 and NET35 builds


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10139](http://oxyplot.codeplex.com/workitem/10139)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-02-26 08:41
Support for XP ends April 8, 2014 

<a href="http://www.microsoft.com/en-us/windows/enterprise/endofsupport.aspx" rel="nofollow">http://www.microsoft.com/en-us/windows/enterprise/endofsupport.aspx</a>

I would like to end support for Silverlight 4, .NET 3.5 and .NET 4.0 in OxyPlot to make maintenance and build of the solution easier.

I suggest to set the date to December 31, 2014.

**[tibel](http://www.codeplex.com/site/users/view/tibel)** 2014-03-01 06:58
I would remove SL4 and NET35 now, as:

<ul>
<li>don't support PCL </li><li>are not part of nuget packages </li><li>newer versions are available that work on WinXP

</li></ul>
And only keep NET40 till the end of the year.

tibel:
I have removed SL4 and NET35.

Also I would like to remove SL5 and NET40 from the portable core project to prepare for future changes (e.g support INotifyCollectionChanged).

Portable profile 259 seems to be the must suitable (future proof):
![portable_profile259](https://cloud.githubusercontent.com/assets/387395/7630124/701615be-fa33-11e4-9a7c-998de787bd80.PNG)

objorke:
I agree SL5 support can be dropped from the PCL. But then we need to create a SL5 core project. Silverlight should still be supported!

tibel:
@objorke That was the idea.

objorke:
Can this be closed? I think removing OxyPlot_NET40.csproj should be a new issue, but I don't think this should be done yet - still many projects are using NET40...

Issue #19, opened 2014-08-14
The minimum value is not mentioned on the axis


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/9967](http://oxyplot.codeplex.com/workitem/9967)**

**[krafty](http://www.codeplex.com/site/users/view/krafty)** 2012-07-25 11:53
There should be a way to start showing minimum value in the data series on the axis. This is useful when there is no padding and we want to see the box bounds.

Cheers.

GeertvanHorrik:
Working on a PR.

GeertvanHorrik:
I have implemented this (including rounder so we don't get messy numbers). I think we should also add "snap" functionality so we only snap to logical containers. What do you think?

GeertvanHorrik:
The snapping works pretty well. For larger values it still snaps back to the closed value (so it's hard to move), but I can fix that as well.

GeertvanHorrik:
See PR #519

objorke:
I have added some example code (including some suggested new properties that are commented out)
The desired output should be the min/max labels 0.146 ang 9.858 on both axes (example I).

PR #519, opened 2015-06-29
Axis now always show the min and max values. Also implemented snapping to valid values.

GeertvanHorrik:


GeertvanHorrik:
Why the "in progress" label?

tibel:
Adding a `Timer` to the axis?

GeertvanHorrik:
Yes, the reason is that we need to snap after a while (when the user is scrolling, it should continue scrolling). It should then snap after x ms not to bother the user.

GeertvanHorrik:
Otherwise the snapping will be too aggressive.

tibel:
I thought that `PlotController` is for handling user interaction.

GeertvanHorrik:
But it should always snap (even when set programmatically / etc). I am not really familiar with the controllers, so I might need some help. If you think there is a better solution, let me know, but I used a system timer to prevent any UI elements inside the axis.

GeertvanHorrik:
Any updates on this?

tibel:
Build server is working again.
With the change of the PCL core library to profile 259 you have no access to `Timer` class (see
https://forums.xamarin.com/discussion/17227/timer-in-portable-class-library).

GeertvanHorrik:
Dear lord, no timer... Ok, will need a different way to solve this issue.

GeertvanHorrik:
Now I remember, I implemented a timer in Catel for PCL:

https://github.com/Catel/Catel/blob/e8c26d4a8a7d09e2799f24f7f547650c4bf79433/src/Catel.Core/Catel.Core.Shared/Threading/Timer.cs

Can I put it in?

tibel:
Yes, we have to include some internal Timer for PCL.

I still think we should not add a Timer to the axis and  would prefer one Timer per PlotModel or PlotController to handle the snapping. (Just my thoughts, no concrete idea how to implement this)

GeertvanHorrik:
Let me start by adding a timer, then implement the way it currently is. Then we can always move it, right?

tibel:
Sure

GeertvanHorrik:
How can I inform a PlotModel that it must be updated from within an axis? We have the AxisChanged / TransformChanged events, but they are never subscribed to?

GeertvanHorrik:
Extracting the snapping logic so it's not implemented inside Axis (but as an extension method).

GeertvanHorrik:
Trying to move this to the controller. Requires additional events / methods to controller, but will keep it as a non-breaking change (feature additions only).

GeertvanHorrik:
I have refactored the controller, but it's too complex, the controller doesn't have all events (for example, Pan, Zoom, etc). I think we really need to add this to the PlotModel instead.

tibel:
I think we should ask @objorke for his input.

GeertvanHorrik:
Now the collection changed stuff is added, I will finish this one.

objorke:
I think the "in progress" tag was intended for "review in progress". First time I used a tag on a PR :-)

GeertvanHorrik:
Ah. I thought it meant in progress by developer. They are set automatically, right?

objorke:
Can the snapping feature be moved to a separate PR?

objorke:
Please remove #region directives. http://docs.oxyplot.org/en/latest/contribute/coding-style.html

objorke:
Can we have a flag to control whether the min/max values can be shown or not?

GeertvanHorrik:
Not really. The reason for that is that for the min / max to be meaningful, we must snap. Otherwise you will get extremely weird min / max 3.0000000000045349 etc.

objorke:
Please update the change log :-)

GeertvanHorrik:
I think I was working on SnapSettings. I wanted to put them on Axis (so you can determine per axis how and if to snap and whether to show min / max values).

objorke:
I see, I will test this later!
I am also a bit concerned about the timer being introduced here... Is it really needed?

GeertvanHorrik:
Yes, we need a timer, unfortunately. I can get away with a Task.Delay, but that is uglier than a timer.

GeertvanHorrik:
Sorry this is going to be so big :-(

GeertvanHorrik:
I think it's ready for review.

GeertvanHorrik:
I don't understand why it's failing, @objorke , can you take a look?

objorke:
There are visual diffs after these changes:

Expected (baseline):
pngexportertests_exportwithresolution_192dpi

Actual:
pngexportertests_exportwithresolution_192dpi

Can this feature be implemented without doing a breaking change? Some users may want the current behavior (with 'padding' on the x-axis).

GeertvanHorrik:
I thought I turned off the feature by default. Checking out why it's not. Will definitely keep it the way it was, so you can enable snapping if you want.

GeertvanHorrik:
I couldn't get my unit tests to succeed. I undoed all my changes 1 by 1 to check what might be causing it, but none of them made the unit tests succeed. So decided to do a fresh branch from develop, but these unit tests also fail?

GeertvanHorrik:
Found the issue, it was the refactoring of UpdateActualMinMax, where the calculation for the actual minimum requires the ActualMaximum to be set first.

MerlinDuChaos:
When will this be available in the main branch ?  I'd need it for my project, this is a great addition ! +1

objorke:
I can review this next week!

GeertvanHorrik:
Awesome, thanks. Let me know if there is anything I can help you with in the meantime.

GeertvanHorrik:
@objorke Hi Oystein, did you find time to review this?

objorke:
Sorry for the delay, will review soon!

objorke:
will try to find time in the weekend!

Source/OxyPlot/Graphics/Snapping.cs
```diff
@@ -0,0 +1,39 @@
...
+//   Snapping data.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot
+{
+    /// <summary>
+    /// Snapping data.
+    /// </summary>
+    public class Snapping
```
objorke: would it be better to add simple properties `IsSnappingEnabled` and `SnappingPrecision` to the axis? I am concerned about these snapping options being a mutable reference type. Also, I don't think it works well with the WPF axis "wrappers" - no change notifications if you change an `IsEnabled` property

objorke:
I think there should be better definition for what snapping is doing.

Source/OxyPlot/PlotModel/PlotModel.cs
```diff
@@ -193,6 +193,16 @@ public enum TitleHorizontalAlignment
...
     {
         /// <summary>
+        /// The snap timer.
+        /// </summary>
+        private readonly Timer snapTimer;
+
+        /// <summary>
+        /// Gets or sets whether the axis is currently snapping.
+        /// </summary>
+        private bool isSnapping;
```
objorke: is this needed? should rather check if any axis has enabled snapping

Source/OxyPlot/PlotModel/PlotModel.cs
```diff
@@ -193,6 +193,16 @@ public enum TitleHorizontalAlignment
     public partial class PlotModel : Model, IPlotModel
     {
         /// <summary>
+        /// The snap timer.
+        /// </summary>
+        private readonly Timer snapTimer;
```
objorke: I don't like adding a timer. I think we discussed this before, must test. How will this work if you are printing the `PlotModel`? Should you wait a few timer intervals before printing? :-)

Source/OxyPlot/Utilities/Timer.cs
```diff
@@ -0,0 +1,250 @@
...
+    /// </summary>
+    /// <param name="state">
+    /// The state.
+    /// </param>
+    public delegate void TimerCallback(object state);
+
+    /// <summary>
+    /// Timer for WinRT since WinRT only provides the DispatcherTimer which cannot be used outside the UI thread.
+    /// </summary>
+    public class Timer : IDisposable
```
objorke: Cool, this is original code - not taken from anywhere, right?

objorke: But, I still hope we don't need a timer in this library. In my opinion timers should only be used in the platform specific projects (in the plot views).

Source/OxyPlot/OxyPlot_SL5.csproj
```diff
@@ -30,7 +30,7 @@
     <Optimize>false</Optimize>
     <OutputPath>Bin\Debug\SL5\</OutputPath>
     <IntermediateOutputPath>obj\Debug\SL5\</IntermediateOutputPath>
-    <DefineConstants>DEBUG;TRACE;SILVERLIGHT</DefineConstants>
+    <DefineConstants>TRACE;DEBUG;SILVERLIGHT;SL5</DefineConstants>
```
objorke: I think we can remove support for SL4. Will handle it in #115/#481

objorke:
There should be on PR for the "show min/max values" issue #19. This one should be easy to approve.
There should be another PR for the "snapping" feature and a corresponding issue. 
@GeertvanHorrik Can you create the issue and split the PR?
The 29 commits must also be squashed. I would like to see 1 commit on each PR/branch!
Sorry that this review has been taking so long.

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ All notable changes to this project will be documented in this file.
 - Added capability to display mean value to BoxPlot (#440)
 - Added TitleToolTip to PlotModel (#508)
 - Expose PlotElement's TextColor property on WPF.Axes.Axis (#452)
+- Added snapping feature to PlotModel and Axis. This also ensures that the minimum and maximum values 
+  on the axis are always displayed (#19, #519). This feature is disabled by default, but can be enabled
```
objorke: I think we should try to only reference issues, not PRs

Source/OxyPlot/Graphics/Snapping.cs
GeertvanHorrik: The reason I used an object is not to cause breaking changes in the future. If you want new snapping data in the future, you can add it without a breaking change or making a mess. But happy to make it direct properties (they were direct properties in the first place).

Source/OxyPlot/OxyPlot_SL5.csproj
GeertvanHorrik: But even then, WP80 / WP80SL also define the SILVERLIGHT definition so I still think SL5 is a good define constant.

Source/OxyPlot/PlotModel/PlotModel.cs
GeertvanHorrik: I know, I wish there was another (easy) way. But because the timer is inside the model, not inside the control, it will also work during printing. And... the timer will only kick in when you pan / zoom / navigate anyway.

Source/OxyPlot/PlotModel/PlotModel.cs
GeertvanHorrik: This is required. Snapping might be enabled, but the axis might not be snapping *at the moment*. We need to prevent that because of the snapping "move", we don't spawn another snapping call (and cause unnecessary calls).

Source/OxyPlot/Utilities/Timer.cs
GeertvanHorrik: Not original code, it comes from Catel (I wrote it, have to fight all the platform differences there as well). It's MIT licensed so you can do whatever you want with it:

https://github.com/Catel/Catel/blob/develop/src/Catel.Core/Catel.Core.Shared/Threading/Timer.cs

GeertvanHorrik:
First of all, thanks for reviewing. I know we are all busy so don't worry too much about it. I will squash everything, but about the 2 separate PR:

I have created the "show min / max values on the scale". That was the initial issue I was trying to solve. Then I saw a *lot* of nonsense values like 0.000000099, etc. Then I decided: I can round up these values and lie about them or I can implement a snapping (that's what users actually want IMHO). So I wanted to snap. But, unfortunately this needed a timer (because when you are panning, you want to snap *once* you are ready panning, etc).

The bottom line is that I am not sure if one can go without the other. How do you want to proceed with this?

objorke:
Instead of a timer, can't there be a "panning complete" event?

objorke:
Sorry for the delays and being very picky on this PR... There is something I don't like with this design, but I don't have a clear picture of how it should be yet. Maybe the snapping functionality should be the responsibility of the Views, the `PlotController` or the `PanManipulator`. I need to look into the code...
Any others with opinions on this implementation? @tibel?

GeertvanHorrik:
_Np, I am happy that you are picky (we all need to be, we want a good product, right :-) )_

Anyway, I tried implementing this inside the controller as well, but when is panning assumed completed? When I hold down the left arrow, when I hit the left arrow 7 times fast in burst mode, etc.

The only thing I can think of is to move the timer to the controller instead of the axis. Then if someone modifies the plot manually, they need to call .Snap() manually, which is fine by me.

objorke:
Right, it is difficult to know when you are done panning with the keyboard... :-) But isn't it ok to snap after each keystroke? The keyboard panning could use the tick interval lengths as "delta values"!

Still no example included in this PR? I created the following model in the example library:
```csharp
        [Example("Axis snapping")]
        public static PlotModel AxisSnapping()
        {
            var plotModel1 = new PlotModel { Title = "Axis snapping" };
            var axis1 = new LinearAxis { Position = AxisPosition.Left };
            var axis2 = new LinearAxis { Position = AxisPosition.Bottom };
            axis1.Snapping.IsEnabled = true;
            axis2.Snapping.IsEnabled = true;
            plotModel1.Axes.Add(axis1);
            plotModel1.Axes.Add(axis2);
            return plotModel1;
        }
```

Regarding the Snapping class, I think it would be nicer if I could write
```csharp
plotModel1.Axes.Add(new LinearAxis { Position = AxisPosition.Left, IsSnappingEnabled = true });
```

I notice that after panning, the axes show numbers like 46, 66, 86 when snapping is enabled, but nice numbers like 40, 60, 80 when snapping is disabled. Was this intended?

What are good values for `Precision` (not auto) for the example above?

Is it possible that we can support this feature by adding some extension points in the plot controller? Virtual methods or events? I am still not convinced that it is a good idea to add this feature in the general library, and particularly not on the model.

GeertvanHorrik:
So to summarize:

1. You want the snapping class to be removed and just be a property directly on the axis?
2. Precision is automatically calculated based on the values the axis is displaying in the current zoom factor. I made it overridable in case it was not behaving as expected so it can always be "hotfixed" by the end-users

What if I remove all the timing stuff and just add the Snap function. Then it's the responsibility of the end-developer to implement this. I will implement it with a timer on my side.

objorke:
I am concerned about the maintainability of this feature in the core library, it is making the "update" pass more complicated.
My preference would be to add some virtual methods or events in the core, then include an example showing how to do snapping. It would be great if such an example was created in the portable example library.
Yes, in my opinion the timing stuff should be removed. Snapping should happen immediately when panning (also by keys) has completed.

GeertvanHorrik:
That's why I don't want anything inside the update pass. I just want a Snap() method on the axis so we can manually call it, that's all. I think that can be put into the core?

objorke:
I don't quite understand... Can we call it something more general than `Snap`?

GeertvanHorrik:
Of course. Anything you want (I will probably put this into my custom usage of OxyPlot anyway). The only thing I needed in the first place is that the outer values on an axis are visible (by default, but also after dragging). Since you can have some nonsense value like 3.00000000000000e2 or something like that which make no sense to an end-user, I wanted to snap such values. That's why I called it snapping in the first place, but other names such as ResizeViewPortToSensibleValues or whatever you prefer.

objorke:
Sorry, I am still not happy with the name - I wanted a general method or event where you could extend the custom functionality (like snapping or something else). But maybe I don't understand correctly... Where should this method or event live? Where should it be called from?

I looked at #19 again, I think snapping is not part of that issue. krafty wanted to see the minimum value of the data, not a 'nice' number.

I see you comment from Aug 20, but I still think #19 and snapping are different features. What about creating a new issue about snapping with clear requirements about what it should do? I understand it as
- snapping: adjust the actual minimum and maximum of an axis to values at major ticks (at 'nice' numbers)
- do the snapping (adjustment) when minimum and maximum has not been defined 
- do the snapping after the axis has been panned or zoomed

I hope we can find a good solution soon!

GeertvanHorrik:
Yes, you want to show the display values of the min and max. But does this number make sense?

3.9e23 or would the end user (and #19) want to see 4 in that case? That's why I introduced snapping. Snapping to logical values that make sense for the values that are actually representing the axis.

If you can come up with a nice name, I am happy to use that as well. Maybe TuneAxis() which is virtual? By default it doesn't do anything, but you can override to implement features like snap.

GeertvanHorrik:
So..... the verdict is? I am happy to close this one. I will implement it in my own solution instead, I Think it's too complex to be implemented in the core.

objorke:
Let us close this PR if it is OK for you, I am not comfortable merging this code yet. [Jeremy Miller](http://jeremydmiller.com/) has a good post about such issues, see the section [Taking pull requests](http://jeremydmiller.com/2015/10/15/thoughts-on-running-an-oss-project/).

I am willing to add "extension points" where you can implement functionality like snapping, but let us create a separate issue/PR for that. I think the important question is *where* the extension point (virtual method, event) should be called from.

I have started looking into #19, will continue discussion on min/max values there (I think 3.9e23 is the correct value to show!)

PR #524, opened 2015-07-06
Update TargetFrameworkVersion to v4.5

juergenhoetzel:
Fixes build error:

     GraphicsRenderContext.cs(24,42): error CS0012: The type `System.Object' is defined in an assembly that is not referenced. Consider adding a reference to assembly `System.Runtime, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'

because OxyPlot already targeting v4.5 since this commit 096c0cf833f969af6f00d01f602f126048e8497e


sevoku:
Sorry, I forgot to answer here. It is not enough to change the projects TargetFrameworkVersion setting. The output paths have to be updated, too. And some people could still need a NET40 configuration. So I've created a new PR #527 which includes everything required.

PR #531, opened 2015-07-08
Updated the pack scripts to use the mono NuGet

mattleibow:
- this has "more" optimised support for non-Windows platforms
 - use the `-BasePath` argument to avoid any working-directory or path issues

This has no theoretical change as I believe the default `BasePath` is actually the directory of the `.nuspec`, but I know that this does not always work - as can be seen by the fact that although it finds the assemblies (no not-found errors), it doesn't actually include them in the `.nupkg` (not in the uploaded NuGet)

### DIFF
--- a/Build/build-xamarin.sh
+++ b/Build/build-xamarin.sh
@@ -45,7 +45,7 @@ ls -al ../Output/Xamarin.Mac/OxyPlot*
 
 
 # Start the packaging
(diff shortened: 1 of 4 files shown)


mattleibow:
You can see this here:
https://www.nuget.org/packages/OxyPlot.Xamarin.Android/2015.1.928-alpha

and fixed here:
https://www.nuget.org/packages/OxyPlot.Xamarin.Android/2015.1.934-alpha

Issue #525, opened 2015-07-06
Oxyplot.GtkSharp: Port to gtk-sharp-3


juergenhoetzel:
are there any plans to update to gtk-sharp-3 from gtk-sharp-2?

See: https://github.com/fslaborg/FSharp.Charting/issues/76

sevoku:
I made a gtk3 port some months ago, but I didn't finish it because of the project structure complexity (I had simply rewritten the Gtk project). The major difference between Gtk2 and Gtk3 is drawing (no Expose event) and the sizing of widgets.

@objorke: what would be the best way to solve this?

1. It must be a separate project/solution file, would OxyPlot.GtkSharp3 be ok?
2. Should Oxyplot.GtkSharp be renamed (e.g. Oxyplot.GtkSharp2, breaking change!)
3. It would make sense to share code between both versions, since the differences are not very big. The easiest (dirty) solution would be to use an #ifdef. But a more cleaner solution would be to move the most of the code to a base class and create version specific classes (Oxyplot.GtkSharp2, Oxyplot.GtkSharp3).

sevoku:
I've pushed a WIP branch to github at https://github.com/sevoku/oxyplot/tree/add-gtk3-support (to see the changes without #527): https://github.com/sevoku/oxyplot/compare/fix-gtk-targetfw...sevoku:add-gtk3-support). The important commit is bfe0408995ee23136f67a2ab773d0075c5156a81.

Description:
* I've added an OxyPlot.GtkSharp3 project and solution (incl. Gtk3 ExampleBrowser and Demo)
* Gtk2 and Gtk3 share the same Namespace ```Oxyplot.GtkSharp```. This is good for two reasons:
  1. It is not possible to target Gtk2 and Gtk3 at the same time, so there is no reason for separate namespaces. If the user wants to support both versions in one assembly, he/she is forced to dynamically load the GtkSharp assembly and has to ensure not to initialize both Gtk versions at the same time.
  2. Its easy to switch Gtk versions by simply replacing the reference (using same client code). So supporting both versions needs only project configuration based (and no code) changes.
[...]

PR #534, opened 2015-07-10
Add Gtk3 support

sevoku:
This PR adds Gtk3 support with all required project files and AppVeyor integration.

Additionally it includes some optimizations to GtkSharp, the Gtk2 project files and the example browser.

Description (original at #525):
* I've added an OxyPlot.GtkSharp3 project and solution (incl. Gtk3 ExampleBrowser and Demo)
* Gtk2 and Gtk3 share the same Namespace ```Oxyplot.GtkSharp```. This is good for two reasons:
  1. It is not possible to target Gtk2 and Gtk3 at the same time, so there is no reason for separate namespaces. If the user wants to support both versions in one assembly, he/she is forced to dynamically load the GtkSharp assembly and has to ensure not to initialize both Gtk versions at the same time.
  2. Its easy to switch Gtk versions by simply replacing the reference (using same client code). So supporting both versions needs only project configuration based (and no code) changes.
* Gtk3 handles drawing slightly different. There is a Draw event (instead of Expose) which sets up the cairo context correctly (with clipping etc.). So I've made the PlotView partial (https://github.com/oxyplot/oxyplot/commit/bfe0408995ee23136f67a2ab773d0075c5156a81#diff-e7a1cde5d4bba13c0eae64a43b9529cbR25) and added 2 cs files with version specific implementations (https://github.com/oxyplot/oxyplot/commit/bfe0408995ee23136f67a2ab773d0075c5156a81#diff-0da024ba35c6466044212f263db03c7cR1, https://github.com/oxyplot/oxyplot/commit/bfe0408995ee23136f67a2ab773d0075c5156a81#diff-0fd751062f85e710d2bde53fae764396R1). Both call the ```DrawPlot``` method (https://github.com/oxyplot/oxyplot/commit/bfe0408995ee23136f67a2ab773d0075c5156a81#diff-e7a1cde5d4bba13c0eae64a43b9529cbR419) which previously was the Exposed handler.
* The Gtk2 area dimensions parameter (from the Exposed event) should not be used for drawing, since OxyPlot has no support for partial redraws. I've patched the drawing method to always use the full widget allocation in both versions (Gtk2/3).

TODO: update changelog

Please don't merge without review! Project files may need additional modifications to compile correctly on Windows, since there may be some problems with the references if both Gtk versions are installed. Lets see what AppVeyor says...


objorke:
This looks good! Thanks for adding the Gtk3 support!!

sevoku:
Thanks, I've quickly pushed 76c84b9304b6c04d4f7eeb86e0801a005c9547c2 to fix gtk installer caching.

PR #541, opened 2015-07-20
Expose PlotElement's TextColor property on WPF.Axes.Axis

### LINKED ISSUE #452: Expose PlotElement's TextColor property on WPF.Axes.Axis
The `TextColor` property can currently only be set in code via a child of `PlotModel.PlotElement`, but not in XAML via `Wpf.Axes.Axis`.

Would be nice to be able to set the `TextColor` property in XAML.

svendu:
This should resolve  #452


objorke:
Excellent, thanks!

Issue #542, opened 2015-07-21
Should Render Exceptions be recorded like UpdateExceptions?


pltaylor:
I am having an issue with an occasional render exception that would be made easier if they were recorded like UpdateExceptions so they could be reviewed using something like the already implemented

```C#
var exception = plotModel.GetLastUpdateException();
```

PR #543, opened 2015-07-21
RenderException Handling

pltaylor:
Handle RenderExceptions in a way that can be accessed easily.  Addresses #542

### DIFF
--- a/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
+++ b/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
@@ -44,7 +44,7 @@ namespace OxyPlot
             {
                 try
                 {
-                    if (this.updateException != null)
+                    if (this.lastPlotException != null)
                     {
                         // There was an exception during plot model update. 
                         // This could happen when OxyPlot is given invalid input data. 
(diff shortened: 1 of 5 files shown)


Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
```diff
@@ -127,6 +127,7 @@ protected virtual void RenderOverride(IRenderContext rc, double width, double he
                             "An exception of type {0} was thrown when rendering the plot model.\r\n{1}",
                             exception.GetType(),
                             exception.GetBaseException().StackTrace);
+                    this.updateException = exception;
```
objorke: Wouldn't it better to create a new field `renderException`? And a `GetLastRenderException()` method? Or rename to `lastException` and `GetLastException()`?

objorke:
Please also update the change log with a reference to the issue. :-)

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
pltaylor: I will glad to do either.  Which way would you prefer?

pltaylor:
This changelog or is there one in code that I should reference?

svendu:
The file to update, is called CHANGELOG.md. You should find it in the root of your oxyplot sources

Issue #533, opened 2015-07-08
Nuget Xamarin.iOS and Xamarin.Android packages corrupt


Gaisuru:
When trying to install, it give the following error.
Attempting to resolve dependency 'OxyPlot.Core (= 2015.1.934-alpha)'.
'Default' tag requires a nonempty 'Extension' attribute. Line 1, position 378.

I assume it is the same as issue https://github.com/oxyplot/oxyplot/issues/184

tibel:
@mattleibow Seems there is still an issue with the Xamarin NuGet packages.

PR #545, opened 2015-07-22
Fix for Xamarin platform NuGets in VS

mattleibow:
For some reason, only the Xamarin platforms struggle to install extension-less files. The same logic applies when installing to any other project type.


mattleibow:
This fixes #533

PR #549, opened 2015-07-31
Fixed issue where a tap would transition PanZoomGestureRecognizer to …

TheAlmightyBob:
…Ended rather than Failed, firing a TouchCompleted for a touch that never started (and potentially interfering with other gesture recognizers that the application might have added, e.g. a UITapGestureRecognizer)


TheAlmightyBob:
(I didn't create an issue for this because it's a bit subtle and there are workarounds for the symptoms... but the previous implementation was still incorrect)

PR #558, opened 2015-08-20
Fix Issue #256

br111an:
OxyPlot.Series.HeatMapSeries was calculating a nonsense value when
double.NaN was present. This is issue #256. Fix interpolation to avoid calculating nonsense.
Labels use integer coordinates so a non-interpolating method was
introduced to get values for integer coordinates.

PR #560, opened 2015-08-21
Update CHANGELOG.md

br111an:
Issue 256.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 ### Added
+- OxyPlot.Series.HeatMapSeries.GetValueForLabels (#256)
 - Created a new OxyPlot.Mobile NuGet package to combine the mobile platforms into a single package. (#362)
 - Support for XWT (#295)
 - TwoColorAreaSeries (#299)
@@ -18,6 +19,7 @@ All notable changes to this project will be documented in this file.
 - Expose PlotElement's TextColor property on WPF.Axes.Axis (#452)
 
 ### Changed
+- Changed OxyPlot.Series.HeatMapSeries.GetValue (#256)
 - Renamed OxyPlot.WindowsUniversal to OxyPlot.Windows (#242)
 - Changed OxyPlot.Xamarin.Forms to require OxyPlot.Mobile dependency instead of each separate NuGet. (#362)
 - Renamed OxyPlot.XamarinIOS to OxyPlot.MonoTouch (#327)


CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 ### Added
+- OxyPlot.Series.HeatMapSeries.GetValueForLabels (#256)
```
objorke: GetValueForLabels is not a public member - and no functionality added? then it should not be necessary to add it here

CHANGELOG.md
```diff
@@ -18,6 +19,7 @@ All notable changes to this project will be documented in this file.
 - Expose PlotElement's TextColor property on WPF.Axes.Axis (#452)
 
 ### Changed
+- Changed OxyPlot.Series.HeatMapSeries.GetValue (#256)
```
objorke: I think this can be improved. What was changed? I suggest "Fix NaN bug in HeatMapSeries displayed values" or something like that

objorke:
Please include in #558

CHANGELOG.md
br111an: Both GetValue and GetValueForLabel are "private static double" methods. The intention is to make them similar. 

GetValueForLabels should be renamed to GetValueForLabels and that will be reflected when I re-sync the edits.

GetValueForLabel is necessary because GetValue returns double.NaN at some boundaries. These boundaries are almost non-detectable in a continuous dimension (that is to say, an interpolated dimension) used for mouse movements and a color spectrum. Unfortunately that boundary is at whole number coordinates. Eliminate the possibility of a calculated double.NaN at some boundaries by introducing GetValueForLabel. Double.NaN is still possible in GetValueForLabel but those values are not calculated, just read from memory. Since RenderLabels always uses integer coordinates, GetValueForLabel receives integers for type-safety. GetValueForLabel is simple (more compute efficient) than GetValue.

br111an: The previous should have been: GetValueForLabels should be renamed to GetValueForLabel.

This will be in the next sync.

br111an: Please disregard the two previous comments.

I found a superior fix for GetValue that eliminates the need to introduce GetValueForLabel.

CHANGELOG.md
br111an: Okay, the revised note is in the "Fixed" section.

br111an:
There is no local copy of CHANGELOG.md when I search via Windows Explorer
on my local clone. I don't know how to include edits to CHANGELOG.md in
pull request #558.

Please advise.

objorke:
did you branch from develop? the change log should be there:
https://github.com/oxyplot/oxyplot/blob/develop/CHANGELOG.md

br111an:
The fork was from "develop". Here are the search results showing there is no CHANGELOG.md file.

http://anony.ws/image/DCUG

br111an:
Since the change log was not found on my hard drive, I have been editing the change log on the github.com website. 

https://github.com/oxyplot/oxyplot/edit/develop/CHANGELOG.md

Here's the latest patch to the "Fixed" section.
https://github.com/oxyplot/oxyplot/compare/develop...br111an:patch-6

Issue #256, opened 2014-11-19
HeatMapSeries value is wrong if to the left of NaN


H2ONaCl:
On a HeatMapSeries, the displayed value is sometimes wrong, even if the color is correct. The wrong value is 0.0. The value that is wrong is always to the left of a heat map cell that contains Double.NaN. All values are supposed to be 4.7 or NaN in this example.
![heatmapbug](https://cloud.githubusercontent.com/assets/5573922/5501500/04ec926c-872a-11e4-9ea5-5c83a2874a1d.png)

    internal class DemoViewModel
    {
        public PlotModel PlotModelInstance { get; private set; } // WPF binds here

        internal DemoViewModel()
        {
            this.PlotModelInstance = new MyHeatMap();
        }

        private class MyHeatMap : PlotModel
        {
            internal MyHeatMap()
            {
                const int xDimension = 3; const int yDimension = 4;

                var colorLegend = new LinearColorAxis();
                colorLegend.InvalidNumberColor = OxyColors.Transparent;
                colorLegend.Position = AxisPosition.Right;
                this.Axes.Add(colorLegend);

                var xAxis = new CategoryAxis();
                xAxis.Position = AxisPosition.Bottom;
                xAxis.Labels.Add("first"); 
                xAxis.Labels.Add("second");
                xAxis.Labels.Add("3rd"); 
                this.Axes.Add(xAxis);

                var yAxis = new CategoryAxis();
                yAxis.Position = AxisPosition.Left;
                for (int y = 0; y < yDimension; y++)
                    yAxis.Labels.Add("category " + y);
                this.Axes.Add(yAxis);
                
                var series = new HeatMapSeries();
                series.Interpolate = false;
                series.LabelFontSize = 0.4d;
                series.LabelFormatString = "0.0";
                series.X0 = 0; series.X1 = xDimension - 1;
                series.Y0 = 0; series.Y1 = yDimension - 1;

                double datum = 4.7d;
                series.Data = new double[xDimension, yDimension];
                for (int x = 0; x < xDimension; x++)
                    for (int y = 0; y < yDimension; y++)
                    {
                        series.Data[x, y] = datum;
                    }
                series.Data[1, 0] = double.NaN;
                series.Data[2, 2] = double.NaN;
                this.Series.Add(series);
            }
        }
    }

objorke:
Fixed in #572

PR #572, opened 2015-09-08
Fix issue #256

br111an:
HeatMapSeries.GetValue returns double.NaN instead of calculating a
nonsense value in proximity to double.NaN.

Introduce tests that generate a flat data heat map to better illustrate
the bug that was fixed.


objorke:
This looks good, thanks for contributing!!

Issue #587, opened 2015-09-17
DateTimeAxis.CalculateActualInterval() infinite loop


XiaoHippo:
I recently came across a situation that caused infinite loop in `DateTimeAxis.CalculateActualInterval()`. I can repro this (on device only) with the latest NuGet packages:
`OxyPlot.Core 2015.1.999`
`OxyPlot.Xamarin.iOS 2015.1.999`

When I looked at the decompiled code, `DateTimeAxis.CalculateActualInterval()` was called with the following values:

```C#
// availableSize = 293.5
// maxIntervalSize = 60
// base.ActualMinimum = 39374.6666666667
// base.ActualMaximum = 42185.5

double interval = array[0];
int num2 = Math.Max((int)(availableSize / maxIntervalSize), 2);
while (num / interval >= (double) num2) {
    double num3 = array.FirstOrDefault((double i) => i > interval);
    if (Math.Abs(num3) < 4.94065645841247E-324) {
        num3 = interval * 2; // Won't run if num3 == 0
    }
    interval = num3;
}
```

In that loop, `interval` eventually was set to `365.25`, so it caused `array.FirstOrDefaul()` to return `0`, but the `if` statement got skipped over, causing the infinite loop.

So essentially, the problem is this code returns `false` while mathematically it should be `true`:
```C#
double num = default(double);
var result = (Math.Abs(num) < 4.94065645841247E-324);
```

It looks like a `double` precision problem. This only started to happen on device (not on simulator) after I upgraded to iOS 9.

(One idea is to change `<` to `<=` because it seemed to have worked in my case.)

objorke:
Thanks for reporting this. Can we make a unit test that reproduce the error? Otherwise just make the condition more robust as you suggest.

XiaoHippo:
OK, I can create a PR, if you could fix the broken link on the home page that points to the document on how to contribute?

objorke:
Thanks for the notice. I have updated the oxyplot.org page. The link should be
http://docs.oxyplot.org/en/latest/contribute/

PR #588, opened 2015-09-17
#587: Fixing a double comparison issue causing infinite loop

XiaoHippo:
So I went the easy way ;) This seems to be more of a data type issue than a issue with OxyPlot's logic. Plus imho that method needs some refactoring in order to extract the logic for testing.


objorke:
Please update the change log
Then squash the two commits (commit, rebase interactively, push -f)

tibel:
Thanks

PR #602, opened 2015-10-01
MonoMac support

jeremyiverson:
This was accomplished with these steps:

1. Copy the Xamarin.Mac implementation
2. Reference MonoMac instead of Xamarin.Mac assemblies
3. Change namespaces to MonoMac namespaces
4. Change CGRect and CGPoint to RectangleF and PointF


objorke:
cool!

Issue #599, opened 2015-09-27
Processor architecture mismatch in GtkSharp projects


objorke:
The build log shows warnings like:

C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.targets(1819,5): warning MSB3270: There was a mismatch between the processor architecture of the project being built "MSIL" and the processor architecture of the reference "atk-sharp, Version=3.0.0.0, Culture=neutral, PublicKeyToken=35e10195dab3c99f", "x86". This mismatch may cause runtime failures. Please consider changing the targeted processor architecture of your project through the Configuration Manager so as to align the processor architectures between your project and references, or take a dependency on references with a processor architecture that matches the targeted processor architecture of your project. [C:\projects\oxyplot\Source\OxyPlot.GtkSharp\OxyPlot.GtkSharp3.csproj]
C:\Program Files (x86)\MSBuild\14.0\bin\Microsoft.Common.CurrentVersion.

objorke:
is it ok to change the OxyPlot.GtkSharp3 projects to x86? this fixes the build warnings...
why are there no warnings on GtkSharp2?

tibel:
I think this is ok for now, as GTKSharp3 is still __beta__ (see https://github.com/mono/gtk-sharp/releases). They only released x86 for Windows yet.

PR #610, opened 2015-10-03
Change GTKSharp3 project to x86

objorke:
to solve #599


objorke:
Not sure if this is correct... See comment on issue. But I get rid of the build warnings...

PR #612, opened 2015-10-04
Include Xamarin projects in AppVeyor build

objorke:
First try on #274


objorke:
Failing with
```
appveyor RegisterXamarinLicense -Email %xamarin_email% -Password %xamarin_password% -Product Android
Error: -Password parameter is not specified.
 
RegisterXamarinLicense options:
 
  -Email      - Required. Xamarin account email address.
  -Password   - Required. Xamarin account password.
  -Product    - Requried. Product to activate: "Android" or "iOS".
```

the environment variable looks correct...

tibel:
see http://www.appveyor.com/docs/how-to/secure-files
Check the last sentence on pull request builds. Maybe it is that simple?

objorke:
Right, this makes sense. But this means it is not possible to include the Xamarin projects in the default build, I think...

It should be possible to set the Xamarin license in the ci.appveyor.com settings instead! Then AppVeyor should handle register and unregister license...

objorke:
I added http://help.appveyor.com/discussions/questions/1728-register-xamarin-license-for-pull-requests

tibel:
For the moment I think it would be ok to have Xamarin only in real builds (not for pull requests).
Not ideal but at least better than the current solution. A step in the right direction :+1:

objorke:
Yes, I agree that is an acceptable but not ideal solution.
I can try `IF DEFINED xamarin_password ...`, or do you have a better solution?

objorke:
I will try a merge...

tibel:
Maybe we should remove `os: Visual Studio 2015` from  `appveyor.yml` to use the latest build image (see http://www.appveyor.com/updates and http://www.appveyor.com/blog).
If this does not help you should ask AppVeyor directly for Xamarin support issues.

objorke:
Thanks, I will try this!

objorke:
It still shows `Android SDK platform for API Level 10 is not installed`.
I can try updating OxyPlot.Xamarin.Android to Level 15, the same as OxyPlot.Xamarin.Forms.Platform.Android.
But this means we loose the support for Android v2.3, right?

objorke:
Completed successfully now!

tibel:
We have Xamarin build :sunglasses:

Issue #615, opened 2015-10-05
Support Windows 10 UWP apps (uap10.0)


objorke:
Add a project that targets Windows 10 / Universal Windows apps

objorke:
Or is this not needed? Can the Windows 8.1 package be used?

csabar:
I have successfully compiled OxyPlot.Windows to support UAP, but it was not so straightforward for me. The binaries are not compatible, but the source code is compatible. I think the only change I made was adding the UNIVERSAL definition to a default UAP project. But you definitively need at least separate nuget package.

objorke:
There is a nuget restore error on appveyor, anyone knows why? see #616

objorke:
the UAP project is built and published to NuGet (included in the OxyPlot.Windows package), but it seems like the OxyPlot.Core package is not working with UAP

tibel:
What do you mean with not working?

objorke:
When I add the OxyPlot.Windows package to a UAP10.0 app, the OxyPlot.Core package is not included in the references. Shouldn't it be there? Or is this correct? It seems to build fine, but does not run. The example in the `OxyPlot.Windows.sln` is working... Will try to replace the project references with NuGet references...

objorke:
I opened OxyPlot.Windows.sln and changed the Windows10/SimpleDemo project:
I replaced the OxyPlot and OxyPlot.Windows10 references with the latest OxyPlot.Windows NuGet package - it builds ok but does not run...

tibel:
What I don't understand is, why win10 OxyPlot looks different in nuget than win81?
- native is not needed for a class library, only for apps; stay with AnyCPU
- nuspec for win10 looks strange with native parts

It should look like this:
``` xml
<file src="..\..\Output\Windows10\OxyPlot.Windows10.???" target="lib\uap10.0" />
<file src="..\..\Output\Windows10\Themes\*.*" target="lib\uap10.0\OxyPlot.Windows10\Themes" />
```
As simple as for portable-windows8+wpa81, no need for other stuff. Maybe that is your problem?

objorke:
I could not get AnyCPU to build on AppVeyor. Can try again!
Thanks, will try changing the nuspec!

GeertvanHorrik:
I was just trying this out, so this is very recent. Whenever I use this, I get the exception that "Oxyplot.Windows" cannot be loaded. Maybe because it's called OxyPlot.Windows10 for UA{?

[...]

PR #616, opened 2015-10-05
Add Windows Universal (uap10) project

objorke:
To solve issue #615


objorke:
nuget restore fails on appveyor. No problems locally. I don't understand... :-)

csabar:
I was unable to build your files. The problem is that Visual Studio Express for Windows 10 is unable to load project "OxyPlot" as reference. I think you need to create a new solution file, which can be loaded with Visual Studio for Windows 10.

objorke:
The build was OK locally with Visual Studo 2015. The solution should be OK, but I don´t understand why nuget restore fails on appveyor for the Windows10 project.

csabar:
Even if you have correct build at your computer, I was unable to build your commit and maybe that is why AppVeyor build has also failed. The OxyPlot.csproj needs to be updated to support UWP platform. Maybe you can create new project from scratch and merge with current project or make a copy with different name and use this guide: https://msdn.microsoft.com/en-us/library/mt148501.aspx
I think the problem might be with AnyCpu because it is mentioned in AppVeyor message and according to the guide:
> 8 . Find <PropertyGroup> elements that are configured for the AnyCPU platform as part of the Condition attribute. Remove these elements and all of their children. AnyCPU is not supported for Windows 10 apps in Visual Studio 2015. 

But I'm just guessing the reason. I don't know how the AppVeyor works.

objorke:
Thanks for the link! I will try to create a UWP version of the core project as you suggest!

objorke:
I had some progress yesterday, now there is a netcore 4.5.1 error.

objorke:
http://help.appveyor.com/discussions/problems/2993-uwp-test-assembly-build-failing
http://help.appveyor.com/discussions/problems/2702-vs2015-uwp-project
http://help.appveyor.com/discussions/questions/1320-windows-10-support

PR #621, opened 2015-10-13
Respect MinimumRange values when Minimum and Maximum values are also set

### LINKED ISSUE #550: MinimumRange gives odd result
I have a LineSeries where I have set the axis to have a minimum of 0. I have also set the minimumrange to be 10. With my data, that gives a display from -5 to 5. I think it would make more sense to use the minimum when deciding what the lower end of the range should be.

I tested this with the 2014.1.546 and 2015.1.982.

classicboss302:
Resolves #550.

### DIFF
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -722,10 +722,21 @@ namespace OxyPlot.Axes
             var range = this.ActualMaximum - this.ActualMinimum;
             if (range < this.MinimumRange)
             {
-                var average = (this.ActualMaximum + this.ActualMinimum) * 0.5;
-                var delta = this.MinimumRange / 2;
-                this.ActualMinimum = average - delta;
-                this.ActualMaximum = average + delta;
+                if (!double.IsNaN(this.Minimum) && !double.IsNaN(this.Maximum))
+                {
+                    var average = (this.ActualMaximum + this.ActualMinimum) * 0.5;
+                    var delta = this.MinimumRange / 2;
+                    this.ActualMinimum = average - delta;
+                    this.ActualMaximum = average + delta;
+                }
+                else if (!double.IsNaN(this.Minimum) && double.IsNaN(this.Maximum))
+                {
+                    this.ActualMaximum = this.Minimum + this.MinimumRange;
+                }
+                else if (!double.IsNaN(this.Maximum) && double.IsNaN(this.Minimum))
+                {
+                    this.ActualMinimum = this.Maximum - this.MinimumRange;
+                }
             }
 
             if (this.AbsoluteMaximum <= this.AbsoluteMinimum)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -104,6 +104,7 @@ All notable changes to this project will be documented in this file.
 - Fix null reference exception when ActualPoints was null rendering a StairStepSeries (#582)
 - Background color in the Xamarin.Forms views (#546)
 - IsVisible change in Xamarin.Forms.Platform.iOS (#546)
+- Fix issue with MinimumRange not taking Minimum and Maximum values into account (#550)
 
 ## [2014.1.546] - 2014-10-22
 ### Added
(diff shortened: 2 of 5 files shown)


Source/OxyPlot/Axes/Axis.cs
```diff
@@ -722,10 +722,21 @@ public virtual void CoerceActualMaxMin()
             var range = this.ActualMaximum - this.ActualMinimum;
             if (range < this.MinimumRange)
             {
-                var average = (this.ActualMaximum + this.ActualMinimum) * 0.5;
-                var delta = this.MinimumRange / 2;
-                this.ActualMinimum = average - delta;
-                this.ActualMaximum = average + delta;
+                if (this.Minimum != double.NaN)
```
objorke: isn't it better to use `!double.IsNan(this.Minimum)` for such NaN comparisons?

objorke:
Thanks for contributing! I have a few comments:
1. Is this issue covered by tests/examples? If not, please add some models in the example library that demonstrates the problem. I will run these before merging the code.
2. Please update the change log an contributors files.
Please have a look at the [contributing guidelines](https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTING.md) :-)

classicboss302:
I think I have everything in place now. Thanks for the feedback.

objorke:
I am starting a rebuild of this PR

Source/Examples/WPF/WpfExamples/Examples/MinRangeDemo/MainWindow.xaml.cs
```diff
@@ -0,0 +1,70 @@
...
+    [Example("Shows OxyPlot with a Minimum Range and Minimum values.")]
+    public partial class MainWindow : Window
+    {
+        Random r = new Random(13);
+
+        public MainWindow()
+        {
+            this.InitializeComponent();
+
+            Model = GenerateRandomPlotModel(string.Format("MinRange plot"));
```
objorke: I would prefer this example to be added to the ExampleLibrary instead!

objorke:
Please remove the MinRangeDemo and move the example PlotModel to the ExampleLibrary

objorke:
The build error seems to be in the GitVersion step:
``` 
ERROR [10/17/15 6:53:09:73] An unexpected error occurred:
System.Net.WebException: The remote server returned an error: (500) Internal Server Error.
   at System.Net.HttpWebRequest.GetResponse()
   at GitVersion.AppVeyor.GenerateSetVersionMessage(String versionToUseForBuildNumber)
   at GitVersion.BuildServerBase.WriteIntegration(Action`1 writer, VersionVariables variables)
   at GitVersion.SpecifiedArgumentRunner.Run(Arguments arguments, IFileSystem fileSystem)
   at GitVersion.Program.VerifyArgumentsAndRun()
INFO [10/17/15 6:53:09:73] 
INFO [10/17/15 6:53:09:73] Here is the current git graph (please include in issue): 
INFO [10/17/15 6:53:09:73] Showing max of 100 com
```

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -722,10 +722,21 @@ public virtual void CoerceActualMaxMin()
             var range = this.ActualMaximum - this.ActualMinimum;
             if (range < this.MinimumRange)
             {
-                var average = (this.ActualMaximum + this.ActualMinimum) * 0.5;
-                var delta = this.MinimumRange / 2;
-                this.ActualMinimum = average - delta;
-                this.ActualMaximum = average + delta;
+                if (!double.IsNaN(this.Minimum))
+                {
+                    this.ActualMaximum = this.Minimum + this.MinimumRange;
```
objorke: this is a breaking change, isn´t it?

objorke: to keep the current behavior you could use the current code `if (!double.IsNaN(this.Minimum) && !double.IsNaN(this.Maximum))`, and use the new code in the two other cases

classicboss302:
I have updated the pull request.

PR #625, opened 2015-10-15
Only return a default controller if a custom controller is not specified

### LINKED ISSUE #436: Setting a custom PlotController on PlotView is ignored
Tested on development branch. Up-to-date 1. April .2015

I have a bit of code like this: 

```xaml
<DataTemplate>
                            <oxy:PlotView x:Uid="plot"
                                          x:Name="plot"
                                          Model="{Binding PlotModel}"
                                          Controller="{Binding Controller}">
                            </oxy:PlotView>
 </DataTemplate>
```

The Plotview fully ignores this. My ViewModel code works fine for the Model property though. 
Neither does this: 

```xaml
<DataTemplate>
                            <oxy:PlotView x:Uid="plot"
                                          x:Name="plot"
                                          Model="{Binding PlotModel}">
                           </oxy:PlotView>
<oxy:PlotView.Controller>
    <CustomPlotController/>
</oxy:PlotView.Controller>
 </DataTemplate>
```

I found that removing the default intialiser in `Oxyplot.Wpf.PlotView` non-argument constructor makes it work. But why i couldn't figure out.

classicboss302:
Fixes #436

### DIFF
--- a/Source/OxyPlot.Wpf.Tests/PlotViewTests.cs
+++ b/Source/OxyPlot.Wpf.Tests/PlotViewTests.cs
@@ -107,5 +107,23 @@ namespace OxyPlot.Wpf.Tests
                 OxyAssert.PropertiesAreEqual(model, view);
             }
         }
+
+        /// <summary>
+        /// Provides unit tests for the <see cref="PlotView.Controller" /> method.
+        /// </summary>
+        public class Controller
+        {
+            /// <summary>
+            /// Assert that the controller exists <see cref="PlotView" />.
+            /// </summary>
+            [Test]
+            public void VerifyController()
+            {
+                var model = new PlotModel();
+                var view = new PlotView();
+
+                Assert.NotNull(view.Controller);
+            }
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


objorke:
Can you add a test or example that reproduces the error with a Wpf DataTemplate?

Source/OxyPlot.Wpf.Tests/PlotViewTests.cs
```diff
@@ -107,5 +107,23 @@ public void PlotModelVsPlot()
...
+        /// </summary>
+        public class Controller
+        {
+            /// <summary>
+            /// Assert that the controller exists <see cref="PlotView" />.
+            /// </summary>
+            [Test]
+            public void VerifyController()
+            {
+                var model = new PlotModel();
```
objorke: not used

Source/OxyPlot.Wpf/PlotView.cs
```diff
@@ -83,6 +82,12 @@ public IPlotController Controller
         {
             get
             {
+                // Set a default value if no custom controller has been specified
+                if ((IPlotController)this.GetValue(ControllerProperty) == null)
+                {
+                    this.Controller = new PlotController();
```
objorke: I think this is not the right place to fix this issue. I will have a look.

objorke:
I tried a different solution in #630

classicboss302:
Your solution resolves the issue for me. Thank you.

objorke:
Ok, I merged #630

PR #630, opened 2015-10-17
Controller and datatemplate issue

objorke:
This should solve #436


ivanbenko:
@objorke I've just noticed a breaking change caused by this PR. Now Controller is shared between all PlotView instances by default (If you don't specify custom Controller).

objorke:
Thanks, I created #682!

PR #582, opened 2015-09-16
Fix NullReferenceException

karlmtc:
If ActualPoints is null, we can't access the count. We need to check for null first.

PR #637, opened 2015-10-26
Fixed NullReferenceException on ScatterSeries{T}.

### LINKED ISSUE #636: Null reference exception in ScatterSeries{T}
Pretty much like #605, the problem seems to be that we are accessing a property of ActualPoints when ActualPoints is null. Here's my stack, in case it helps.

image

karlmtc:
Accessing actualPoints.Count throws an exception when actualPoints is null. This fixes #636.

### DIFF
--- a/Source/OxyPlot/Series/ScatterSeries{T}.cs
+++ b/Source/OxyPlot/Series/ScatterSeries{T}.cs
@@ -316,11 +316,13 @@ namespace OxyPlot.Series
         public override void Render(IRenderContext rc)
         {
             var actualPoints = this.ActualPointsList;
-            int n = actualPoints.Count;
-            if (n == 0)
+            
+            if(actualPoints == null || actualPoints.Count == 0)
             {
                 return;
             }
+            
+            int n = actualPoints.Count;
 
             var clippingRect = this.GetClippingRect();
 


objorke:
As for StairstepSeries (#582, #606), I would like to see
1. an issue with steps to reproduce the error
2. update change log
3. reference to the issue from the PR

https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTING.md

objorke:
The fix looks perfectly fine - I am just trying to enforce the "contributing guidelines" here ;-)

interdrift:
Shouldn't ActualPoints never be null, except when the DataSource is null?

karlmtc:
In that case it will throw an exception, that's what I was trying to fix.

interdrift:
Oh okay, I guess I'm used to throwing an exception when the DataSource is null.

objorke:
I think `ActualPoints` should never be null. If `ItemsSource` is null, `ActualPoints` should use the `points` collection that is read-only.

Could it be a race condition going on?

objorke:
I included your commit in #703!
Updated the change log.
No example is included since I cannot reproduce the issue.
But I think it is ok since this makes it consistent with the LineSeries implementation.

objorke:
Will merge #703

PR #661, opened 2015-11-26
Update AreaSeries.cs

less0:
Implemented the calculation of baseline values if Points2 is empty.

### DIFF
--- a/Source/OxyPlot/Series/AreaSeries.cs
+++ b/Source/OxyPlot/Series/AreaSeries.cs
@@ -121,7 +121,7 @@ namespace OxyPlot.Series
             {
                 return this.ItemsSource != null 
                     ? this.itemsSourcePoints2 
-                    : (this.points2.Any() ? this.points2 : this.Points.Select(p => new DataPoint(p.X, ConstantY2)));
+                    : (this.points2.Any() ? this.points2 : this.Points.Select(p => new DataPoint(p.X, ConstantY2)).ToList());
             }
         }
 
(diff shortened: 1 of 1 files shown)


objorke:
Please create an issue first. :-)
See also https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTING.md

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -118,7 +119,9 @@ protected List<DataPoint> ActualPoints2
         {
             get
             {
-                return this.ItemsSource != null ? this.itemsSourcePoints2 : this.points2;
+                return this.ItemsSource != null 
+                    ? this.itemsSourcePoints2 
+                    : (this.points2.Any() ? this.points2 : this.Points.Select(p => new DataPoint(p.X, ConstantY2)).ToList());
```
tibel: Only use ToList() for the `Select()` clause as `this.points2` is a list already.
Also I think the points should be cached in `itemsSourcePoints2` similar to the implementation of DataPointSeries.

objorke:
I think two points should be sufficient for the second point set using "ConstantY2" values.

objorke:
Reference: #662

objorke:
I squashed your commits in https://github.com/objorke/oxyplot/tree/bugfix/areaseries-pauls-solution
and created a slightly alternative solution in https://github.com/objorke/oxyplot/tree/bugfix/areaseries-constant-baseline
See PR in #701 - includes examples, updated change log and everything in 1 commit

objorke:
I merged #701

Issue #668, opened 2015-12-03
Limiting the size of the legend


redcurry:
I don't see a way to limit the size of the legend. The more series I add to the plot, the larger the legend becomes and the smaller the plot (specifically for an outside legend at the bottom). At the very least, I think there should be a way to specify a maximum height (or width) of the legend. This would also make `PlotModel.LegendColumnSpacing` more useful because, as far as I can tell, the only way for there to be multiple columns in the legend is if it has taken up the entire plot height.

redcurry:
I see that there's a `PlotModel.LegendMaxWidth` but not one for height.

objorke:
It sounds good to add a `LegendMaxHeight` similar to the existing one.
Remember to include example in a PR.

redcurry:
I added the feature to the OxyPlot project and tested it in my own solution. Can I do a pull request as is, or do I need to test it with the other projects (Wpf, MonoTouch, Silverlight, etc.)?

objorke:
Please include an example in the example library - then we can test on all platforms.
See https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTING.md

PR #671, opened 2015-12-07
Add LegendMaxHeight property to PlotModel (#668)

redcurry:


Source/OxyPlot/PlotModel/PlotModel.cs
```diff
@@ -478,6 +479,12 @@ public IPlotView PlotView
         public double LegendMaxWidth { get; set; }
 
         /// <summary>
+        /// Gets or sets the max height of the legend.
+        /// </summary>
+        /// <value>The max height of the legend.</value>
+        public double LegendMaxHeight { get; set; }
```
objorke: OxyPlot.Wpf.Plot should also be updated with a dependency property

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
```diff
@@ -500,7 +500,8 @@ private void UpdatePlotArea(IRenderContext rc)
 
             // Find the available size for the legend box
             var availableLegendWidth = plotArea.Width;
-            var availableLegendHeight = plotArea.Height;
+            var availableLegendHeight = double.IsNaN(this.LegendMaxHeight) ?
```
objorke: looks good,

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
```diff
@@ -500,7 +500,8 @@ private void UpdatePlotArea(IRenderContext rc)
 
             // Find the available size for the legend box
             var availableLegendWidth = plotArea.Width;
```
objorke: why is availableLegendWidth not consistent with availableLegendHeight? can this line be made consistent with the other?

objorke:
Excellent PR! Looks 99% ok...

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
redcurry: Originally, I hadn't changed the line above, but it caused items that didn't fit in the legend to be shown outside and to the right of the legend. This happened because, if the height was modified to the max, the width of the legend wasn't changed in order to accommodate all items. In other words, the legend was being rendered with the width measured for the original height, not the new height. The same thing happens with LegendMaxWidth. Items that don't fit the new legend are shown outside and to the bottom of the legend. Is this the intended behavior? What I want is to be able to set a max height and let the plot model figure out the width needed to fit all of the items. My changes do that for height but not for width because I didn't want to change any expected behavior.

objorke: I think legend items that don't fit within the max width/height should be skipped. Not sure if that works, will test!
I see there is a lot of cases to test.. Will play with the new code! 
It sounds good not to change the current behavior.

Source/OxyPlot/PlotModel/PlotModel.cs
objorke: I can handle this

Issue #692, opened 2015-12-19
All structs should follow "Struct Design" principles


tibel:
https://msdn.microsoft.com/en-us/library/ms229031.aspx
- no default constructor
- no property setter or settable field
- implement `IEquatable<T>`

This should be checked and fixed for all `struct` types in OxyPlot.
Especially `DataPoint` is violating the immutability constraint!

objorke:
I will change the `DataPoint`.

objorke:
Struct types:
- [ ] DataPoint
- [ ] PlotLength
- [ ] SelectionItem
- [ ] OxyColor
- [ ] OxyRect
- [ ] OxySize
- [ ] OxyThickness
- [ ] ScreenPoint
- [ ] ScreenVector
- [ ] BoxPlotItem

objorke:
What about item types like ScatterPoint, BarItem etc. (reference types) - should they also be immutable?

objorke:
I will change BoxPlotItem to a reference type. It is too big for a value type.

objorke:
I think this one is complete now.

PR #694, opened 2015-12-20
Make OxyPlot structs immutable

objorke:
For #692

### DIFF
--- a/Source/OxyPlot/Rendering/OxySize.cs
+++ b/Source/OxyPlot/Rendering/OxySize.cs
@@ -24,28 +24,49 @@ namespace OxyPlot
         public static readonly OxySize Empty = new OxySize(0, 0);
 
         /// <summary>
+        /// The height
+        /// </summary>
+        private readonly double height;
+
+        /// <summary>
+        /// The width
+        /// </summary>
+        private readonly double width;
+
+        /// <summary>
         /// Initializes a new instance of the <see cref="OxySize" /> struct.
         /// </summary>
         /// <param name="width">The width.</param>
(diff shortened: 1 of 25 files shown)


objorke:
@tibel can you review this PR?

Source/OxyPlot/Rendering/OxySize.cs
```diff
@@ -31,21 +41,33 @@ public struct OxySize : IFormattable
         public OxySize(double width, double height)
             : this()
```
tibel: I think calling a default constructor on a struct is not needed.
All memory is zeroed by default.

tibel:
Looks good. 

Maybe adding some helper methods/operators to make code simpler:
e.g. OxyRect add/subtract operator for size
or similar so you don't have to call the constructor with all parameters again

objorke:
I have fixed the OxySize ctor.

Source/OxyPlot/Rendering/OxySize.cs
objorke: Fixed!

objorke:
Yes, I noticed changes to `OxyRect` instances became more complicated.
I prefer `Inflate` and `Offset` methods over operators, as in
https://msdn.microsoft.com/en-us/library/system.windows.rect(v=vs.110).aspx

objorke:
I added some methods to OxyRect, but I am not very happy with the result...
The Inflate/Deflate methods are ok, but the others don't help much...

objorke:
I removed the confusing "change size" methods...

tibel:
Great. Thanks for changing this.

Issue #662, opened 2015-11-26
AreaSeries ConstantY2 is not implemented


less0:
The default behavior of the AreaSeries is to display the area between the graph and the line between the first and the last DataPoint. If no DataPoints are provided in Points2 the behavior should be to use ConstantY2, which is not implemented, yet. 

I've already implemented a first basic version of it and I am only waiting for my you-take-it assignment.

I am using the latest unstable Version, WindowsForms, .NET 4.5.2

objorke:
The comment on ConstantY2 should be updated
https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Series/AreaSeries.cs#L41

I think it should be "ConstantY2 is used if `ItemsSource` is set and `DataField2` is `null`, or if `ItemsSource` is `null` and `Points2` is empty."

objorke:
An alternative could be to use `ConstantY2` if it is not NaN. But that would be a breaking change.

objorke:
It seems `ConstantY2` is currently not used...

objorke:
Please include examples for all cases: 
- with ItemsSource
- null ItemsSource 
- null DataField2
- empty Points2
etc.

PR #701, opened 2015-12-31
Implement constant baseline for AreaSeries

objorke:
Solves #662

### DIFF
--- a/Source/OxyPlot/Series/AreaSeries.cs
+++ b/Source/OxyPlot/Series/AreaSeries.cs
@@ -9,6 +9,7 @@
 
 namespace OxyPlot.Series
 {
+    using System.Linq;
     using System.Collections.Generic;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -324,5 +349,22 @@ protected internal override void UpdateMaxMin()
...
             this.InternalUpdateMaxMin(this.ActualPoints2);
         }
+
+        /// <summary>
+        /// Gets the points when <see cref="P:ConstantY2" /> is used.
+        /// </summary>
+        /// <returns>A sequence of <see cref="T:DataPoint"/>.</returns>
+        private IEnumerable<DataPoint> GetConstantPoints2()
+        {
+            var actualPoints = this.ActualPoints;
```
objorke: base the constant second data set on the first and last point of the first data set

Issue #401, opened 2015-03-04
Zoom constraints


TheAlmightyBob:
It would be nice to be able to specify a min/max zoom level (scale) for an axis, separate from the absolute min/max axis values.

(so that zooming could be restricted without restricting panning)

Gaisuru:
It would be nice, but you can hack it together as is to accomplish this.

objorke:
Good idea. Could the properties be called `MinimumRange` and `MaximumRange`, where 'range' is the difference between actual maximum and minimum? Default values `0` and `+INF`.

tibel:
@tclaff @objorke Can this issue be closed now?

tclaff:
Yes, it can be closed.

PR #710, opened 2016-01-01
MinimumRange and MaximumRange to limit the zoom

tclaff:
Reworked the MinimumRange behaviour and added MaximumRange to limit zooming out. (#401)

### DIFF
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -68,6 +68,7 @@ namespace OxyPlot.Axes
             this.MinimumPadding = 0.01;
             this.MaximumPadding = 0.01;
             this.MinimumRange = 0;
+            this.MaximumRange = double.PositiveInfinity;
 
             this.TickStyle = TickStyle.Outside;
             this.TicklineColor = OxyColors.Black;
@@ -362,6 +363,11 @@ namespace OxyPlot.Axes
         public double MaximumPadding { get; set; }
 
         /// <summary>
+        /// Gets or sets the maximum range of the axis. Setting this property ensures that <c>ActualMaximum-ActualMinimum < <c>MaximumRange</c>. The default value is <c>double.PositiveInfinity</c>.
+        /// </summary>
+        public double MaximumRange { get; set; }
+
+        /// <summary>
         /// Gets or sets the minimum value of the axis. The default value is <c>double.NaN</c>.
         /// </summary>
         public double Minimum { get; set; }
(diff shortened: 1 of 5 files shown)


Source/OxyPlot/Axes/Axis.cs
```diff
@@ -362,6 +363,11 @@ public bool IsReversed
         public double MaximumPadding { get; set; }
 
         /// <summary>
+        /// Gets or sets the maximum range of the axis. Setting this property ensures that <c>ActualMaximum-ActualMinimum < <c>MaximumRange</c>. The default value is <c>double.PositiveInfinity</c>.
```
objorke: &lt; ?

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -373,7 +379,7 @@ public bool IsReversed
         public double MinimumPadding { get; set; }
 
         /// <summary>
-        /// Gets or sets the minimum range of the axis. Setting this property ensures that <c>ActualMaximum-ActualMinimum > MinimumRange</c>. The default value is <c>0</c>.
+        /// Gets or sets the minimum range of the axis. Setting this property ensures that <c>ActualMaximum-ActualMinimum > <c>MinimumRange</c>. The default value is <c>0</c>.
```
objorke: Is the ">" valid?

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -718,25 +724,24 @@ public virtual void CoerceActualMaxMin()
...
             }
 
-            // Coerce the minimum range
             var range = this.ActualMaximum - this.ActualMinimum;
+
+            // Coerce the minimum range
             if (range < this.MinimumRange)
+            { 
+                var average = (this.ActualMaximum + this.ActualMinimum) * 0.5;
+                var delta = this.MinimumRange * 0.5;
```
objorke: this is changing the current behavior if Minimum or Maximum are set

objorke: I think this was fixed recently, check the history

objorke:
Please fix build errors and verify that the "MinimumRange with Minimum" example is still working.

Source/OxyPlot/Axes/Axis.cs
tclaff: I think it's correct, but the second <c> is not.

Source/OxyPlot/Axes/Axis.cs
tclaff: I think it is correct.

Source/OxyPlot/Axes/Axis.cs
tclaff: This explains what the code was intended for. However, i think the fix for #550 broke the intended feature of the MinimumRange property. In the example "MinimumRange with Minimum" the MinimumRange is set to 5 but the user can clearly zoom beyond 5 units on the axis. I am not sure what to do about this.

tclaff:
I have reverted the MinimumRange behavior to the old one. Since i think there is an issue with it, i will open a separate ticket for that.

objorke:
You are right, something is wrong with the MinimumRange implementation. #550

Issue #672, opened 2015-12-07
Axis titles (right-side only) change position when using ReportItem.AddPlot vs. PDFExporter


kfpulver:
![pdfexporter](https://cloud.githubusercontent.com/assets/15821076/11638907/05e89752-9cef-11e5-9719-0b69fcdfed97.PNG)
Above created with PDFExporter.Export - appearance is as expected.
![addplot](https://cloud.githubusercontent.com/assets/15821076/11638912/0b52aaf2-9cef-11e5-81d7-3f7172c93bd3.PNG)
Above created with ReportItem.AddPlot - titles are shifted unexpectedly.

It seems as if the text block "handles" are at the top of the block rather than the bottom of the text.  The left-side primary y-axis is not affected.  Is there a fix or known workaround?

Thanks.

- Edited above to read 'y-axis'
- X-axis is also not affected

objorke:
What is the target output of the report? Is the plot with issue exported as a png in the report?

kfpulver:
```
var r = new Report();
var main = new ReportSection();
r.Add(main);
main.AddPlot(dataset.PrintablePlot, dataset.SaveName, 800, 600);
```

PrintablePlot is a PlotModel.  The report is being written with OxyPlot.Pdf.PdfReportWriter.  Output is a PDF file.

```
using (var w = new OxyPlot.Pdf.PdfReportWriter(saveFileDialog1.FileName))
{
     w.WriteReport(r, reportStyle);
}
```

Thanks in advance for your help.

objorke:
Seems like there is an issue with the pdf plot exporter based on pdfsharp!

kfpulver:
Any chances you know a quick fix?  Is there a better method to use for PDF report style exporting? I’m noticing other major issues with image annotations as well (Inverting with other displaying issues - I’ll detail in another issue thread at a later time).  The axis title issue is my primary concern.  Would this be better directed at pdfsharp?

objorke:
I guess there is an issue with vertical alignment of rotated text in the OxyPlot.Pdf render context (DrawText or MeasureText method)
https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot.Pdf/PdfRenderContext.cs#L237

kfpulver:
Thanks for the pointer.  I ended up making the following edit to PDFRenderContext.cs.
```
if (Math.Abs(rotate) > double.Epsilon)
{
[...]

PR #723, opened 2016-01-06
Update PdfRenderContext.cs

kfpulver:
Per Issue #672 

Rotation was not compensating for the initial text alignment translations.


objorke:
Good! Please update the change log and contributors files.

objorke:
thanks!

kfpulver:
Thanks again for the guidance.  Let me know if there's anything else that should be done.

Issue #697, opened 2015-12-24
Add support for Xamarin.Forms UWP


choenden:
With xamarin.forms, I created a UWP application. When the nugget package for OxyPlot is added, there is no Platform.UWP for which I could call this initialization method:
OxyPlot.Xamarin.Forms.Platform.UWP.PlotViewRenderer.Init();

I'm not yet sure whether I missed something or whether this really is the case. Looking at the source code, there only seems to be a win8/wp8 project. I made a small local fork where I modified the win8 project to use the UWP platform instead.

choenden:
Sorry, turns out this is a duplicate of #615

objorke:
No, I think this is a valid issue. In my opinion, #615 does not include support for Xamarin.Forms

objorke:
https://developer.xamarin.com/guides/xamarin-forms/windows/getting-started/universal/
Only available in preview yet?

tibel:
Also related to #461

choenden:
Cool, @tibel @objorke can I take this then? I've been meaning to contribute to something opensource for a while. UWP is available in the released xamarin forms 2.0 but it is in the preview phase.

objorke:
You are very welcome to take this task! I'll assign it to you.
I think #615 is a prerequisite, I have run into some build problems on this, see https://github.com/oxyplot/oxyplot/pull/616

choenden:
Thanks @objorke . I have created the UWP project and it seems to work fine with my own example project. I expanded the SimpleDemo to also have a UWP target, but there I am getting this exception. Is this a known issue or do you have any advice?
		Message	"Cannot deserialize XBF metadata type list as 'ThicknessConverter' was not found in namespace 'OxyPlot.Windows'. [Line: 0 Position: 0]"	string

choenden:
Ah. Forgot to reference that assembly in the sample project -_-

andrii-borysov-me:
Hi guys. It seems we're still missing the support on nuget. Any update or expectation, when we'll be happy to have it? Thanks.

dimitrijevic:
[...]

PR #746, opened 2016-01-21
Update OxyPlot.Xamarin.Forms nuspec XF 2.0.0.6490

dimitrijevic:
Additionally, the portable library target string could be extended to include +UAP10 support, but I'm guessing the build should need to be extended then also to copy output from UWP release build directory to a lib/UAP10 folder to enable adding after line 52 these lines: 

    <file src="..\..\Output\Xamarin.Forms.Platform.UWP\Xamarin.Forms.Platform.UWP.???" target="lib/UAP10" />
    <file src="..\..\Output\Xamarin.Forms.Platform.UWP\Xamarin.Forms.???" target="lib/UAP10" />

I don't have AppVeyor experience for that, but I see the UWP .csproj builds, although it's not included in the .sln (or such file hasn't been pushed yet)


objorke:
Thanks! Yes, we should include support for UWP - see #697

PR #752, opened 2016-01-31
Fix HeatMapSeries cannot plot on Universal Windows

tibel:
Current implementation of `ConvertToRandomAccessStream()` deadlocked with the await and the `.Result` in the calling method.


objorke:
great!!

Issue #758, opened 2016-02-03
CalculateActualInterval in Axis class stuck in infinite loop for certain values (OxyPlot.Wpf)


vhoehn:
Hi there,
 I found the CalculateActualInterval method in the Axis class (OxyPlot.Wpf) to be stuck in an infinite loop if range or maxIntervalSize is 0. (Current OxyPlot Version)

For maxIntervalSize = 0, maxIntervalCount will be infinite which prevents the while loop from terminating. For range = 0, the interval will be 0, therefore, the first intervalCandidate will also be 0. The division range / intervalCandidate is double.NaN and also preventing the while loop to terminate.

```c#
protected double CalculateActualInterval(double availableSize, double maxIntervalSize, double range)
        {
            if (availableSize <= 0)
            {
                return maxIntervalSize;
            }

            Func<double, double> exponent = x => Math.Ceiling(Math.Log(x, 10));
            Func<double, double> mantissa = x => x / Math.Pow(10, exponent(x) - 1);

            // reduce intervals for horizontal axis.
            // double maxIntervals = Orientation == AxisOrientation.x ? MaximumAxisIntervalsPer200Pixels * 0.8 : MaximumAxisIntervalsPer200Pixels;
            // real maximum interval count
            double maxIntervalCount = availableSize / maxIntervalSize;

            range = Math.Abs(range);
            double interval = Math.Pow(10, exponent(range));
            double intervalCandidate = interval;

            // Function to remove 'double precision noise'
            // TODO: can this be improved
            Func<double, double> removeNoise = x => double.Parse(x.ToString("e14"));

            // decrease interval until interval count becomes less than maxIntervalCount
            while (true)
            {
                var m = (int)mantissa(intervalCandidate);
                if (m == 5)
                {
                    // reduce 5 to 2
                    intervalCandidate = removeNoise(intervalCandidate / 2.5);
                }
                else if (m == 2 || m == 1 || m == 10)
                {
                    // reduce 2 to 1, 10 to 5, 1 to 0.5
                    intervalCandidate = removeNoise(intervalCandidate / 2.0);
                }
                else
                {
                    intervalCandidate = removeNoise(intervalCandidate / 2.0);
                }

                if (range / intervalCandidate > maxIntervalCount)
                {
                    break;
                }

                if (double.IsNaN(intervalCandidate) || double.IsInfinity(intervalCandidate))
                {
                    break;
                }

                interval = intervalCandidate;
            }

            return interval;
        }
```

[...]

PR #762, opened 2016-02-05
Fix CalculateActualInterval() stuck in infinite loop (#758).

vhoehn:
Suggested fix for issue #758.

### DIFF
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -1593,6 +1593,16 @@ namespace OxyPlot.Axes
                 return maxIntervalSize;
             }
 
+            if (Math.Abs(maxIntervalSize) < Double.Epsilon)
+            {
+                throw new ArgumentException("Maximum interval size cannot be zero.", "maxIntervalSize");
+            }
+
+            if (Math.Abs(range) < Double.Epsilon)
+            {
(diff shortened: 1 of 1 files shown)


objorke:
I added the example and updated the change log in #766

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -1593,6 +1593,16 @@ protected double CalculateActualInterval(double availableSize, double maxInterva
                 return maxIntervalSize;
             }
 
+            if (Math.Abs(maxIntervalSize) < Double.Epsilon)
```
objorke: should use double

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -1593,6 +1593,16 @@ protected double CalculateActualInterval(double availableSize, double maxInterva
...
+            {
+                throw new ArgumentException("Maximum interval size cannot be zero.", "maxIntervalSize");
+            }
+
+            if (Math.Abs(range) < Double.Epsilon)
```
objorke: double

objorke:
I merged your change in #766

alexdegroot:
This one is solved in 32-bit, but still reproducible in 64-bit.

Issue #759, opened 2016-02-04
PngExporterTests.ExportWithResolution not working on windows 8 or newer


benjaminrupp:
The tests PngExporterTests.ExportWithResolution(2) and PngExporterTests.ExportWithResolution(4) are not working on new versions of windows cause the used font "Segoe UI" has changed with windows 8 (see https://en.wikipedia.org/wiki/Segoe#Segoe_UI).
I suggest changing the font of the chart of these tests. I think "Times New Roman" would be a good choise. At least this is working both on windows 7 and windows server 12.

objorke:
I get different results for "Times New Roman" on Windows 10 (local) and Windows Server 2012 R2 (appveyor).
Should we include a font in the test assembly?
Maybe something else than Times New Roman? See license terms: https://www.microsoft.com/typography/Fonts/family.aspx?FID=9

benjaminrupp:
You are right. A font in test assembly would be the best. This would solve the problem for all windows versions.
Unfortunately I am not able to access my developer pc in the momemt. 
Do you have any preferences regarding the font in test assembly?
I will make a new pull request as soon as I am back in office.

objorke:
no preference. but it must have a compatible license

objorke:
The tests are failing in #797.
Maybe the issue is font rendering and not the font itself?

PR #772, opened 2016-02-12
#759 fixed test to run on all versions of windows

benjaminrupp:
added test font


objorke:
the broken build seems to be a Xamarin license issue. I will investigate it

objorke:
now it seems like gitversion is the problem.

objorke:
should the font be installed before running the test?

benjaminrupp:
I did not install the font on my system. I just copied it to the output folder by the setting "copy if newer".

benjaminrupp:
I have no idea what's the problem with gitversion. Do I have to change something on my commit?

objorke:
I tried pack uris for the fonts. See #775

objorke:
The builds never succeeded, right? Should we close this PR?

benjaminrupp:
I have no idea why this is not running on the build server.
It is running correctly on my win 7 developer pc and also on my windows 2012 team city server.
If i am using the old test its not running on win 2012 cause the microsoft changed the font with win 8 and all following windows versions.

tibel:
We fixed some issues related to the build server in the meantime.
Try to rebase your changes so the build server takes another try.
When it still fails, I will close this pull request.

benjaminrupp:
Hi tibel,

the AppVeyor server is behaving strange. 
Yesterday it was running. I had just one bug in my test. Now it is not running anymore. But my test is okay now. Its running local and on my own build server.

I just don't now what I should do to fix the AppVeyor.

Regards

Benjamin

tibel:
According to the build log it seems that _gitversion_ has some problem.

You can try to use newer _gitversion_ by updating `appveyor.yml` and set version to `3.5.2`.
I haven't tested this, but worth to try.

benjaminrupp:
I tried. But it is not working :-(

objorke:
The unit tests are failing in the last build.
Are the tests too strict? Should we not ask for pixel perfect?

benjaminrupp:
If the Tests are failing, then the test Font is not used. Arial is in this Case used as default. I am not sure why the Test does not find the fonts. Unfortunately, i broke my arm and i will not be able to access my computer for the next  Weeks. I will take care later.

objorke:
Can't merge this since it breaks the appveyor build.
Maybe you could ignore the tests on the Win2012 server?

PR #776, opened 2016-02-15
Add animation example (WPF) based on extension methods. This can be a…

GeertvanHorrik:
… great starting point for people that want to add animations to their plots without requiring OxyPlot to actually implement this.


Source/Examples/WPF/WpfExamples/Examples/AnimationsDemo/AnimationExtensions.cs
```diff
@@ -0,0 +1,112 @@
...
+// --------------------------------------------------------------------------------------------------------------------
+
+
+namespace AnimationsDemo
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Threading.Tasks;
+    using System.Windows.Media.Animation;
```
objorke: could this code be made portable? it would be interesting to see it in the portable example library...

objorke:
very good!

Source/Examples/WPF/WpfExamples/Examples/AnimationsDemo/AnimationExtensions.cs
GeertvanHorrik: The only reason I have animation is because of the EasingFunctionBase because I needed an easy way to show different values. CalculateEaseValues can be edited so it has it's own implementation and can be portable :)

objorke:
the build error seems to be

monodroid error XA9001: Trial period has expired. [C:\projects\oxyplot\Source\OxyPlot.Xamarin.Android\OxyPlot.Xamarin.Android.csproj]
C:\Program Files (x86)\MSBuild\Xamarin\Android\Xamarin.Android.Common.targets(529,2): 
error XA9010: Unable to determine license edition. 

but the Xamarin account (business plan) seems to be ok...
I will merge this PR anyway.

Issue #497, opened 2015-05-30
This PlotModel is already in use by some other PlotView control.


sunwayking:
This issue occures at line 165 in PlotViewEx.cs.
As my thought, reattaching same plotmodel to the same plotview should not throw this InvalidOperationException showing on title.
Then a possible solution is to avoid throwing InvalidOperationException and reattaching plotmodel when this.currentModel.PlotView equals current plotview object(this).

Code modification is here:

```C#
private void OnModelChanged()
{
    lock (this.modelLock)
    {
        if (this.currentModel != null)
        {
            ((IPlotModel)this.currentModel).AttachPlotView(null);
        }

        this.currentModel = this.Model;

        if (this.currentModel != null && !Equals(this.currentModel.PlotView, this))
        {
            if (this.currentModel.PlotView != null)
            {
                throw new InvalidOperationException(
                    "This PlotModel is already in use by some other PlotView control.");
            }

            ((IPlotModel)this.ActualModel).AttachPlotView(this);
        }
    }

    this.InvalidatePlot();
}
```

tibel:
I think this use case should never happen. Can you describe your scenario, please.

HippoBaro:
@tibel I had experiended those errors using Xamarin.Forms when creating multiple instances of the same ContentPage (including a PlotView) with different data.
Xamarin.Forms does a great job in reusing existing UIViewController to save memory, but it can cause OxyPlot to raise this exception.

tibel:
I have moved the check to `PlotModel.AttachPlotView()` and allow attaching the same `plotView` again.

staylr:
In OxyPlot.Xamarin.{Android,iOS}.PlotView you also need to clear the Model when disposing. Not sure what the solution is (if needed) on Windows Phone.


        /// <summary>
        /// Dispose the instance.
        /// </summary>
        /// <param name="disposing">Is the instance disposing.</param>
        protected override void Dispose(bool disposing)
        {
            if (disposing)
            {
                this.Model = null;
            }

            base.Dispose(disposing);
        }

omsara:
Bonjour;
Actuellement j'utilise la version "OxyPlot.Xamarin.Forms.2015.1.889-alpha"; et j'ai le même soucis cité au début; y-a-t'il une solution  ? 
 et aussi j'ai un autre soucis sur cette même version  le plot ne s'affiche pas au niveau de windows-phone ? 
merci de votre aide .

BWouters:
[...]

PR #780, opened 2016-02-16
"This PlotModel is already in use by some other PlotView control"

icerocker:
Ref. issue [#497](https://github.com/oxyplot/oxyplot/issues/497). I experience the same issue when using oxyplot in a WPF DataGrid DataTemplate.

I've added a demo application which can trigger this bug. The application lists 10 rows in a DataGrid. Expand multiple rows (Ctrl+Click). When scrolling (either with the scroll wheel or using the scroll bar) the application crashes with "XamlParseException - This PlotModel is already in use by some other PlotView control". This application only instanciates a PlotModel, nothing else oxyplot related to keep it as simple as possible. I did get this exception when running a more complete program too.

(Not sure if this is the correct way of reporting a bug with code attached)


objorke:
Thanks for creating a good example!
I have some requests:
- move the code into the WpfExamples project (https://github.com/oxyplot/oxyplot/tree/develop/Source/Examples/WPF/WpfExamples/Examples) - we must avoid "project bloat" in the solution. In the past, all these examples were different projects and the solution took much longer time to build
- do not depend on NuGet packages inside this repository (dependencies will already be there for the WpxExamples project)

objorke:
I think you have to rebase and `git push -f` this PR

objorke:
The code must be added to the WpfExamples project.

tibel:
I have integrated your example in WpfExamples

Issue #679, opened 2015-12-12
Add `RenderInLegend` property


ch-hristov:
I would like to remove some of the series from the legend, currently I can only see that only all of the series of the PlotModel are rendered.

objorke:
If the `Title` property is not specified, the series should not be shown in the legend.

interdrift:
Verified, works as intended. It might be better to expose a dependancy property on Series named RenderInLegend since its a bit more intuitive.

objorke:
This will be a breaking change. 
I see the possibility to have a title that is not shown in the legend but shown in the tracker. That's a valid enhancement issue :-)

benjaminrupp:
I agree! I´d like to have a series not rendered in legend but in tracker.

VisualMelon:
Can this be closed?

PR #781, opened 2016-02-17
#679 RenderInLegend property in series

benjaminrupp:
#679


objorke:
oops, I think you forgot to update the change log...

objorke:
I can handle it later

benjaminrupp:
Oh. Your are right. My fault. Sorry.

PR #796, opened 2016-02-26
Revert platform specification changes in appveyor.yml

objorke:


objorke:
I forgot that this does not work...

PR #797, opened 2016-02-26
Update WPF png exporter tests

objorke:
#759 
#772


objorke:
This solution is not ready to be merged...

Issue #802, opened 2016-02-27
Heatmaps don't support Logarithmically distributed data


VisualMelon:
The HeatMapSeries class assumes that the data it is supplied with is linearly and uniformly distributed. This results in ill-computed Min/Max values for the series, and so an offset and incorrectly scaled result, as well as misplaced Labels and tracking functionality.

I would suggest (and am currently testing) a solution that allows the setting of properties to determine whether the X and/or Y axis is logarithmically distributed, and performs all rendering computation in Screen Space (to save many Exp and Log calls).

Issue #801, opened 2016-02-27
Can't render Heatmaps uninterpolated to PDF


VisualMelon:
Rendering a Heatmap to PDF results in an interpolated bitmap when HeatMapSeries.Interpolate is set to false, presumably due to a limitation of the platform or renderer.

A solution has been proposed which renders a set of discrete rectangles in place of the current bitmap.

Issue #535, opened 2015-07-12
Heatmap Transpose not working


AlexeiScherbakov:
If I swap Position of X and Y axis in plot axis are changed, but data is drawing same as not transposed variant. I expect that, HeatMapSeries take axis by XAxisKey & YAxisKey. But HeatMapSeries always draws as XAxis is located left/right and Y is top/bottom

 How I can add sample for problem demonstration?

objorke:
Please see the "Issues" models in the example library. Please add a similar sample!

AlexeiScherbakov:
I'm added problem demonstration to my fork https://github.com/AlexeiScherbakov/oxyplot
It consist of 2 issues:
1 - Transposed heatmap demonstration
2 - Modified Peak example with not transposed plotmodel

VisualMelon:
It seems this can be solved by reorienting Screen Space coordinates when they are Transformed and (Inverse) Transformed, flipping the X and Y coordinates if the X axis is determined to be vertical.

VisualMelon:
I believe this issue be closed.

Issue #164, opened 2014-08-14
HeatMapSeries not correct on non-linear axis


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10239](http://oxyplot.codeplex.com/workitem/10239)**

**[abertier](http://www.codeplex.com/site/users/view/abertier)** 2014-08-08 14:31
When adding a heatmapseries to a plotmodel which has a non-linear X-axis, the heatmap is still drawn as if it was linear.

PR #803, opened 2016-02-28
Address HeatMapSeries issues #802, #801, and #535

VisualMelon:
# Address issues #802, #801, and #535

## Implements handling for Logarithmic data space (#802)

LogX and LogY properties are introduced, each indicating whether or not the appropriate dimension should be treated as logarithmic (that is, the immaterial DataPoints corresponding to each entry in the Data are uniformly distributed along a log axis, each is a constant times the last)

## Implements rendering of HeatMapSeries with rectangles (#801)

Only employed when Interpolate is set to false, and RenderDiscreteRectangles is set to true

## Implements support for vertical X axes in HeatMapSeries (#535)

This should solve the transposition issue, though the manner by which it is done may prove problematic

## Adds a couple of example Logarithmic HeatMaps

One of these uses DiscreteRectangles

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,11 +13,6 @@ OxyPlot is a cross-platform plotting library for .NET
 
 ![License](https://img.shields.io/badge/license-MIT-red.svg)
 [![Build status](https://img.shields.io/appveyor/ci/objorke/oxyplot.svg)](https://ci.appveyor.com/project/objorke/oxyplot)
-![NuGet downloads](https://img.shields.io/nuget/dt/oxyplot.core.svg)
-![Version](https://img.shields.io/nuget/v/OxyPlot.Core.svg)
-![Pre-release version](https://img.shields.io/nuget/vpre/OxyPlot.Core.svg)
-![GitHub release](https://img.shields.io/github/release/oxyplot/oxyplot.svg)
-![GitHub tag](https://img.shields.io/github/tag/oxyplot/oxyplot.svg)
 
 ![Plot](http://oxyplot.org/public/images/normal-distributions.png)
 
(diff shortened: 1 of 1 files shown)


Source/OxyPlot/Series/HeatMapSeries.cs
```diff
@@ -122,6 +122,21 @@ public HeatMapSeries()
...
 
         /// <summary>
+        /// Gets or sets a value indicating whether to use discrete rectangles when rendering (ignored if Interpolate is set to true). The default value is <c>false</c>.
+        /// </summary>
+        public bool RenderDiscreteRectangles { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to treat the x dimension as logarithmic. The default value is <c>false</c>.
+        /// </summary>
+        public bool LogX { get; set; }
```
objorke: can´t this be determined from the actual axis?

Source/OxyPlot/Series/HeatMapSeries.cs
```diff
@@ -122,6 +122,21 @@ public HeatMapSeries()
         public bool Interpolate { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to use discrete rectangles when rendering (ignored if Interpolate is set to true). The default value is <c>false</c>.
+        /// </summary>
+        public bool RenderDiscreteRectangles { get; set; }
```
objorke: what about changing this to an enum: RenderingMethod { Image, Rectangles }?

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 ### Added
+- Added support for vertical X axis to HeatMapSeries (#535)
```
objorke: is #535 covered by an example?

objorke:
Very good! I have added some comments. Ideally, I would also like to see each issue solved in a separate commit.

VisualMelon:
My apologies indeed, I did have each issue solved separately, but then I read CONTRIBUTING.md and misinterpreted the squashing of logical commits - I am happy to recommit each solution if that would be helpful.

CHANGELOG.md
VisualMelon: It is not, though the example for issue #535 does demonstrate it neatly at this time.

Source/OxyPlot/Series/HeatMapSeries.cs
VisualMelon: I would not be whole against this, but since Rectangles don't suppose Interpolation at this time, it seems generous to elevate them to prestige of having an Enum named after them. If it is anticipated that they may somehow support interpolation in the future (perhaps with vast numbers of rectangles... this doesn't sound healthy though) then sorting the interface out now would be sensible.

Source/OxyPlot/Series/HeatMapSeries.cs
VisualMelon: I would be hesitant to suppose this given #164, though I do think that plotting a linearly defined HeatMap on a log plot is unlikely to occur (it is something this commit makes no effort to address, and isn't compatible, so to speak, with the method of using Screen Space coordinates to place the rectangles). If this would be deemed desirable, then I'd gladly look into it, though I can't at this time think how we would detect whether an axis is logarithmic or not, short of checking the type and inferring the mode from the Translate/InverseTranslate methods, neither of which I think very nice, though the I see that the later is employed by the HeatMap image generation to determine whether the X or Y axes need to be reversed.

Source/OxyPlot/Series/HeatMapSeries.cs
```diff
@@ -170,6 +185,58 @@ public void Invalidate()
...
+            return Orientate(base.Transform(p));
+        }
+
+        /// <summary>
+        /// Transforms orientated ScreenSpace coords to DataSpace coors
+        /// </summary>
+        public new DataPoint InverseTransform(ScreenPoint point)
+        {
+            return base.InverseTransform(Orientate(point));
+        }
```
VisualMelon: I would be interested to hear some comments on these methods - I can't say I'm especially fond of hiding underlying methods like this, but intercepting these does, I think, make sense, and would be applicable to other XYAxisSeries types (line and scatter, for example) which also struggle with Transposing, as per #535. I am wondering if these 'overloads' could be moved into XYAxisSeries, or some intermediary (by way of example, LineSeries can be 'fixed' by just pasting these shadow methods into LineSeries.cs, though there may be some marker details that need attention). I should highlight that while HeatMapSeries does have to work around this fix itself, it shouldn't be a common issue.

PR #805, opened 2016-02-28
Address HeatMap issues #802, #801, and #535

VisualMelon:
Address issues #802, #801, and #535 concerning HeatMaps, and provide examples of the new functionality.

_Apologies again, I managed to delete the old Pull Request without realising while sorting out the commit history_

### DIFF
--- a/Source/OxyPlot/Series/HeatMapSeries.cs
+++ b/Source/OxyPlot/Series/HeatMapSeries.cs
@@ -120,7 +120,12 @@ namespace OxyPlot.Series
         /// </summary>
         /// <remarks>This property is not supported on all platforms.</remarks>
         public bool Interpolate { get; set; }
-        
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to use discrete rectangles when rendering (ignored if Interpolate is set to true). The default value is <c>false</c>.
+        /// </summary>
+        public bool RenderDiscreteRectangles { get; set; }
+
         /// <summary>
         /// Gets or sets a value indicating whether to treat the x dimension as logarithmic. The default value is <c>false</c>.
         /// </summary>
@@ -238,21 +243,49 @@ namespace OxyPlot.Series
             var s00 = this.Transform(left, bottom);
             var s11 = this.Transform(right, top);
             var rect = new OxyRect(s00, s11);
-            
+
+
+            bool needImage = Interpolate || !RenderDiscreteRectangles;
+
             var currentDataHash = this.Data.GetHashCode();
             var currentColorAxisHash = this.ColorAxis.GetElementHashCode();
-            if (this.image == null || currentDataHash != this.dataHash || currentColorAxisHash != this.colorAxisHash)
+            if ((needImage && this.image == null) || currentDataHash != this.dataHash || currentColorAxisHash != this.colorAxisHash)
             {
-                this.UpdateImage();
+                if (needImage)
+                    this.UpdateImage();
                 this.dataHash = currentDataHash;
                 this.colorAxisHash = currentColorAxisHash;
             }
 
             var clip = this.GetClippingRect();
-            if (this.image != null)
+            if (needImage)
             {
-                rc.DrawClippedImage(clip, this.image, rect.Left, rect.Top, rect.Width, rect.Height, 1, this.Interpolate);
+                if (this.image != null)
+                {
+                    rc.DrawClippedImage(clip, this.image, rect.Left, rect.Top, rect.Width, rect.Height, 1, this.Interpolate);
+                }
             }
+            else
+            {
+                double sdx = (s11.X - s00.X) / m;
+                double sdy = (s11.Y - s00.Y) / n;
+
+                // draw lots of rectangles
+                for (int i = 0; i < m; i++)
+                {
+                    for (int j = 0; j < n; j++)
+                    {
+                        OxyColor rectcolor = this.ColorAxis.GetColor(this.Data[i, j]);
+
+                        var pointa = new ScreenPoint(s00.X + i * sdx, s00.Y + j * sdy);
+                        var pointb = new ScreenPoint(s00.X + (i + 1) * sdx, s00.Y + (j + 1) * sdy);
+                        OxyRect rectrect = new OxyRect(pointa, pointb);
+
+                        rc.DrawClippedRectangle(clip, rectrect, rectcolor, OxyColors.Undefined, 0);
+                    }
+                }
+            }
+
 
             if (this.LabelFontSize > 0)
             {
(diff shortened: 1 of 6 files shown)


Source/OxyPlot/Series/HeatMapSeries.cs
```diff
@@ -172,6 +187,58 @@ public void Invalidate()
...
+            return Orientate(base.Transform(p));
+        }
+
+        /// <summary>
+        /// Transforms orientated ScreenSpace coords to DataSpace coors
+        /// </summary>
+        public new DataPoint InverseTransform(ScreenPoint point)
+        {
+            return base.InverseTransform(Orientate(point));
+        }
```
VisualMelon: These methods implement a fix for #535, the Transposing issue, which is also an issue for series, for example Line and Scatter. I am not very fond of this solution, but I think it does make sense, and is likely applicable to other series types.

Source/OxyPlot/Series/HeatMapSeries.cs
```diff
@@ -122,6 +122,21 @@ public HeatMapSeries()
         public bool Interpolate { get; set; }
 
         /// <summary>
+        /// Gets or sets a value indicating whether to use discrete rectangles when rendering (ignored if Interpolate is set to true). The default value is <c>false</c>.
+        /// </summary>
+        public bool RenderDiscreteRectangles { get; set; }
```
VisualMelon: It was suggested in a previous PullRequest that this might be replaced with an Enum, selecting either rectangles or image rendering. I am not wholly against this, but rectangles do not lend themselves to interpolated rendering, so I would be disinclined to give them the same prestige as the image rendering.

Source/OxyPlot/Series/HeatMapSeries.cs
```diff
@@ -122,6 +122,21 @@ public HeatMapSeries()
...
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to treat the x dimension as logarithmic. The default value is <c>false</c>.
+        /// </summary>
+        public bool LogX { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to treat the y dimension as logarithmic. The default value is <c>false</c>.
+        /// </summary>
+        public bool LogY { get; set; }
```
VisualMelon: It was suggested in a previous PullRequest that it may be possible to detect these automatically from the axes configuration. I am dubious of this, in part because of the existence of #164 (which these changes make no effort to address). I can also not think of a nice way of determining if the axes are logarithmic or not.

Source/OxyPlot/Series/HeatMapSeries.cs
objorke: What will happen if a 3rd method of rendering heat maps is implemented? Then a bool property will have to be marked obsolete, I think.

I think it is better to do one of the following
- use an enum
- create a new class
- fall back automatically to rectangle rendering if image rendering will not work (e.g. log axes) or is not supported on the current platform

Source/OxyPlot/Series/HeatMapSeries.cs
objorke: I understand that these properties say something about the data. 
E.g. if you have a log axis from 1 to 100 and you have heat map data with 3 points, then the `LogX` and `LogY` properties tell if the data should be given at positions (1, 10, 100) or (1, 50.5, 100). Is that right?

objorke:
Thanks for reworking the commits! I will pull your fork locally and have a look at this! I am a little concerned about the `RenderDiscreteRectangles` property

Source/OxyPlot/Series/HeatMapSeries.cs
VisualMelon: I did some testing today, and just copying these shadow methods into LineSeries and Column appears to make them behave as one would expect them (though I appreciate that Column charts tend to have columns, not bars) - ScatterSeries is also easy 'fixed' with these methods, but the calls to XAxis.Transform() have to replaced with calls to the shadow methods (I can't see any reason for not using the base Transform(double, double) as it is), and a similar story for AreaSeries. I shall continue to look into this, because it would be nice if this could be pushed down to XYAxisSeries or an intermediate class.

objorke:
Can you please follow the coding style of this project? 
I want it to be stylecop compliant, and will have to clean it up later if you don't do it...
The main style issues are: missing "this." (SA1126), order of methods (SA1202), body of if statements not in {},

objorke:
I fixed it for you!

VisualMelon:
Certainly, I meant to run it though StyleCop but forgot having installed it. Is it acceptable to fix the styling issues in a single commit on top of the others?

objorke:
Yes, I accept a commit on top of the others.

objorke:
Are contour series also supporting "transposed" axes? If not, we should add an issue. Should be easy to test in the 6x4 examples.

Source/OxyPlot/Series/HeatMapSeries.cs
VisualMelon: Indeed, if you have X0 = 1, X1 = 100, and 3 data points, then LogX = true indicates that in Data Space they are at x=1, x=10, x=100 - but the code assumes that they will be rendered onto a log axis, so they are positioned at x = 1, x=50.5, x=100 in Screen Space. Because the code assumes the axis follows suit and is logarithmic, rendering involves simply computing the MinX and MinY, and linearly interpolated between the two to find ordinates in Screen Space.

Attached are a couple of PDFs that are the motivation for all this work which show what I consider sensible behaviour, and both generated with my local copy of OxyPlot (these changes and none further to the HeatMap code).

[linxyheatmap.pdf](https://github.com/oxyplot/oxyplot/files/150186/acctest_2.pdf)
[logxyheatmap.pdf](https://github.com/oxyplot/oxyplot/files/150185/acctest_.pdf)

Source/OxyPlot/Series/HeatMapSeries.cs
VisualMelon: Ah, yes, I suppose if we may have more than two rendering modes, that would make sense. I would be in favour of creating an enum, given just how much code is shared that has nothing to do with rendering.

Image rending does (by my reckoning) work with the log axes, I attach an interpolated version of the logarithmic dataset PDF supplied below.

[interpolatedlogxyheatmap.pdf](https://github.com/oxyplot/oxyplot/files/150190/acctest_2img.pdf)

VisualMelon:
The ContourSeries does not support transposed axes at current. However, copying and pasting the transposing shadow methods from HeatMapSeries appears to result in a well formed transposed Peak example (including the text alignment).

Source/OxyPlot/Series/HeatMapSeries.cs
VisualMelon: Another node, is that while the HeatMap labels end up in the right place, trying to select them (in a PDF) results in there underlying ordering becoming apparent - this would suggest that the approach of simply repositioning them is not adequate, and for things like text more careful measures may be warranted.

Source/OxyPlot/Series/HeatMapSeries.cs
objorke: Change it to an enum, and I will accept this PR!

Source/OxyPlot/Series/HeatMapSeries.cs
objorke: the last pdf is blank

Something is strange when LogX is false, see d44dbf7

PR #817, opened 2016-03-27
Area series plots only visible range. Added two colors mode for AreaS…

olegtarasov:
Fixes # .

### Checklist

+  I have included examples or tests
+ I have updated the change log
+ I am listed in the CONTRIBUTORS file
+ I have cleaned up the commit history (use rebase and squash)

Area series plots only visible range. Added two colors mode for AreaSeries. See the provided example.

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Series/AreaSeriesExamples.cs
+++ b/Source/Examples/ExampleLibrary/Series/AreaSeriesExamples.cs
@@ -32,7 +32,20 @@ namespace ExampleLibrary
             return plotModel1;
         }
 
-        [Example("Crossing lines")]
+		[Example("Two colors mode")]
+		public static PlotModel TwoColorsMode()
+		{
+			var plotModel1 = new PlotModel { Title = "AreaSeries with different area colors" };
+			var areaSeries1 = CreateExampleAreaSeries();
+			areaSeries1.TwoColorMode = true;
+			areaSeries1.ConstantY2 = 60;
+			areaSeries1.Color = OxyColors.Red;
+			areaSeries1.Color2 = OxyColors.Blue;
+			plotModel1.Series.Add(areaSeries1);
+			return plotModel1;
+		}
+
+		[Example("Crossing lines")]
         public static PlotModel CrossingLines()
         {
             var plotModel1 = new PlotModel { Title = "AreaSeries with crossing lines" };
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -331,6 +331,8 @@ namespace OxyPlot.Wpf
             StreamGeometryContext sgc = null;
             PathGeometry pathGeometry = null;
             int count = 0;
+	        bool fillDefined = !fill.IsUndefined();
+	        bool strokeDefined = !stroke.IsUndefined();
 
             foreach (var polygon in polygons)
             {
@@ -338,7 +340,7 @@ namespace OxyPlot.Wpf
                 {
                     path = this.CreateAndAdd<Path>();
                     this.SetStroke(path, stroke, thickness, lineJoin, dashArray, 0, aliased);
-                    if (!fill.IsUndefined())
+                    if (fillDefined)
                     {
                         path.Fill = this.GetCachedBrush(fill);
                     }
@@ -363,14 +365,14 @@ namespace OxyPlot.Wpf
                     {
                         if (usg)
                         {
-                            sgc.BeginFigure(point, !fill.IsUndefined(), true);
+                            sgc.BeginFigure(point, fillDefined, true);
                         }
                         else
                         {
                             figure = new PathFigure
                             {
                                 StartPoint = point,
-                                IsFilled = !fill.IsUndefined(),
+                                IsFilled = fillDefined,
                                 IsClosed = true
                             };
                             pathGeometry.Figures.Add(figure);
@@ -382,11 +384,11 @@ namespace OxyPlot.Wpf
                     {
                         if (usg)
                         {
-                            sgc.LineTo(point, !stroke.IsUndefined(), true);
+							sgc.LineTo(point, strokeDefined, true);
                         }
                         else
                         {
-                            figure.Segments.Add(new LineSegment(point, !stroke.IsUndefined()) { IsSmoothJoin = true });
+							figure.Segments.Add(new LineSegment(point, strokeDefined) { IsSmoothJoin = true });
                         }
                     }
                 }
--- a/Source/OxyPlot/OxyPlot.csproj
+++ b/Source/OxyPlot/OxyPlot.csproj
@@ -66,6 +66,7 @@
     <Compile Include="Rendering\OxySizeExtensions.cs" />
     <Compile Include="Reporting\PlotModelExtensions.cs" />
     <Compile Include="Series\BarSeries\LinearBarSeries.cs" />
+    <Compile Include="Series\DataPointHelper.cs" />
     <Compile Include="Series\IScatterPointProvider.cs" />
     <Compile Include="Input\Gestures\OxyMouseEnterGesture.cs" />
     <Compile Include="PlotController\EventArgs\OxyInputEventArgs.cs" />
--- a/Source/OxyPlot/Series/AreaSeries.cs
+++ b/Source/OxyPlot/Series/AreaSeries.cs
@@ -9,8 +9,8 @@
 
 namespace OxyPlot.Series
 {
-    using System;
-    using System.Collections.Generic;
+	using System;
+	using System.Collections.Generic;
     using System.Linq;
 
     /// <summary>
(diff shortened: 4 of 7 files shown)


objorke:
There is already a `TwoColorAreaSeries` - what is the difference?
https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Series/TwoColorAreaSeries.cs
Can you create an issue describing the desired functionality?

Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -331,6 +331,8 @@ public void DrawEllipses(IList<OxyRect> rectangles, OxyColor fill, OxyColor stro
             StreamGeometryContext sgc = null;
             PathGeometry pathGeometry = null;
             int count = 0;
+	        bool fillDefined = !fill.IsUndefined();
```
objorke: good!

olegtarasov:
Created an issue #819 with a description of desired functionality.

Source/OxyPlot/OxyPlot.csproj
```diff
@@ -66,6 +66,7 @@
     <Compile Include="Rendering\OxySizeExtensions.cs" />
     <Compile Include="Reporting\PlotModelExtensions.cs" />
     <Compile Include="Series\BarSeries\LinearBarSeries.cs" />
+    <Compile Include="Series\DataPointHelper.cs" />
```
objorke: the SL5 and NET40 projects must also be updated

Source/Examples/ExampleLibrary/Series/AreaSeriesExamples.cs
```diff
@@ -32,7 +32,20 @@ public static PlotModel DifferentColors()
             return plotModel1;
         }
 
-        [Example("Crossing lines")]
+		[Example("Two colors mode")]
```
objorke: Thanks for the example! But see my comment on #819. This can be done by two `AreaSeries`. It would be good to have an example that cannot be created by two `AreaSeries`.

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -9,8 +9,8 @@
 
 namespace OxyPlot.Series
 {
-    using System;
-    using System.Collections.Generic;
+	using System;
```
objorke: SA1027 violation. Please follow the current coding style! :-) Otherwise I need to make a commit simply to fix the coding style...

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -33,14 +33,25 @@ public class AreaSeries : LineSeries
...
         private List<DataPoint> actualPoints2;
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref = "AreaSeries" /> class.
-        /// </summary>
-        public AreaSeries()
+		/// <summary>
+		/// The index of the data item at the start of visible window
+		/// </summary>
+		private int winIndex;
```
objorke: how much performance do you gain by persisting this? I am concerned about premature optimization...

objorke: The comment is good, but the name of the field does not tell me much...

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -123,6 +152,15 @@ public List<DataPoint> Points2
...
         public bool Reverse2 { get; set; }
 
+		/// <summary>
+		/// Gets or sets a value indicating whether two color mode should be used.
+		/// </summary>
+		/// <value><c>true</c> if two color mode should be used; otherwise, <c>false</c>.</value>
+		/// <remarks> When using two color mode, Points and Points2 collections are considered
+		/// as different areas. They are painted with different colors and ConstantY2 is used to divide
+		/// the polygons horizontally. </remarks>
+		public bool TwoColorMode { get; set; }
```
objorke: I think an enum would be better here. If someone comes up with a third method we will get in trouble by a bool, I think.

objorke: But as commented in #819, I think the two color mode belongs in `TwoColorAreaSeries`...

Source/OxyPlot/Series/BarSeries/RectangleBarSeries.cs
```diff
@@ -152,15 +157,25 @@ public override void Render(IRenderContext rc)
...
+			var xmax = this.XAxis.ActualMaximum;
+			this.winIndex = FindBarIndex(Items, xmin, this.winIndex);
+
+	        if (this.winIndex > 0)
+	        {
+		        this.winIndex--;
+	        }
 
-            int i = 0;
+			var clippingRect = this.GetClippingRect();
```
objorke: please create a separate branch for changes to the RectangleBarSeries.
Also, create a separate issue for this change.
This is performance optimizations, right?

Source/OxyPlot/Series/BarSeries/RectangleBarSeries.cs
```diff
@@ -302,5 +324,75 @@ protected virtual bool IsValid(double v)
...
+		/// <param name='targetX'>
+		/// target x.
+		/// </param>
+		/// <param name='guess'>
+		/// initial guess index.
+		/// </param>
+		/// <returns>
+		/// index of x with max(x) &lt;= target x or -1 if cannot find
+		/// </returns>
+		public static int FindBarIndex(IList<RectangleBarItem> items, double targetX, int guess)
```
objorke: This looks like a duplicate of the `FindIndex` method below. I think you can create a generic version - with a `Func<TItem, double>` to give the X0.

Source/OxyPlot/Series/DataPointHelper.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System.Collections.Generic;
```
objorke: SA1200 violation

Source/OxyPlot/Series/DataPointHelper.cs
```diff
@@ -0,0 +1,80 @@
...
+		/// <param name='targetX'>
+		/// target x.
+		/// </param>
+		/// <param name='guess'>
+		/// initial guess index.
+		/// </param>
+		/// <returns>
+		/// index of x with max(x) &lt;= target x or -1 if cannot find
+		/// </returns>
+		public static int FindIndex(List<DataPoint> items, double targetX, int guess)
```
objorke: I am a bit reluctant to adding such static methods to the library. But I see it can be used in two series types, then it might be ok...

objorke: This method requires unit tests, in my opinion.

Source/OxyPlot/Series/DataPointHelper.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System.Collections.Generic;
+
+namespace OxyPlot.Series
+{
+	/// <summary>
+	/// Data point helper functions.
+	/// </summary>
+	public class DataPointHelper
```
objorke: I think this class/method belongs under the Utilities folder. Or could it be a protected member under the DataPointSeries?

objorke:
Hi Oleg, this is a great contribution, but you see I am not quite satisfied yet... I would really like to merge the performance optimizations (could these be a separate pull request?) and also the two color mode, but we need to see that it has general value and that it does not make the API difficult to maintain in the future! :-)

Source/OxyPlot/Series/AreaSeries.cs
olegtarasov: Persisting the last visible window index is important as it allows for constant rendering time at any position. When we are rendering a window at the start of the series, we find start index pretty quickly. But when we are closer to the end, it takes more time to find the index. Persisting the last index gives us a starting point for the search, since there is a high chance that next rendering window will be near the last one (zooming and panning mostly occurs gradually with small deltas).

Source/OxyPlot/Series/BarSeries/RectangleBarSeries.cs
olegtarasov: Sorry for that, I didn't intend for RectangleBarSeries code to get into this pull request at all — I guess, GitHub adds latest commits to an open pull request automatically. That was only the first commit and RectangleBarSeries needs more work.

Source/OxyPlot/Series/BarSeries/RectangleBarSeries.cs
olegtarasov: Yes, this is a problem that I contemplated upon. This method is performance-critical as it is called on every render event and it uses property getters excessively. I'm afraid that adding an additional function call to get a coordinate will impact performance. I haven't had the time to test it with performance profiler yet, I think I'll do it today. Maybe there is another way to make this method more generic so that I don't need to copy-paste it for every data item class.

Source/OxyPlot/Series/DataPointHelper.cs
olegtarasov: Yes, as I noted in a previous comment, ideally this method should be generic and have unit tests covering it. I didn't write this method myself, I took it from CandleBarSeries and adapted it to DataPoints.

olegtarasov:
Thank you for the feedback! I'm closing this pull request and will open a new one when I rework the code. Looking forward to continue the discussion in issue #819 :)

Issue #133, opened 2014-08-14
Improve automatic MinorStep when MajorStep is 2


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10188](http://oxyplot.codeplex.com/workitem/10188)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-04-28 10:39
When MajorStep is 2, the automatic MinorStep should be 0.5, not 0.4. 

Divide by 4 not by 5 in this case.

VisualMelon:
Can this be closed?

Jonarw:
Yes, I would think so.

PR #831, opened 2016-04-01
Change MinorInterval calculation, add unit test (#133)

Jonarw:
I hope I got it right this time, this whole GitHub thing is still very confusing to me ;)

Is the unit test ok? That's also something I do not have a lot of experience with...

### DIFF
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -639,6 +639,24 @@ namespace OxyPlot.Axes
         protected double ViewMinimum { get; set; }
 
         /// <summary>
+        /// Calculates the minor interval.
+        /// </summary>
+        /// <param name="majorInterval">The major interval.</param>
+        /// <returns>The minor interval.</returns>
+        public double CalculateMinorInterval(double majorInterval)
+        {
+            // check if majorInterval = 2*10^x
+            // uses the mathematical identity log10(2 * 10^x) = x + log10(2)
+            // -> we just have to check if the modulo of log10(2*10^x) = log10(2)
+            if (Math.Abs(((Math.Log10(majorInterval) + 1000) % 1) - Math.Log10(2)) < 1e-10)
+            {
+                return majorInterval / 4;
+            }
+
+            return majorInterval / 5;
+        }
+
+        /// <summary>
         /// Creates tick values at the specified interval.
         /// </summary>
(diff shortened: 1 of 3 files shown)


Source/OxyPlot/Axes/Axis.cs
```diff
@@ -639,6 +639,24 @@ protected internal double ActualTitleFontWeight
...
 
         /// <summary>
+        /// Calculates the minor interval.
+        /// </summary>
+        /// <param name="majorInterval">The major interval.</param>
+        /// <returns>The minor interval.</returns>
+        public double CalculateMinorInterval(double majorInterval)
+        {
+            // check if majorInterval = 2*10^x
+            // uses the mathematical identity log10(2 * 10^x) = x + log10(2)
```
objorke: thanks for better explanation of this expression!

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -1679,31 +1697,6 @@ protected double CalculateActualInterval(double availableSize, double maxInterva
...
-        protected double CalculateMinorInterval(double majorInterval)
-        {
-            // if major interval is 100, the minor interval will be 20.
-            return majorInterval / 5;
-
-            // The following obsolete code divided major intervals into 4 minor intervals, unless the major interval's mantissa was 5.
-            // e.g. Major interval 100 => minor interval 25.
-
-            // Func<double, double> exponent = x => Math.Ceiling(Math.Log(x, 10));
-            // Func<double, double> mantissa = x => x / Math.Pow(10, exponent(x) - 1);
```
objorke: isn't this doing almost the same as the log10 + modulo?

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -1679,31 +1697,6 @@ protected double CalculateActualInterval(double availableSize, double maxInterva
...
-            // var m = (int)mantissa(majorInterval);
-            // switch (m)
-            // {
-            // case 5:
```
objorke: if "case 1:" was added, wouldn't this give the same result?
which code is easier to read?

Jonarw: Yes, I guess it would do the exact same thing. I find my code to be more intuitive, but then again, I wrote it, so I'm not to judge. I guess you choose which one you like better (and I'm not mad either case) :)

Issue #819, opened 2016-03-28
Need an ability to create area plots with two polygons of different colors


olegtarasov:
### Feature description

I have a use case when I need to plot an area chart with two polygons filled with different colors. First polygon is built using Points collection and ConstantY value. Second polygon is built with ConstantY and Points2 collection. Furthermore, data points in Points and Points2 collections may have identical X values. So TwoColorAreaSeries is not suitable for this task as it only has one data point collection and two polygons can't have same X values. In pull request #817 I propose a solution where AreaSeries can solve this task, meanwhile greatly improving rendering performance. With my pull request I attached an example which shows the use case I need.

If you decide to go forward with this pull request, TwoColorAreaSeries will largely become obsolete as AreaSeries will solve the same problem with more flexibility.

objorke:
I don't think two color mode should be added to the `AreaSeries`. In my opinion this series should only do rendering of single colored areas. The outline of the area should be defined by `Points` and `Points2` or `Points` and `ConstantY`. But it would be great if the `AreaSeries` rendering was general enough to allow customizations that rendered the area by two colors, three colors or *n* colors.

Another approach could be to add support for `Points2` in the `TwoColorAreaSeries`. Is this possible?
I see `TwoColorAreaSeries` is derived from `TwoColorLineSeries` - maybe it should be derived from `AreaSeries` instead?

objorke:
I added two `AreaSeries` to the same model to get exactly the same output as the example in #817:
```csharp
            var plotModel1 = new PlotModel { Title = "AreaSeries with different area colors" };
            var areaSeries1 = new AreaSeries
                                  {
                                      Color = OxyColors.Red,
                                      Color2 = OxyColors.Undefined,
                                      ConstantY2 = 60
                                  };
            areaSeries1.Points.Add(new DataPoint(0, 50));
            areaSeries1.Points.Add(new DataPoint(10, 40));
            areaSeries1.Points.Add(new DataPoint(20, 60));

            var areaSeries2 = new AreaSeries
                                  {
                                      Color = OxyColors.Blue,
                                      Color2 = OxyColors.Undefined,
                                      ConstantY2 = 60
                                  };
            areaSeries2.Points.Add(new DataPoint(0, 60));
            areaSeries2.Points.Add(new DataPoint(5, 80));
            areaSeries2.Points.Add(new DataPoint(20, 70));

            plotModel1.Series.Add(areaSeries1);
            plotModel1.Series.Add(areaSeries2);
[...]

PR #832, opened 2016-04-01
Extracted visible window search code from CandleStickSeries and made a generic version in XYSeries. Used it to omptimize AreaSeries performance.

olegtarasov:
As part of issue #819 I optimized performance of AreaSeries. This is a separate pull request so changes are granular and visible.

@oxyplot/admins


olegtarasov:
Will pull the latest version and rebase.

PR #873, opened 2016-04-30
Color minor and major ticks differently (#417)

### LINKED ISSUE #417: Color minor and major ticks differently
I need to color major ticks the same color thar major grid lines
and minor ticks the same color same color that minor grid lines
but I can find only one single property ticklinecolor for both ticks
can we split it?

Decatf:
Fixes #417.

### Checklist

-  Color minor and major ticks differently #417
- This adds MinorTicklineColor to Axis. It defaults to Automatic color. In this case it takes the value of TicklineColor.

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
+++ b/Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
@@ -51,6 +51,29 @@ namespace ExampleLibrary
             return CreateTickStyleExample(TickStyle.Outside);
         }
 
+        [Example("TickStyle: Color major and minor ticks differently")]
+        public static PlotModel TickLineColor()
+        {
+            var plotModel1 = new PlotModel { Title = "Color major and minor ticks differently" };
+            plotModel1.Axes.Add(new LinearAxis
+            {
+                Position = AxisPosition.Left,
+                MajorGridlineThickness = 3,
+                MinorGridlineThickness = 3,
+                TicklineColor = OxyColors.Blue,
+                MinorTicklineColor = OxyColors.Gray,
+            });
+            plotModel1.Axes.Add(new LinearAxis
+            {
+                Position = AxisPosition.Bottom,
+                MajorGridlineThickness = 3,
+                MinorGridlineThickness = 3,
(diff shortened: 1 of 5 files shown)


Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
```diff
@@ -51,6 +51,29 @@ public static PlotModel TickStyleOutside()
...
+        [Example("TickStyle: Color major and minor ticks differently")]
+        public static PlotModel TickLineColor()
+        {
+            var plotModel1 = new PlotModel { Title = "Color major and minor ticks differently" };
+            plotModel1.Axes.Add(new LinearAxis
+            {
+                Position = AxisPosition.Left,
+                MajorGridlineThickness = 3,
+                MinorGridlineThickness = 3,
+                TicklineColor = OxyColors.Blue,
```
objorke: we should consider to rename this to MajorTicklineColor...

Source/OxyPlot/Axes/Rendering/AxisRendererBase.cs
```diff
@@ -181,9 +181,11 @@ public virtual void Render(Axis axis, int pass)
         /// <param name="axis">The axis.</param>
         protected virtual void CreatePens(Axis axis)
         {
+            var minorTickColor = axis.MinorTicklineColor.Equals(OxyColors.Automatic) ? axis.TicklineColor : axis.MinorTicklineColor;
```
objorke: you could also use axis.MinorTicklineColor.IsAutomatic() ? ...

Decatf: I've updated this in pull request #884

Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
Decatf: To me either makes sense. Prefixing with Major keeps consistent naming. But TicklineColor alone follows with the current behavior in that it represents the color of all the tick lines. That is unless the minor color is set to a value other than automatic.

Issue #891, opened 2016-05-19
PdfReportWriter doesn't close Stream


a5rGithub:
[reproduction]
* Create a Report.
* Use "PdfReportWriter" class to write to pdf.
* try open pdf file while client application still open.
* pdf reader will report that pdf file is still in use and can't open the file.
```
            string path = @"c:\Users\me\Desktop\test.pdf";
            var r = this.CreateReport(fileName);
            var reportStyle = new ReportStyle();
            using (var w = new PdfReportWriter(fileName))
            {
                w.WriteReport(r, reportStyle);
            }
            Process.Start(path);
```
[expectation]
* When "PdfReportWriter" gets disposed it should Close the Stream it opened.

PR #892, opened 2016-05-19
pdfreportwriter doesn;t close stream

a5rGithub:
fix for #891


Source/OxyPlot.Pdf/PdfReportWriter.cs
```diff
@@ -139,6 +139,7 @@ public virtual void Close()
             r.RenderDocument();
             r.PdfDocument.Save(this.Output);
             this.CleanTemporaryPlotFiles();
+            this.Output.Close();
```
objorke: Could there be an issue if someone use the ctor with stream and already close the stream themselves? I think it would be better to remove the constructor with path, and don't take the responsibility for closing the stream here...

objorke:
Please update the change log

Source/OxyPlot.Pdf/PdfReportWriter.cs
a5rGithub: if the ctor with stream is being used the close method of the stream can be called after the dispose of the pdfreportwriter. if the close method of the stream is being called before the dispose of the pdfreportwriter you get into trouble.

a5rGithub:
Ok, I am a bit new to github. So I didnt understand that a commit with a reference to 891 was automatically attached to this ticket as well. That is why i committed the changelog with two references.

objorke:
Thanks! I will squash and merge this PR..

PR #893, opened 2016-05-20
Add missing virtual keywords to allow overridding

scastria:
All of the touch/gesture based methods in ControllerBase seemed to be missing the virtual keyword.  All other methods had virtual.  In order for me to add custom tap behavior in Xamarin.Android, I need to override the touch/gesture methods here.


objorke:
I agree these methods should be consistent. But I also think it is possible to bind gestures and commands without creating a derived controller.

scastria:
Thanks.  I am slowly getting acquainted with gesture and commands and touch interaction within oxyplot on Xamarin.  I may eventually come to that conclusion as well, but yes, consistency is good too.

objorke:
I am not sure how good the gestures and commands implementation is, let us know if you see something that should be improved...

PR #894, opened 2016-05-22
Optimized TwoColorAreaSeries performance and added two polygon mode

olegtarasov:
As discussed in #819, I added two polygons mode to TwoColorAreaSeries.

### DIFF
--- a/Source/OxyPlot/Series/AreaSeries.cs
+++ b/Source/OxyPlot/Series/AreaSeries.cs
@@ -75,7 +75,7 @@ namespace OxyPlot.Series
         /// Gets the actual color of the line for the second data set.
         /// </summary>
         /// <value>The actual color.</value>
-        public OxyColor ActualColor2
+        public virtual OxyColor ActualColor2
         {
             get
             {
@@ -139,6 +139,11 @@ namespace OxyPlot.Series
         /// Gets or sets the last visible window start position in second data points collection.
         /// </summary>
         protected int WindowStartIndex2 { get; set; }
+
+        /// <summary>
+        /// Gets a value indicating whether Points2 collection was defined by user.
+        /// </summary>
+        protected bool IsPoints2Defined { get; private set; }
 
         /// <summary>
         /// Gets the nearest point.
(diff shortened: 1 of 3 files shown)


olegtarasov:
So how about a merge on this PR? :)

objorke:
Sorry I have not had much time available lately. Can anyone else have a look at this PR? It is quite a lot of changes! I keep it in my backlog!

objorke:
Sorry for the delay. I have reviewed this code now, and it looks ok. I am a bit worried about the growing complexity of the AreaSeries, but it seems to be well organized code with small and well composed methods. Thanks for contributing! And for including an example!

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -75,7 +75,7 @@ public AreaSeries()
         /// Gets the actual color of the line for the second data set.
         /// </summary>
         /// <value>The actual color.</value>
-        public OxyColor ActualColor2
+        public virtual OxyColor ActualColor2
```
objorke: I find it a bit strange that only the ActualColor2 property is virtual (why isn't ActualColor and ActualFill virtual...) But I see it is useful for the TwoColorAreaSeries.

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -238,11 +243,27 @@ public override void Render(IRenderContext rc)
             var clippingRect = this.GetClippingRect();
             rc.SetClip(clippingRect);
 
-            var chunksOfPoints = new List<List<ScreenPoint>>();
-            var chunksOfPoints2 = new List<List<ScreenPoint>>();
+            var areaContext = new AreaRenderContext
```
objorke: this is a new pattern, but I think this is something that can be used in more places. Maybe more of the rendering code could use composition over inheritance?

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -141,6 +141,11 @@ protected List<DataPoint> ActualPoints2
         protected int WindowStartIndex2 { get; set; }
 
         /// <summary>
+        /// Gets a value indicating whether Points2 collection was defined by user.
+        /// </summary>
+        protected bool IsPoints2Defined { get; private set; }
```
objorke: OK

Source/OxyPlot/Series/TwoColorAreaSeries.cs
```diff
@@ -15,58 +15,103 @@ namespace OxyPlot.Series
     /// <summary>
     /// Represents a two-color area series.
     /// </summary>
-    public class TwoColorAreaSeries : TwoColorLineSeries
+    public class TwoColorAreaSeries : AreaSeries
```
objorke: Good

Issue #118, opened 2014-08-14
Hierarchy or filter in example browser


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10147](http://oxyplot.codeplex.com/workitem/10147)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-03-01 14:04
The number of examples are growing, and there will probably be more... 

Currently there is only one level in the hierarchy and the list of folders is long.

I suggest to reduce the number of top level folders:

<ul>
<li>Showcases - a few examples to show some of the functionality of the library </li><li>Annotations - sub folders for each annotation type </li><li>Axes - sub folders for each axis type </li><li>Series - sub folders for each series type </li><li>Input events - examples of event handling </li><li>PlotController - examples using customized plot controllers </li><li>Discussion - examples from discussions </li><li>Open issues - examples related to open issues </li><li>Solved issues - examples related to closed issues

</li></ul>

An alternative could be to add filtering/tag/search functionality, then we could probably get rid of the hierarchy?

<pre><code>[x] Show case
[x] Annotations
[ ] Axes
[ ] Series
[ ] Discussion
[ ] Open issue
[ ] Solved issue
Search for [_______] (search tags, title, subtitle)</code></pre>

objorke:
I like the filtering shown in #895! It would be great if we could get a common view model for all the example browsers, and common functionality (filtering, grouping etc.) in all of them...

PR #895, opened 2016-05-22
Implemented a search box with some basic filtering (issue #118).

Isolocis:
Addresses #118. Please check performance of solution on Windows Phone and other computers, I have a feeling this solution might be a bit sluggish. Furthermore, this solution does not add a hierarchy, it's probably not impossible but please let me know if anyone has suggestions.

### Checklist

- Added filter to example browser.

@oxyplot/admins


objorke:
I like this! But it is only for Windows 8.1 - so issue #118 is not complete yet.

PR #913, opened 2016-06-02
Use SetClip in DrawClippedLine

### LINKED ISSUE #904: RenderingExtensions.DrawClippedLine need to call RenderContext.SetClip?
### RenderingExtensions.DrawClippedLine optimization?

Why RenderingExtensions.DrawClippedLine method don't call RenderContext.SetClip and only if it returns false clips polyline manually? This done in all DrawClippedXXX methods, but not in DrawClippedLine. I can make trivial pull request for this.

justslon:
Fixes #904.

### DIFF
--- a/Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs
+++ b/Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs
@@ -93,6 +93,25 @@ namespace OxyPlot
                 return;
             }
 
+            if (rc.SetClip(clippingRectangle))
+            {
+                rc.DrawLine(points, stroke, strokeThickness, dashArray, lineJoin, aliased);
+                rc.ResetClip();
+
+                if (outputBuffer != null)
+                {
+                    outputBuffer.Clear();
+                    outputBuffer.AddRange(points);
+                }
+
+                if (pointsRendered != null)
+                {
+                    pointsRendered(outputBuffer);
+                }
+
+                return;
+            }
+
             if (outputBuffer != null)
             {
                 outputBuffer.Clear();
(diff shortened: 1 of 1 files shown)


Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs
```diff
@@ -93,6 +93,25 @@ M3    M3
                 return;
             }
 
+            if (rc.SetClip(clippingRectangle))
+            {
+                rc.DrawLine(points, stroke, strokeThickness, dashArray, lineJoin, aliased);
+                rc.ResetClip();
```
objorke: shouldn't it pop the clip instead?

justslon: After small research I think you are right. It's better to call SetClip in, for example, LineSeries render. It seems clipped lines coords mostly used in annotation classes for label rendering. I think this PR can broke them. I'll try search for DrawClippedLine calls where last 2 arguments (output buffer and callback) not used and surround them with SetClip/ResetClip. I try to open new PR in a couple of days.

objorke: I change the SetClip method in the Wpf.CanvasRenderContext to:

```csharp
        public bool SetClip(OxyRect clippingRect)
        {
            if (this.clip != null)
            {
                throw new InvalidOperationException("SetClip cannot be nested.");
            }

            this.clip = this.ToRect(clippingRect);
            return true;
        }
```

this causes a large number of tests to break.

objorke:
I could not see anything wrong in the tests/example library png export... I think we can merge this to make the code consistent, then solve #982

PR #950, opened 2016-08-09
Fixed issue #925, minor code style changes

### LINKED ISSUE #925: WPF Application/Visual Studio freezes when LogarithmicAxis is reversed
### Steps to reproduce

In XAML (causes Visual studio 2013 to freeze):
1. Create Plot in XAML
2. Add a logarithmic axis 
3. Set the StartPosition of the Axis to 1, and the EndPosition to 0

`<oxy:Plot Name="plotView">`
`        <oxy:Plot.Axes>`
`              <oxy:LogarithmicAxis Position="Bottom" StartPosition="1" EndPosition="0" />`
`        </oxy:Plot.Axes>`
`</oxy:Plot>`

Via PlotModel (causes application to freeze):
1. Create a PlotModel and PlotView
2. Add a LogarithmicAxis with a StartPosition of 1 and end Position of 0
3. Set the plotModel as the PlotView`s model

`PlotModel plotModel = new PlotModel();`
`LogarithmicAxis axis = new LogarithmicAxis();`
`axis.StartPosition = 1;`
`axis.EndPosition = 0;`
`plotModel.Axes.Add(axis);`
`plotView.Model = plotModel;`

Platform:
.NET version: 4.5.2

### Expected behaviour

The chart should be displayed with the axis inverted (the order of the numbers should be reversed, e.g. counting down rather than up)

### Actual behaviour

If set in XAML this causes Visual Studio to freeze, if set in code the the application will freeze.
In the C# code provided above, the window will never actually be shown. If the axis is assigned after the window has been shown the application UI will become totally unresponsive.

### Cause

The issue seems to be an infinite loop in the LogDecadeTicks method of the OxyPlot.Axes.LogarithmicAxis class. I found changing lines 88 of the GetTickValues method to the following fixed the issue, though I do not know if this is getting to the route cause of the problem:
double axisBandwidth = Math.Abs(this.IsVertical() ? this.ScreenMax.Y - this.ScreenMin.Y : this.ScreenMax.X - this.ScreenMin.X);

Currently the code does not contain the call to Math.Absolute which results in axisBandwidth being negative. This in turn results in ticksPerDecade and decadesPerMajorTick being negative, which means the exit condition of LogDecadeTicks is never true. While the fix works, I'm not sure if the issue is actually that the ScreenMin values can be smaller than the ScreenMax values, or if this is expected in an inverted axis.

Isolocis:
Fixes #925 and takes care of some StyleCop warnings in the LogarithmicAxis class.

### Checklist

- [x] I have cleaned up the commit history (use rebase and squash

@oxyplot/admins

…h.Abs. Solved some StyleCop warnings in the LogarithmicAxis class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -177,6 +177,7 @@ All notable changes to this project will be documented in this file.
 - Exception when legend is outside and plot area is small (#880)
 - Axis alignment with MinimumRange (#794)
 - Fixed strange number formatting when using LogarithmicAxis with very large or very small Series (#589)
+- Fixed LogarithmicAxis to no longer freeze when the axis is reversed (#925)
 
 ## [2014.1.546] - 2014-10-22
 ### Added
--- a/Source/Examples/ExampleLibrary/Issues/Issues.cs
+++ b/Source/Examples/ExampleLibrary/Issues/Issues.cs
@@ -1620,6 +1620,7 @@ namespace ExampleLibrary
             return model;
         }
 
+        /// <summary>
         /// Creates a demo PlotModel with the data from the issue.
         /// </summary>
         /// <returns>The created PlotModel</returns>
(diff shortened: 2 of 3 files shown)


objorke:
thanks for cleaning up a lot of code! where is the actual line that fixes the issue?

Isolocis:
I added calls to Math.Abs on lines 65 and 66 of the LogarithmicAxis class, as was pointed out by @tecouchman.

objorke:
Thanks! 
But something must be wrong if the actual maximum value < the actual minimum value...
Will merge this anyway.

PR #958, opened 2016-08-17
Tooltip support for annotations and series data refresh fix

mrtncls:
### Checklist

- Adds support for WPF tooltips on Annotations
- Fixes refresh series when not visible (WPF)

@oxyplot/admins


Source/OxyPlot.Wpf/Annotations/Annotation.cs
```diff
@@ -113,6 +113,7 @@ public virtual void SynchronizeProperties()
             a.Layer = this.Layer;
             a.XAxisKey = this.XAxisKey;
             a.YAxisKey = this.YAxisKey;
+            a.ToolTip = this.ToolTip as string;
```
objorke: OK. Thanks!

Source/OxyPlot.Wpf/Series/Series.cs
```diff
@@ -64,7 +64,7 @@ public abstract class Series : ItemsControl
         /// </summary>
         static Series()
         {
-            VisibilityProperty.OverrideMetadata(typeof(Series), new PropertyMetadata(Visibility.Visible, AppearanceChanged));
+            VisibilityProperty.OverrideMetadata(typeof(Series), new PropertyMetadata(Visibility.Visible, DataChanged));
```
objorke: It would be great to have an example showing this issue!

objorke:
Please update the change log.

Source/OxyPlot.Wpf/Series/Series.cs
mrtncls: Small example added. See Examples/WPF/LiveDemo

Steps to reproduces:
1. Enable both checkboxes (show and generate) for a few seconds. You should see data plotting.
2. Uncheck both
3. Toggle 'show' and verfiy data is still there
4. Check 'generate' and wait a few seconds before unchecking
5. Toggle 'show' to check if data is correct.

Prior the fix, the old data would still be shown.

Hope all it's clear.

objorke:
I created #969 and #970

objorke:
See #971. I added examples and moved the example commits before the fixes. That makes it easier to check that the bug is actually fixed.

objorke:
I have merged #971.
Thanks for contributing!!

PR #961, opened 2016-08-19
Patch for resize bug with LogarithmicAxis (#957)

### LINKED ISSUE #957: resize with LogarithmicAxis
A bug freezes the application when a logarithmic axis gets too small.
Steps to reproduce:
- run example code
- resize the form window with mouse and make it as small as possible
--> the application hangs and the process has to be killed
```cs
using OxyPlot;
using OxyPlot.Axes;
using OxyPlot.WindowsForms;
using System;
using System.Windows.Forms;
static class Program
{
	[STAThread]
	static void Main()
	{
		var m = new PlotModel();
		m.Axes.Add(new LogarithmicAxis());
		var v = new PlotView();
		v.Model = m;
		v.Dock = DockStyle.Fill;
		var f = new Form();
		f.Controls.Add(v);
		Application.Run(f);
	}
}
```
tested with:
Windows 7, 64bit
.NET version: 4.5
OxyPlot 1.0.0-unstable2180

udoliess:
Fixes #957
Method LogarithmicAxis.LogDecadeTicks had to be fixed. If step<=0 the function could run endless. Even a very small step value could lead to unchanged loop variable exponent. Now the progress of iteration is checked.
In order to get only reasonable ticks the parameter step is checked to be >0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -178,6 +178,7 @@ All notable changes to this project will be documented in this file.
 - Axis alignment with MinimumRange (#794)
 - Fixed strange number formatting when using LogarithmicAxis with very large or very small Series (#589)
 - Fixed LogarithmicAxis to no longer freeze when the axis is reversed (#925)
+- Prevent endless loop in LogarithmicAxis (#957)
 
 ## [2014.1.546] - 2014-10-22
 ### Added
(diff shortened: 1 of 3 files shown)


objorke:
I included your commits in #968. Added an example and rebased.

objorke:
Thanks for contributing!!

PR #976, opened 2016-08-26
Improve text rendering in OxyPlot.GtkSharp by using Pango rather than…

### LINKED ISSUE #972: Various font issues with Gtk version
The current text rendering in Gtk has a number of failings. For examples, look in the "9 Rendering capabilities" section of the Example Browser, where "DrawText -Rotation" and "DrawText - MaxSize" do not render as expected. Another problem (and the one which got me looking into this) is that text annotations cannot cope with embedded line breaks; that is, they are restricted to a single line.
Issue #804 may be another example

These problems can be resolved by using Pango to assist in text rendering, rather than restricting the code to the Cairo level. I have revised code (almost) ready for submission that corrects these problems.

zur003:
… Cairo primitives.

Fixes #972 .

### Checklist

- [X ] I have included examples or tests
- [X ] I am listed in the CONTRIBUTORS file

Use Cairo to render text in OxyPlot.GtkSharp.

Earlier attempt failed on Gtk3 builds. Now trying to work around git issues.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,6 +91,7 @@ All notable changes to this project will be documented in this file.
 - Push packages to myget.org (#847)
 - Improve tracker style (Windows Forms) (#106)
 - Change the default format string to `null` for TimeSpanAxis and DateTimeAxis (#951)
+- Font rendering in OxyPlot.GtkSharp improved by using Pango (#972)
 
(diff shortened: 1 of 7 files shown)


objorke:
Looks good to me! Thanks for contributing!!

PR #978, opened 2016-08-29
Implemented display of tracker information in OxyPlot.GtkSharp.

### LINKED ISSUE #973: OxyPlot.GtkSharp should include a tracker implementation
The "tracker" normally seen when clicking on a series has not been implemented in the GtkSharp version of OxyPlot. It should be. Ideally this should be done with a TrackerControl like that now used in OxyPlot.WPF. It would be good if the WindowsForms implementation were also improved (e.g., so the label being displayed does not vanish off the edges of the plot when approaching the ends of the series).

zur003:
Fixes #973 .

### Checklist

- [x] I have included examples or tests (tests in ExampleBrowser suffice)

These changes implement a tracker in OxyPlot.GtkSharp, based largely on the implementation in OxyPlot.WindowsForms. It uses a Gtk Label to display the tracker information. This necessitated deriving PlotView from a Gtk Layout, rather than a Gtk DrawingArea, so that child widgets could be supported.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ All notable changes to this project will be documented in this file.
 - Support for PieSeries in OxyPlot.Wpf (#878)
 - Filter in example browser (#118)
 - Support for tooltips on WPF annotations
+- Support for tracker in OxyPlot.GtkSharp
 
 ### Changed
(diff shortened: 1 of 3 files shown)


objorke:
Great!

Issue #494, opened 2015-05-19
Add more interpolation algorithms for LineSeries/AreaSeries smoothing


xeo-it:
The Chart below is a simple AreaSeries Chart of the following values:  
0,4,100,100,100,100,100,100,100,0,0,0,0,0,0  with minimum = -10 and maximum = 110

![Chart](https://cloud.githubusercontent.com/assets/7143502/7695329/33383894-fe42-11e4-9e8b-9b362ba6d075.png)

I would expect the line to never go below 0 and above 100 but as you can see, it looks like few points have negative value and others are above 100.  
Below in red is how I would expect it to be:

![expected](https://cloud.githubusercontent.com/assets/7143502/7695381/01a0fc7a-fe43-11e4-82a3-53a08b725744.png)

objorke:
I think this is the correct behavior of the current spline interpolation. Should other interpolators be supported?

KaliCZ:
Yeah, we're dealing with the same issue right now, that the data actually go out of view in certain places. And the customer would really like it not to be that way. We may have to use a different tool to use if this won't get fixed soon.

objorke:
It would be great to see more interpolation algorithms implemented! Currently we are using a cardinal (canonical) spline applied to the points in screen space:
http://en.wikipedia.org/wiki/Cubic_Hermite_spline#Cardinal_spline
https://msdn.microsoft.com/en-us/library/4cf6we5y.aspx
http://www.charlespetzold.com/blog/2009/01/Canonical-Splines-in-WPF-and-Silverlight.html

Other splines:
https://en.wikipedia.org/wiki/B-spline
http://en.wikipedia.org/wiki/Cubic_Hermite_spline
http://en.wikipedia.org/wiki/Centripetal_Catmull%E2%80%93Rom_spline

PR #999, opened 2016-09-24
Fix small typo in Properties, additional interpolation algorithm

gro-ove:


gro-ove:
#494

How it looks in action:
image

Hmm, sorry, but why have checks failed? I guess I shouldn’t change WPF library at the same time as the main one?

objorke:
Very good! But we need some examples! It would be great to have the 5 plots in the ExampleLibrary.
I also think we should refactor the `InterpolationAlgorithm` to a more extensible data structure at some point. e.g. create an `IInterpolationAlgorithm` interface and use that for the `InterpolationAlgorithm` property.
Will you add the examples before I merge? And what about refactoring the enum?

gro-ove:
@objorke yes, of course! But just to be sure (sorry, I don’t have much experience):

- Replace `InterpolationAlgorithm` with `IInterpolationAlgorithm`;

- Transform `CanonicalSplineHelper` from an internal static class to `public class CanonicalSpline : IInterpolationAlgorithm`;

- Add something like `BaseCatmulRomSpline : IInterpolationAlgorithm` with `CatmullRom`, `UniformCatmullRom` and `ChordalCatmullRom` as children classes;

So, it will be used like `lineSeries.InterpolationAlgorithm = new CatmullRom()`, is it right?

objorke:
yes, this is exactly what I meant.

To make it easy to use we could also create a static class
``` cs
public static class InterpolationAlgorithms {
    public static IInterpolationAlgorithm CatmullRom { get; } = new CatmullRom();
}
```

then we could also write `lineSeries.InterpolationAlgorithm = InterpolationAlgorithms.CatmullRom;`

With this change, I think the `Smooth` property should be marked obsolete
https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Series/LineSeries.cs#L209
something like:
```cs
[Obsolete("Use the InterpolationAlgorithm property instead")]
public bool Smooth {
    get { return this.InterpolationAlgorithm is CanonicalSpline; }
    set { this.InterpolationAlgorithm = value ? InterpolationAlgorithms.CanonicalSpline : null; }
}
```

objorke:
this looks good! thanks!!

PR #1004, opened 2016-09-30
Add Axis Clipping Support for RectangleAnnotation

### LINKED ISSUE #977: RectangleAnnotation: Axis Clipping not configurable
While the LineAnnotation has bool properties ClipByXAxis, ClipByYAxis, the RectangleAnnotation does not have them. Instead, a RectangleAnnotation is always clipped to the used x/y axis.

I guess, this is ok for most scenarios, but it gets in the way when using multiple axes and the annotation should span multiple axes.

### Steps to reproduce
- one global horizontal (bottom) axis for everything
- two vertical (left) axes:
  - CategoryAxis from 0 to 0.45 ["Main Points", "Elements", "Images"]
  - LinearAxis from 0.55 to 1.0 (where the colored scatter points show up)
- A vertical LineAnnotation (cyan) that spans the full vertical range (possible via ClipByYAxis=false)
- A RectangleAnnotation (half transparent skyblue) that does **not** span the full vertical range.
  - Only MinimumX and MaximumX properties are set
  - RectangleAnnotation has no properties related to clipping

Platform: WPF (Windows 10)
.NET version: 4.6 (64bit)

### Expected behaviour

The RectangleAnnotation should support clipping to certain axes via bool properties and span the full graph range if that property is false.

### Actual behaviour

image

The vertical LineAnnotation spans the full y range, the RectangleAnnotation does not.

nzain:
RectangleAnnotation now has two bool properties ClipByXAxis, ClipByYAxis
(analogous to PathAnnotation) so that it is not necessarily clipped to a
single axis range, which fixes #977 . Added demo to example library.

### DIFF
--- a/Source/Examples/ExampleLibrary/Issues/Issues.cs
+++ b/Source/Examples/ExampleLibrary/Issues/Issues.cs
@@ -20,6 +20,124 @@ namespace ExampleLibrary
     [Examples("Z1 Issues")]
     public class Issues
     {
+        [Example("#977 RectangleAnnotation Axis Clipping not configurable")]
+        public static PlotModel RectangleAnnotationAxisClipping()
+        {
+            var model = new PlotModel
+            {
+                Title = "RectangleAnnotation Axis Clipping",
+                PlotAreaBorderThickness = new OxyThickness(0),
+                Axes =
+                {
+                    new LinearAxis
+                    {
+                        Position = AxisPosition.Bottom,
+                        AxislineStyle = LineStyle.Solid,
+                        EndPosition = 0.45
+                    },
+                    new LinearAxis
+                    {
+                        Position = AxisPosition.Bottom,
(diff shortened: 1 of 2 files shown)


nzain:
This PR is kind of a minimum-invasive solution to #977 (only change the RectangleAnnotation class) in order to align with the code style I found.

However, the Annotation class itself has properties for X/Y Axis and a protected method to get the clipping rectangle. I don't know about other annotations, but eventually the "do not clip by axis" stuff found in PathAnnotation (and with this PR in RectangleAnnotation, too) could move to the Annotation class itself. Unfortunately, each annotation still has to properly compute its maximum bounds to have any effect - and that part cannot be done in a generic way, but is specific to each annotation implementation.

objorke:
thanks for the contribution!

PR #1009, opened 2016-10-12
Automatically update Min and Max if no min and max setted

CedreLo:
Fixes # .
### Checklist
## 
## 
## 

@oxyplot/admins

Automatically UpdateActualMaxMin 

update Min and Max if no min and max setted


objorke:
Can you create an issue for this feature? I think undefined values should be `double.NaN`.
I don´t think `0` is an undefined value.
But I agree if minimum and maximum are set to the same value, it should be handled. Let us discuss how to handle it in the issue!

objorke:
Missing reference to issue.

CedreLo:
yes you are right, the same code with
double.NaN instead of zero  or if minimum and maximum are set to the same value
would be better

PR #1015, opened 2016-10-15
Fixed bug in LineAnnotation where calculated Y values could exceed set maximums

Hugoagogo:
Fixes bug with LineAnnotation generating line segment that exceeds set min/max y Values


objorke:
Could this bug be reproduced in any of the existing examples? Otherwise, please also include an example to show that the bug is fixed.

Hugoagogo:
Testing this again I can get the issue to occur with the release branch but not develop in the annotation demo, I think I just fixed something that was already fixed...

I will not be able to confirm this 100% for a few days.

objorke:
1. missing reference to an issue
2. missing example

PR #1016, opened 2016-10-17
Implemented zoom capability up to Milliseconds

NilesDavis:
In the DateTimeAxis the minum value intervals reached up to seconds, so I added milliseconds to "CalculateActualInterval". As well as ths ActualStringFormat in "UpdateIntervals".


NilesDavis:
How can I fix this failed check from appveyor?

Issue #34, opened 2014-08-14
Rotated axis titles


objorke:
Add a "TitleAngle" for the axes. ("Angle" is used for the tick labels)

objorke:
Is #22 a better solution? I think the rotated titles is only needed for vertical axes.

LukaBedi:
Hi Objorke, I really love your work and oxyplot is amazing :) i have recently come across the need to rotate the Y-axis label 90° and so i was wondering if issue #34 is solved ?

objorke:
No, this issue is open. The question is if it should be implemented as described in #22.

PR #1018, opened 2016-10-28
patch for oxyplot/xamarin issue #34, fixes plotview not rendering on …

clarkis117:
Fixes Oxyplot/Xamarin Issue 34 Plotview not rendering on Xamarin.Forms UWP 

https://github.com/oxyplot/oxyplot-xamarin/issues/34

The main cause of the issue was the function UpdateVisuals() in Oxyplot.Windows.PlotView, was not being called. This patch adds a one time condition that calls UpdateVisuals() thus forcing the plot to render.

### DIFF
--- a/Source/OxyPlot.Windows/PlotView.cs
+++ b/Source/OxyPlot.Windows/PlotView.cs
@@ -791,30 +791,44 @@ namespace OxyPlot.Windows
             e.Handled = this.ActualController.HandleMouseLeave(this, e.ToMouseEventArgs(this));
         }
 
-        /// <summary>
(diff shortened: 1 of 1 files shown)


objorke:
Thanks for providing a fix for this!!
I still want to check if the UWP PlotView is different from the other PlotViews. I don't think it should be necessary to add such custom code to the Xamarin Forms renderers.

clarkis117:
@objorke  I thinking one of the main problems is that the size parameter is only provided to the `protected override Size ArrangeOverride(Size finalSize)` method one, and in all my debugging it's only been 0 for both x and y thus causing `UpdateVisuals()` to not be called.

PR #1037, opened 2016-12-14
Ensure temporary file for PNG export is written in a suitable folder.

### LINKED ISSUE #1034: PNG export can fail in GtkSharp due to inappropriate path
### Steps to reproduce

1. Write code for the Gtk# version that calls PngExporter.Export(IPlotModel model, Stream stream)
2. Execute this code from a directory where the user does not have write access.
3. This will fail, with an exception along the lines of "Could not find file "/858e6ab1-65d4-4f7b-8988-12127b93b904.png"

Platform: OS X (El Capitan)
.NET version: 4.5

### Expected behaviour
The image should export without an error occurring.

### Actual behaviour
The code will fail, and an exception will be raised.

The problem clearly lies in OxyPlot.GtkSharp\PngExporter.cs, lines 103:105, as follows

                    // write to a temporary file
                    var tmp = Guid.NewGuid() + ".png";
                    bm.WriteToPng(tmp);

Note that the name of the temporary file does not include a full path. This file should be generated in a folder where the user is guaranteed to have write access.

zur003:
Fixes #1034 .

### Checklist

- When a temporary file is created for use in PNG export in OxyPlot.GtkSharp, it will now be created in a suitable folder.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 - Manipulation when using touch is not working in Windows (#1011)
+- Ensure a suitable folder is used when creating a temporary file for PNG export in Oxyplot.GtkSharp (#1034)
 
 ## [1.0.0] - 2016-09-11
 ### Added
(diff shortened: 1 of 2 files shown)


objorke:
thanks for fixing the issue!

PR #1055, opened 2017-01-19
Fix for #382: OxyMouseEvents not caught due to InvalidatePlot()

### LINKED ISSUE #382: OxyMouseEvents not caught due to InvalidatePlot()
I noticed that my MouseDown (attached to my Model or Series) event handlers appear to not fire when my plot is updating. Is there a workaround/solution for this? 

E.g) I have a PlotModel that I update fairly rapidly, ~30Hz. Certain MouseDown events (~ 1 out of 5) are not recieved by my handler(s), and it appears to be caused by the mouse events coinciding with occurrences of Model.InvalidatePlot(true/false).

HermanEldering:
Fixes #382.

### Checklist

### Remarks

I don't use git often, so I'm reading up on the rebase and squash. With regards to the tests this issue probably not easy to unit test. If I get around to it the best I could do is create an example showing the issue.

This fix is tested in my own use of OxyPlot, and it is somewhat similar to a fix mentioned in the referenced issue #382. I also tried some of the OxyPlot WPF examples and in general they appear to function properly (context menus/zoom/pan/trackers), however my change breaks the tooltips. 

For WPF tooltips the [mouse must be directly over](https://referencesource.microsoft.com/PresentationFramework/src/Framework/System/Windows/Controls/PopupControlService.cs.html#107) the control, now a grid is in between. I did not find a way to resolve this using the native WPF tooltips.

@oxyplot/admins


objorke:
Thanks! This looks OK to me. The other XAML platforms should probably be updated too?
I'll squash when completing this PR!

HermanEldering:
Thanks. I haven't looked into the other platforms yet. I think removing all controls and generation new controls is what causes the issue, so it not only depends on OxyPlot but also a bit on the implementation of the XAML platform itself.

holgerschl:
I had the follwing issue when using OxyPLot 1.0.0.0:
The mousdown event does not fire when I click on a line series. It works fine outside of the line series. 
With OxyPlot 2.0.0.0 unstable it went away - was this what you fixed?
In that case a test is easy: Plot a line series and click on it - the mousedown event should fire.

HermanEldering:
@holgerschl I am not sure what exactly your question is. I think in my case the problem was mostly with clicking a line annotation or the plot area in general, but the person that created the associated issue #382 had the problem clicking a series.

In this case some events will be lost, and the problem will be worse when the plot is invalidated more frequently. If you have your problem even when you're not invalidating/updating the plot or if you don't see any events at all for the series it is probably another problem.

The basic problem in this ticket is that the UI elements are being updated while events are generated. So what I believe that happened is that you would click on an UI element, but before the OxyPlot event would fire that UI element might be destroyed and replaced with a new element. Resulting in events getting lost.

The solution that I have implemented in this pull request is to create a transparant grid on top of all UI elements so that all mouse events are caught by the grid which is not destroyed during the invalidate.

holgerschl:
Ok - that naturally fixed my problem as well. But the underlying problem was different - on a series the MouseDown event (either on the PlotController or on the series  itself) never fired.

PR #1080, opened 2017-04-17
Fix for #937, When Color property of LineSeries is set Markers are not…

### LINKED ISSUE #937: When Color Property of LineSeries is set Markers are not shown
### Steps to reproduce

1.Create a LineSeries as below.
2. var series2 = new LineSeries { Title = "Series 2", MarkerType = MarkerType.Square, C**olor= OxyColors.Black**, Smooth = false, StrokeThickness = 1 };
3. Add points to series and add series to PlotModel 

Platform: Windows/WPF
.NET version: 4.0
### Expected behaviour

Line with Markers shall be shown.
### Actual behaviour

Markers are not shown when Color property of series is set.

Image is attached
image

shnaramn:
… shown. In this fix, the marker color is set to the line color or default color if it was not set explicitly. Tested using a sample plot (included it in the fix)

Fixes #937 When Color property of LineSeries is set Markers are not shown

### Checklist

- [Y] I have included examples or tests
- [Y] I have updated the change log
- [Y] I am listed in the CONTRIBUTORS file
- [Y] I have cleaned up the commit history (use rebase and squash)

- Fixing the LineSeries. If the marker color is not explicitly set, use the color of line if set or use a default color.
- Updating the example line series app to include series which have Marker and Line color set/unset.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ All notable changes to this project will be documented in this file.
 - Ensure a suitable folder is used when creating a temporary file for PNG export in Oxyplot.GtkSharp (#1034)
 - RangeColorAxis is not rendered correctly if the axis is reversed (#1035)
 - OxyMouseEvents not caught due to InvalidatePlot() in WPF (#382)
+- When Color Property of LineSeries is set Markers are not shown (#937)
 
 ## [1.0.0] - 2016-09-11
 ### Added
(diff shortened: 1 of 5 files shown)


objorke:
the change looks ok. 
But I would like to have the changes on the "SimpleDemo" application reverted, please add the examples of the different marker color options in the ExampleLibrary instead! Can you commit this change to the PR?

shnaramn:
I've updated the fix per suggestions. Let me know if anything is missing.

objorke:
thanks!!

PR #1083, opened 2017-04-25
Added check to LineAnnotation.GetScreenPoints()

### LINKED ISSUE #1029: Possible Infinite Loop in LineAnnotation.GetPoints() when ActualMinimum=ActualMaximum
### Steps to reproduce

1. Create a y-axis has StartPosition=EndPosition=0, but do not use LinearAxis (derived classes hit this issue, too)
2. Create a vertical LineAnnotation

Platform: any
.NET version: any

### Expected behaviour

No exception is thrown. Infinite loops have to be avoided at all times either by checking the increment or using a maximum iteration count.

### Actual behaviour

When either the X or Y Axis are _not_ instances of LinearAxis (thus, bool curvedLine is true), the line annotation produces not two but many points in LineAnnotation.GetPoints(). It produces an infinite "while (true)" loop because the increment is zero because ActualMaximum=ActualMinimum.

The result is a OutOfMemoryException:
image

### Suggested Fix Strategy
When ActualMaximum=ActualMinimum the GetPoints() method can quickly return an empty result.

bryanfree66:
to check if ActualMaximumX==ActualMinimumX for non-curved lines.

Fixes #1029.

### Checklist

-
-
-

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -133,6 +133,7 @@ All notable changes to this project will be documented in this file.
 - ListFiller (#705)
 
 ### Fixed
+- Added check to LineAnnotation.GetScreenPoints to check if ActualMaximumX==ActualMinimumX for non-curved lines. (#1029)
 - Incorrect placment of axis title of axes with AxisDistance (#1065)
 - SharpDX control not being rendered when loaded
 - SharpDX out of viewport scrolling.
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -17,6 +17,7 @@ Benoit Blanchon <>
 br
 brantheman
 Brannon King
+Bryan Freeman
 Brian Lim <brian.lim.ca@gmail.com>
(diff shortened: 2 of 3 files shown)


objorke:
I need some more time to review this change properly.
Thanks for the contribution!!

objorke:
this PR is on my todo list!

objorke:
it is still on my todo list, I just want to get some other issues done first!

objorke:
1. Missing example/unit test that reproduces the issue
2. I don't think the changes in this PR solves the issue. The inifinite loop may still happen in 
https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Annotations/LineAnnotation.cs#L121
or 
https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Annotations/LineAnnotation.cs#L138

objorke:
I created #1112 - the bug should be fixed there.

Issue #1145, opened 2017-09-19
Continuous Histograms


VisualMelon:
### Feature Proposal: Continuous Histogram Series

_Disclaimer: I'm not a statistician; my terminology may be somewhat wrong_

#### Summary of Concern

People use Histograms all the time, though OxyPlot doesn't seem to have fantastic support for them (I usually end up trying to coerce `ColumnSeries`). I would suggest adding a HistogramSeries so that people can plot continuous-axis Histograms. Oddly, I can't find a single result when searching for "Histogram" in the Issues.

Two features of histograms that are currently not served well by the library:

 - They are inherently continuous (i.e. not categorical)

 - They can have different width 'bins'

#### Proof of Concept

I have (very quickly, for other purposes) thrown together a trivial `ContinuousHistogramSeries` implemention which can be found [in a repository](https://github.com/VisualMelon/TrivialHistogram) (see the ReadMe for a picture). It is by no means 'pull request ready', but I'm prepared (time permitting) to put the effort into shaping it up/re-writing it properly if this would be a desirable addition. It is hurriedly built from `RectangleSeries` and `BarSeriesBase`, and practically untested. It sits on a pair of continuous X/Y Axes, and supports variable width bins.

#### API Misc

It would expect the user to pass the series are set of Bin descriptions (in the example, they are `ContinuousHistogramItem`s), which indicate the range (x start -> x end) and area: bar height (y) is then `area / width`. Methods would be provided to collect(responsively or otherwise) continuous samples into bins which can be fed to the Series. (Two simple `Collect` methods can be found used in the example repo).

I would expect colors to work like a `BarSeries`, but `ColorAxis` support might also make sense.

#### Misc Misc

I haven't given much thought to the Stacked Histograms, but it is certainly a possibility (can't have side-by-side entries like in a Column plot, owing to variable-width bins).

Horizontal Histogram support would also be simple and useful.

I'm happy to throw together some more examples if they would be of interest.

objorke:
I like the way matlab and matplotlib defines the histograms
- https://se.mathworks.com/help/matlab/ref/histogram.html
- https://matplotlib.org/1.2.1/examples/pylab_examples/histogram_demo.html

VisualMelon:
Indeed, I would also like to have such automatic binning methods, but I would see those as a layer on top of the series, and such a translator could be supplied as an ItemSource.

[...]

PR #1154, opened 2017-10-11
Add ContinuousHistogramSeries and examples

VisualMelon:
Addresses #1145 

### Checklist

- Addition of a ContinousHistogramSeries and associated ContinousHistogramItem and ContinousHistogramHelpers
- Addition of 3 examples of the above series

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Series/ContinuousHistogramSeriesExamples.cs
+++ b/Source/Examples/ExampleLibrary/Series/ContinuousHistogramSeriesExamples.cs
@@ -1,0 +1,104 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="ContinuousHistogramSeriesExamples.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Creates example histograms
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace ExampleLibrary
+{
+    using System;
+
+    using OxyPlot;
+    using OxyPlot.Axes;
+    using OxyPlot.Series;
+    using System.Collections.Generic;
+
+    [Examples("ContinuousHistogramSeries"), Tags("Series")]
+    public class ContinuousHistogramSeriesExamples
+    {
+        [Example("Exponential Distribution")]
+        public static PlotModel ExponentialDistribution()
+        {
+            return CreateExponentialDistribution();
+        }
+
+        [Example("Custom Bins")]
+        public static PlotModel CustomBins()
+        {
+            return CreateExponentialDistributionCustomBins();
+        }
+
+        [Example("Disconnected Bins")]
+        public static PlotModel DisconnectedBins()
+        {
+            return CreateDisconnectedBins();
+        }
+
+        public static PlotModel CreateExponentialDistribution(double mean = 1, int n = 10000)
+        {
+            var model = new PlotModel { Title = "Exponential Distribution", Subtitle = "Uniformly distributed bins (" + n + " samples)" };
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Left, Title = "Frequency" });
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom, Title = "x" });
+
+            Random rnd = new Random();
+
+            ContinuousHistogramSeries chs = new ContinuousHistogramSeries();
+            chs.Items.AddRange(ContinuousHistogramHelpers.Collect(SampleExps(rnd, 1.0, n), 0, 5, 15, true));
+            chs.StrokeThickness = 1;
+            model.Series.Add(chs);
+
+            return model;
+        }
+
+        public static PlotModel CreateExponentialDistributionCustomBins(double mean = 1, int n = 50000)
+        {
+            var model = new PlotModel { Title = "Exponential Distribution", Subtitle = "Custom bins (" + n + " samples)" };
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Left, Title = "Frequency" });
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom, Title = "x" });
+
+            Random rnd = new Random();
+
+            ContinuousHistogramSeries chs = new ContinuousHistogramSeries();
+
+            chs.Items.AddRange(ContinuousHistogramHelpers.Collect(SampleExps(rnd, 1.0, n), new double[] { 0, 0.1, 0.2, 0.3, 0.4, 0.5, 0.75, 1.0, 2.0, 3.0, 4.0, 5.0 }, true));
+            chs.StrokeThickness = 1;
+            chs.FillColor = OxyColors.Purple;
+            model.Series.Add(chs);
+
+            return model;
+        }
+
+        public static PlotModel CreateDisconnectedBins()
+        {
+            var model = new PlotModel { Title = "Disconnected Bins" };
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Left, Title = "Representation" });
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom, Title = "x" });
+
+            Random rnd = new Random();
+
+            ContinuousHistogramSeries chs = new ContinuousHistogramSeries();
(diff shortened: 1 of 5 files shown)


VisualMelon:
I was uncertain how the helper classes might present themselves; for the time being they are in a `Helpers` static class, but I suspect that isn't ideal, and shall gladly address this among any other feedback.

Apologies for any stupid errors, this is my first significant contribution.

Source/OxyPlot/Series/ContinuousHistogramHelpers.cs
```diff
@@ -0,0 +1,117 @@
...
+namespace OxyPlot.Series
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    
+    /// <summary>
+    /// Provides methods to collect data samples into bins for use with a <see cref="ContinuousHistogramSeries" />.
+    /// </summary>
+    public static class ContinuousHistogramHelpers
```
objorke: Is this a builder? https://en.wikipedia.org/wiki/Builder_pattern

objorke: Should it be a non-static class? I think I would prefer that, even if it is a single call... Not sure...

Source/OxyPlot/Series/ContinuousHistogramItem.cs
```diff
@@ -0,0 +1,120 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="ContinuousHistogramItem.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Represents an item in a RectangleSeries.
```
objorke: wrong comment

Source/Examples/ExampleLibrary/Series/ContinuousHistogramSeriesExamples.cs
```diff
@@ -0,0 +1,103 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="ContinuousHistogramSeriesExamples.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Creates a simple example heat map from a 2×3 matrix.
```
objorke: wrong comment

Source/OxyPlot/Series/ContinuousHistogramItem.cs
```diff
@@ -0,0 +1,120 @@
...
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Series
+{
+    /// <summary>
+    /// Represents an item in a <see cref="ContinuousHistogramSeries" />, a bin (range) and its area.
+    /// </summary>
+    public class ContinuousHistogramItem
```
objorke: do we need these items? Couldn't the continuous histogram series be defined simple by start, end and an array of values?

Source/OxyPlot/Series/ContinuousHistogramItem.cs
```diff
@@ -0,0 +1,120 @@
...
+            this.Area = area;
+
+            this.Color = OxyColors.Automatic;
+        }
+
+        /// <summary>
+        /// Gets or sets the color of the item.
+        /// </summary>
+        /// <remarks>If the color is not specified (default), the color of the series will be used.</remarks>
+        public OxyColor Color { get; set; }
```
objorke: do we really need this?

Source/OxyPlot/Series/ContinuousHistogramSeries.cs
```diff
@@ -0,0 +1,560 @@
...
+        /// Gets the actual fill color.
+        /// </summary>
+        /// <value>The actual color.</value>
+        public OxyColor ActualFillColor
+        {
+            get { return this.FillColor.GetActualColor(this.defaultFillColor); }
+        }
+        
+        /// <summary>
+        /// Gets or sets the color of the interior of the bars when the value is negative.
```
objorke: can histograms contain negative values?

Source/OxyPlot/Series/ContinuousHistogramHelpers.cs
VisualMelon: I didn't intend to make it a builder: it is just data processing really (data which can be fed directly to the series). I contemplated making it just part of the `ContinousHistogramItem` series, but I couldn't find a real precedent either way in the codebase.

Source/OxyPlot/Series/ContinuousHistogramItem.cs
VisualMelon: Probably not, I'll strip it out (forgot it was there, to be quite honest)

Source/OxyPlot/Series/ContinuousHistogramSeries.cs
VisualMelon: No, I don't think so, but I felt like providing a half-meaningful behaviour rather than crashing on 'invalid' data. While a 'real' histogram won't contain negatives, I don't see much harm in facilitating a 'dodgy' histogram. I'd be inclined to leave this in, but if I'm missing/underestimating a good reason to remove it I shall do so without complaint.

VisualMelon:
(Certainly up for some discussion; just hopefully doesn't require too much work on your part! I was waiting for some simulations to ran and just sort of got on with this, but I still don't have a very good idea of how the internals are meant to look). My git skills are atrocious, but I shall try to address the above feedback in short order.

Source/OxyPlot/Series/ContinuousHistogramItem.cs
VisualMelon: Something needs to decide how to allocate the bins: this way the user has complete freedom (and we can provide methods to make it easy for, e.g. start/end/samples as presented). Not to mention that the user may not actually have the sample data, only counts collected in bins, in which case anything else would present a miserable challenge.

VisualMelon:
(That last push has gone very wrong... consider it being fixed...)

VisualMelon:
Frustratingly, I have been unable to gleam any intelligence from the errors presented in the AppVeyor log:

```
[00:00:33] INFO [10/11/17 16:20:45:60] Applicable build agent found: 'AppVeyor'.
[00:00:33] Executing GenerateSetVersionMessage for 'AppVeyor'.
[00:00:34] ERROR [10/11/17 16:20:46:47] An unexpected error occurred:
[00:00:34] System.Net.WebException: The remote server returned an error: (500) Internal Server Error.
[00:00:34]    at System.Net.HttpWebRequest.GetResponse()
[00:00:34]    at GitVersion.AppVeyor.GenerateSetVersionMessage(VersionVariables variables)
```

As far as I can tell, the PR looks well formed (pulled my code back: compiles; no apparent opportunity for conflict; commits are in the right order, AppVeyor log suggests (I think) that it was able to merge successful).

Is there a way to rerun the tests (could this be an unfortunate chance network failure?), or am I missing something? (Not that I expect this to be merged in its current spate, but mysteriously failing tests doesn't inspire much confidence, especially when it passed the first time)

I shall consult with someone less useless at git when possible to check whether I have done something outrageous in my attempts to rewrite history.

VisualMelon:
Thank you kindly... I am sorry about this needless complexity: I'm sure it's my fault (if anyone's), but I really can't work out what I've done to cause it (I use git every day, but almost never as a team).

The latest version adds better ToCode support to the items, which is observable in the examples.

objorke:
image
I think rendering should be a bit better (this is from WPF) - need to make sure the pixel snapping gets correct. There should not be double line thickness between the bars. And absolutely no space. I know this is a difficult topic...

VisualMelon:
Apologies, by chance that doesn't present itself for me when I first open the example browser, but it appears to be fixed by switching to `DrawClippedRectangleAsPolygon` rather than `DrawClippedRectangle`, though I am not aware if there are significant performance implications of this.

(Ah, I see you noticed heatmap as well!)

objorke:
I updated to the latest gitversion, but it seems like the only solution is to create a new branch and a new PR...

VisualMelon:
That is fine by me, if it needs doing. Would you rather discuss the code some more before I do so, or should I get on with it? (Anticipating a number of changes before pulling)

objorke:
could also try adding more commits to the current branch, maybe that solves the issue

VisualMelon:
Still unhappy, will push another branch...

PR #1161, opened 2017-10-23
Fixed behavior of GetAxisOrDefault and added GetAxis as a replacement

### LINKED ISSUE #1137: GetAxisOrDefault throws InvaidOperationException when key is not found
### Steps to reproduce

1. Create PlotModel
2. Try to call `GetAxisOrDefault` with a key for an axis that doesn't exist
3. Boom

Platform: Xamarin.iOS
.NET version:

### Expected behaviour

With methods that have the OrDefault postfix, I'd expect them to return the default value in case of what I query is not present and not throw an exception.

For instance in a LINQ query where I have an enumerable where I use FirstOrDefault, in case first is not found, I just get the default value.

### Actual behaviour

InvalidOperationException is thrown telling me that the key is not present.

This means that the default value can never be obtained and the `OrDefault` postfix is very silly to have in this case.

tersers:
- GetAxisOrDefault will no longer throw an exception if the key does not exist.
- Added GetAxis to replicate the old functionality of GetAxisOrDefault
- Replaced instances of GetAxisOrDefault with GetAxis

Fixes #1137.

### Checklist

- A new function named 'GetAxis' has been introduced which throws when its key argument is not found. Calls to GetAxisOrDefault --primarily in functions named 'EnsureAxis-- were replaced with GetAxis so their behaviour remains the same.

@oxyplot/admins

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -99,6 +99,7 @@ Surfin Bird <illvdg13@gmail.com>
 Sven Dummis
 Taldoras <taldoras@googlemail.com>
 Tandy Carmichael <tcarmichael@frontier.com>
+Tasos Stamadianos <tasos.a.stam@gmail.com>
 Thorsten Claff <tclaff@gmail.com>
 thepretender
 tephyrnex
--- a/Source/Examples/ExampleLibrary/CustomSeries/FlagSeries.cs
+++ b/Source/Examples/ExampleLibrary/CustomSeries/FlagSeries.cs
@@ -185,7 +185,7 @@ namespace ExampleLibrary
         /// </summary>
         protected override void EnsureAxes()
         {
-            this.XAxis = this.PlotModel.GetAxisOrDefault(this.XAxisKey, this.PlotModel.DefaultXAxis);
+            this.XAxis = this.PlotModel.GetAxis(this.XAxisKey);
         }
 
         /// <summary>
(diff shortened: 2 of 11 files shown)


tersers:
I'm aware some of the tests are failing. It looks like my changes broke the behaviour when axis labels are `null`. Currently, `GetAxisOrDefault` works similarly to the `FirstOrDefault` extension of `IEnumerable`, where it will return the a default value if the key is not found or is `null`. This breaks any functions that expected `GetAxisOrDefault` to throw, so I replaced them with `GetAxis` to ensure this throwing behaviour. For consistency, `GetAxis` was written similarly to `IEnumerable.First`, and will therefore throw an exception if the key is `null`. My mistake was in assuming the internals of `Series` wouldn't allow for a `null` axis key.

There are two options going forward:

1. Wrap calls to `GetAxis` in `if`s in order to return a default value if the key is null.
2. Create another function, `PlotModel.GetAxisOrDefaultIfNull` (or some other nicer name), that wraps the code mentioned in item 1.

I favour the first option since it's more LINQ-esque. What do you think?

tersers:
The most recent commit addresses the above issue and fixes it via option 1.

objorke:
Excellent. I like that you follow the LINQ-esque pattern from `FirstOrDefault` :-)

Could we simplify this even more? I don´t think we need both `GetAxis` and `GetAxisOrDefault`. Could we have just a method `Axis GetAxis(string key, Axis defaultAxis)`. 
The big question is what should happen if the key is invalid - should it throw an exception or return the default axis. I added a comment on the issue. What´s your thoughts?

objorke:
Sorry for the delay. I will merge this now!

PR #1167, opened 2017-11-02
Make comparison match intended semantics and possibly fix unconfirmed bug

chaselal:
[An old commit](https://github.com/trainerroad/oxyplot/commit/8dca8dd0515aedc3278c03a97080e515962be4b1) in TrainerRoad's `oxyplot` fork claims that a comparison incorrectly returns `false` when it should return `true` on some old iOS devices, specifically iPhone 4 and possibly 5th gen iPod touch. I have not tested this. However, I investigated and saw that the comparison doesn't really match the intent of the code, so I changed the comparison to make the intent more obvious and (hopefully) fix the comparison issue (if there is one) on affected devices.

### Checklist

- Use a direct comparison to `default(double)` to check if an item was not found in an array. The existing code checks if a LINQ `FirstOrDefault` call's return value is `< double.Epsilon`. The only possible values of the call are `default(double)` and any of the items in the array. The smallest element in the array is `1.0`, so the `if` predicate should evaluate to `true` when the `FirstOrDefault` call doesn't find anything.

@oxyplot/admins

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -12,3 +12,4 @@
 Oystein Bjorke <oystein.bjorke@gmail.com>
 DNV GL AS
 LECO® Corporation
+TrainerRoad, LLC <lorentz@trainerroad.com>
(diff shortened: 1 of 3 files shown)


objorke:
Sorry for the slow response!
Agree this matches the original intent. Will merge this change.
This should be improved, I don't think "good intervals" are `1h, 2h, 4h, 8h, 16h, 32h`... but rather `1h, 2h, 3h, 4h, 6h, 12h, 24h, 48h, 24*7h`... :-)

PR #1185, opened 2018-01-21
Fix issue #1184

### LINKED ISSUE #1184: Pan/Zoom doesn't work on touch devices
GeertvanHorrik introduced a bug with commit:  https://github.com/oxyplot/oxyplot/commit/4edac8250db56a183c50002fae645a1b6f7291fa#diff-475cc427b3b23fc83502b42b11f06316 by introducing a new protected property called bool SetHandledForPanOrZoom but NOT setting the default value to true.  He overrides the value to false in TouchTrackerManipulator but forgot to set the base class TouchManipulator to true.

scastria:
Fixes #1184  .

### Checklist

-
-
-

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/PlotController/Manipulators/TouchManipulator.cs
+++ b/Source/OxyPlot/PlotController/Manipulators/TouchManipulator.cs
@@ -21,6 +21,7 @@ namespace OxyPlot
         public TouchManipulator(IPlotView plotView)
             : base(plotView)
         {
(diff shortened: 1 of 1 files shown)


scastria:
I would also like to remove the OnTouchDown binding for the tracker since you can't have 2 TouchDown bindings at the same time.  I think pan/zoom is more important than the tracker.

PR #1192, opened 2018-02-05
Mitigate unsightly gaps when using HeatMapRenderMethod.Rectangles

### LINKED ISSUE #1191: HeatMap drawn with rectangles shows white lines
### Steps to reproduce

1. Add a PlotView to a WPF view with a binding to a PlotModel in the view model:
    `<oxy:PlotView Model="{Binding MyPlotModel}"/>`

2. Instantiate a PlotModel as followed:

        private PlotModel myPlotModel;

        public MyViewModel()
        {
            myPlotModel = new PlotModel { Title = "My Plot Model" };
            myPlotModel.Axes.Add(new LinearColorAxis
            {
                Maximum = 100,
                Minimum = 0,
                Palette = OxyPalettes.Cool(100)
            });

            myPlotModel.Axes.Add(new LinearAxis()
            {
                Title = "Z",
                StartPosition = 1,
                EndPosition = 0,
                Position = AxisPosition.Left
            });
            myPlotModel.Axes.Add(new LinearAxis()
            {
                Title = "X",
                Position = AxisPosition.Bottom
            });
        }

I also have a bindable property named MyPlotModel.

3. Add data to my plot model with some method accepting `plotModel` and `heatmapData` as parameters:

            var heatMap = new HeatMapSeries
            {
                X0 = parentViewModel.MinX,
                X1 = parentViewModel.MaxX,
                Y0 = parentViewModel.MinZ,
                Y1 = parentViewModel.MaxZ,
                RenderMethod = HeatMapRenderMethod.Rectangles,
                Interpolate = false,
                Data = heatmapData
            };

           plotModel.Series.Add(heatMap);
           plotModel.InvalidatePlot(true);

Platform:
.NET version: 4.6.1
WPF application

### Expected behaviour

I want to see a plot without these strange white lines running through it.

### Actual behaviour

I see a plot with strange white lines running through it, as seen in the following example:

image

VisualMelon:
This has the effect of preventing unsightly gaps appearing with some rendering methods (e.g. WPF/exported PDF (PDFs being the original use-case for this Heatmap rendering method)).

Fixes #1191 

### Checklist

### Changes proposed in this pull request

 - Changes the method of rendering rectangles from `rc.DrawClippedRectangle` to `rc.DrawClippedRectangleAsPolygon`, with a stroke the same colour as the rectangle of width `1.0`. Evidence of the change can be observed when viewing the "Logarithmic X, discrete rectangles" example from the WPF Example Browser.

### Undesirable side effects

 - Has an observable (negative) performance impact when rendering Heatmaps with WPF.
 - Can create anti-aliasing artefacts on the edges of the Heatmap owing to overlapping stroke rendering (a somewhat liberal width of 1.0 was chosen, as gaps were observable with widths between 0.001 and 0.1; it is possible that a better understanding of the rendering routines could identify a better choice of width if these present a future concern)

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ All notable changes to this project will be documented in this file.
 - OverflowException when zoomed in on logarithmic axis (#1090)
 - ScatterSeries with DateTimeAxis/TimeSpanAxis (#1132)
 - Exporting TextAnnotation with TextColor having 255 alpha to SVG produces opaque text (#1160)
+- Change method used to render Rectanges in HeatmapSeries to prevent unsightly gaps when rendering (#1191)
 
 ## [1.0.0] - 2016-09-11
 ### Added
--- a/Source/OxyPlot/Series/HeatMapSeries.cs
+++ b/Source/OxyPlot/Series/HeatMapSeries.cs
@@ -324,7 +324,8 @@ namespace OxyPlot.Series
                         var pointb = this.Orientate(new ScreenPoint(s00.X + ((i + 1) * sdx), s00.Y + ((j + 1) * sdy))); // re-orientate
                         var rectrect = new OxyRect(pointa, pointb);
 
-                        rc.DrawClippedRectangle(clip, rectrect, rectcolor, OxyColors.Undefined, 0);
+                        // stroke is necessary to ensure there is no rendered gap between rectangles
+                        rc.DrawClippedRectangleAsPolygon(clip, rectrect, rectcolor, rectcolor, 1);
                     }
                 }
             }


objorke:
Thanks for a well documented contribution! I have some comments:
- the referenced issue is already closed, can we reopen it or create a new one?
- I don't think we should make a change that impacts the performance. I think it would be better to create a new render method (e.g. HeatMapRenderMethod.Polygons)
- I am a bit concerned about the anti-aliasing artefacts due to the stroke...

I would prefer to keep the current drawing method using rectangles, but improve the calculation of the rectangle boundaries (applying some kind of pixel snapping rounding?).

VisualMelon:
Thanks for the feedback; this came up a while back, and I couldn't remember if I'd agreed to actually implement it or not: I figured that was no harm in putting the PR together (since I had to make the change to see it) rather than trying to start another conversation. At the time it looked like a quick and easy fix (before I realise the strokes were 'necessary'), but it hasn't quite gone to plan. I am now also of the opinion that the performance costs are too high to take at all lightly.

The reference issue recognises the unsightly gaps; the 'solution' was to use an alternative rendering method (really just a work-around); however, that solution doesn't work on all platforms (e.g. PDF, which has always been my use-case (why I implemented the rectangles rendering mode in the first place), I just don't really care if my heat maps have lines on them day-to-day, so I've not bothered to look into it). I'm happy to create another issue if that makes sense, specific to PDFs perhaps.

Having since tried out some more of these heat-maps, it seems I was right to choose the word 'mitigate', because (at least for PDF) the lines are still manifested at extreme zoom levels. I agree that a pixel-snapping solution would be good, but I don't know if that makes sense for PDF.

For reference, here are a couple of test heatmap PDFs:
 - Using code from this PR: [heatmapNew.pdf](https://github.com/oxyplot/oxyplot/files/1710737/heatmapNew.pdf)
 - Using old/current code: [heatmapOld.pdf](https://github.com/oxyplot/oxyplot/files/1710738/heatmapOld.pdf)

I think I don't presently have a deep enough understanding of the OxyPlot -> PDF -> Screen processes to make especially useful comments, so perhaps I'll do some more research into this/look at how other systems produce beautiful PDFs (and hope the solution is platform independent).

objorke:
in WPF it is possible to use the device coordinate transforms:

.xaml:
```xml
        <Canvas x:Name="canvas" Background="Black"/>
```

.cs:
```cs
            this.canvas.Children.Clear();
            var dx = this.canvas.ActualWidth / M;
            var dy = this.canvas.ActualHeight / N;
            var x = new double[M + 1];
            var y = new double[N + 1];

            var source = PresentationSource.FromVisual(this.canvas);
            var transformToDevice = source.CompositionTarget.TransformToDevice;
            var transformFromDevice = source.CompositionTarget.TransformFromDevice;
            for (int i = 0; i <= M; i++)
            {
                var p = new Point(i * dx, 0);
                var pixel = transformToDevice.Transform(p);
                var adjusted = transformFromDevice.Transform(new Point(Math.Round(pixel.X), 0));
                x[i] = adjusted.X;
            }

            for (int j = 0; j <= N; j++)
            {
                var p = new Point(0, j * dy);
                var pixel = transformToDevice.Transform(p);
                var adjusted = transformFromDevice.Transform(new Point(0, Math.Round(pixel.Y)));
                y[j] = adjusted.Y;
            }

            for (int i = 0; i < M; i++)
            {
                for (int j = 0; j < N; j++)
                {
                    var r = new Rectangle
                    {
                        Fill = Brushes.White,
                        Width = x[i + 1] - x[i],
                        Height = y[j + 1] - y[j]
                    };
                    Canvas.SetLeft(r, x[i]);
                    Canvas.SetTop(r, y[j]);

                    this.canvas.Children.Add(r);
                }
            }
```

objorke:
Agree it does not make much sense to add `HeatMapRenderMethod.Polygons` - it is really the rectangle rendering that should be fixed.

objorke:
We could add a new method in `IRenderContext`:
```cs
        ScreenPoint SnapToDevicePixel(ScreenPoint p)
```
and implement a default in `RenderContextBase` that does nothing:
```cs
        public ScreenPoint SnapToDevicePixel(ScreenPoint p)
        {
            return p;
        }
```

for WPF `CanvasRenderContext` it could be something like:
```cs
        public ScreenPoint SnapToDevicePixel(ScreenPoint p)
        {
            var pixel = this.transformToDevice.Transform(this.ToPoint(p));
            var adjusted = this.transformFromDevice.Transform(new Point(Math.Round(pixel.X), Math.Round(pixel.Y)));
            return adjusted.ToScreenPoint();
        }
```

objorke:
I reopened the issue

VisualMelon:
I was worried that explicit snapping would cause different rectangles to be different dimensions (in screen pixels), but this is already an issue with axes and other series (e.g. `BarSeries`), so I don't see a problem with that. Indeed, if you zoom out enough, you can see the same gaps between rectangles with `BarSeries` (not surprising when you look at the code); it's essentially hidden by the same workaround this PR implements.

I was wondering if it might make sense for the render context to automatically snap to device pixels when it is rendering to the screen. I don't know if it makes sense for series' to handle the snapping: when would they not want to explicitly snap? It feels like the render context should be dealing with this 'behind the scenes', especially since this specific observation seems to be of specific concern for WPF/PDF. Part of the fun is that the extension methods for `DrawClippedRectangleAsPolygon` etc. always set `aliased` to `true`, whereby `CanvasRenderContext` attempts to align the screen points with the screen by offsetting them by half a pixel, so it's already doing this (sometimes (as I recall, `DrawRectangle` and `DrawRectangles` differ in behaviour)).

Issue #1056, opened 2017-01-27
Extend polar plot to full plot area


NilesDavis:
I would like to extend the polar plot to the full plot area which could look like this:
![lvkk](https://cloud.githubusercontent.com/assets/1811595/22368100/295660fa-e485-11e6-83a7-9e535a91d4e4.jpg)

At this very moment, I already have a modified AngleAxisRenderer and MagnitudeAxisRenderer which results in a Plot like this:
![polar1](https://cloud.githubusercontent.com/assets/1811595/22375976/8b1353c6-e4ac-11e6-9144-c7b5270dfd16.jpg)

Where I need help is to decide if this way is the right way to implement something like this. I do not have that much experience working in the oxyplot sourcecode so I am not sure if this is the right place to put this modifications. In the next step one has to be able to select the AxisRenderer in some way. At the Moment it is hard coded in the Axis itself. Or should I write new axis?
I would really appreciate some assitance!

objorke:
this looks good! It is possible the implementation of the axis base classes should be changed.
Can you share your changes in a fork so we can review?

NilesDavis:
Wow, half a year later I managed to come back to this issue, sorry for that! My fork with the changes can be found here: https://github.com/NilesDavis/oxyplot
I modified the AxisRenderers for MagnitudeAxis and AngleAxis and also extended the MagnitudeAxis to make it movable. I am really keen on feedback!

NilesDavis:
Hi @objorke, did you have a chance to take a lot at it? IMHO it makes sense to not touch the existing renderers and introduce new ones, but would that be the right way?

objorke:
Hi @NilesDavis, 
your screenshots above look very good, I hope we can include this function!

My remarks:
- can we restructure the axis renderers to avoid that the existing ones does net get more complicated? Maybe introduce a base class and make subclasses with the different ways of rendering. Then I think we can also avoid the DisplayStyle enum
- we need an example with the new functionality - in ExampleLibrary/Axes/PolarPlotExamples
- the coding style must follow the existing code (StyleCop, comments in English )
- make a branch and a PR just for this issue

The PrintIssue commit is related to another issue, right?

cheers, Øystein :-)

NilesDavis:
Hi @objorke ,
you are right, the PrintIssue is for #1138 where I am still search for someone to confirm ;)
[...]

PR #1209, opened 2018-05-02
Extended polar plot to full plot area, including Examples

NilesDavis:
This new style for polar plots is connected to issue #1056 and uses specific renderers and axis. Two examples derived from former polar plot examples were added in degrees and radians.
So far, just moving the plot center is enabled by right mouse button, but zooming in is still causing some unwanted effects. I hope this is useful and I'm looking forward on feedback.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 ### Added
+- New PolarPlot filling the full plot area (#1056)
+
+### Added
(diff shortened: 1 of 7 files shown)


NilesDavis:
Anyone had the chance to review this? @objorke perhaps? ;)

objorke:
sorry for the delay here, this looks great!

NilesDavis:
Hi @objorke , thanks a lot! I can imagine you are pretty busy, no worries about a late reaction. Thank you for checking it out!

Issue #1198, opened 2018-03-07
Legend symbol and text alignment is incorrect


Sidrivetest:
Steps to reproduce:

Create a SVG-Format Plotmodel with a legend. 

Here is my code:

var model = new PlotModel();

            var horizontalStrategy = new PaddingAxisSizeStrategy(_lineChart.Lines.HorizontalMaxCoordinate, _lineChart.HorizontalAxis);
            var horizontalGridInterval = _lineChart.HorizontalAxis.GridInterval ?? 0.0;
            var verticalGridInterval = _lineChart.VerticalAxis.GridInterval ?? 0.0;
            model.Axes.Add(CreateAxis(horizontalGridInterval, AxisPosition.Bottom, _translations.HorizontalAxisTitle, horizontalStrategy));
            model.Axes.Add(CreateAxis(verticalGridInterval, AxisPosition.Left, _translations.VerticalAxisTitle, new GridDistancingAxisSizeStrategy(_lineChart)));

            model.SetTopPaddingTo(20.0);

            model.PlotMargins = new OxyThickness(50.0, 10.0, 50.0, 50.0);
            model.LegendMargin = 0.5;

            model.Background = OxyColors.White;
            model.AddLines(_lineChart.Lines, _translations.LineLegends);

            model.LegendOrientation = LegendOrientation.Horizontal;
            model.LegendPosition = LegendPosition.LeftTop;

### Expected behaviour

I want to generate a legend where the  symbol alignment is set correctly to the respective text.

### Actual behaviour

The height of the text is not matching exactly the height of the symbol. It has a upper displacement.
Like this:

![attachfilehandler](https://user-images.githubusercontent.com/37150353/37100886-726de14a-2224-11e8-8330-a1eea28bc92b.png)

**This bug had his first occurence after changing the format from PNG to SVG**
**The problem is that we are obligated to use the SVG format!**

VisualMelon:
This is probably the same issues as #1146, which is a distilled version of #459.

I did some work on a workaround for this (documented in #459) as a [fork](https://github.com/VisualMelon/oxyplot/tree/svgvalign) ([diff](https://github.com/VisualMelon/oxyplot/commit/b3472d6343f0d8bb612dcc4646c24f8034e25ef1)) a while back, but I'm less than well informed regarding SVG.

If you can use a different `IRenderContext` for the `textMeasurer` parameter when exporting, you may see better results.

Issue #459, opened 2015-04-21
SVG not rendered correctly


issame:
Hi

I noticed that the Axis label doesnt get properly placed when the Model is Exported to SVG:

PlotView in WinForms
![plot_winform](https://cloud.githubusercontent.com/assets/1009013/7250179/06a7c474-e818-11e4-99aa-57c59d56747c.PNG)

SVG
![error_svg](https://cloud.githubusercontent.com/assets/1009013/7250183/0b740260-e818-11e4-942d-4b1eb3dba559.PNG)

objorke:
What application are you using to view the SVG? I have seen some differences e.g. when using IE and Chrome as viewers.

issame:
Actually, I am using the SVG for creating a latex report (which internally uses inkscape)

issame:
The same thing happens with ImageMagick... maybe giving the SvgExporter an offset parameter to adjust Position...
![err1](https://cloud.githubusercontent.com/assets/1009013/7254773/04170a60-e83d-11e4-848e-4da64f222b6c.PNG)
![err2](https://cloud.githubusercontent.com/assets/1009013/7254786/1b91158c-e83d-11e4-8417-46300599b1ea.PNG)

issame:
I think that the main issue is because of Text being render as text, so latex processor, firefox, opera, chrome, inkscape... render the text in a different way. Maybe converting the text to a drawing will solve this issue, so it will look the same everywhere

objorke:
Are you using a standard font? Maybe the problem is the text measurements... The svg rendering context depends on text measurements to do the horizontal and vertical alignments. I have not investigated whether it is possible to use the `text-anchor` etc. attributes of the svg `text` element. Also, I did not see any attributes controlling vertical alignment.

https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/text-anchor
https://developer.mozilla.org/en-US/docs/Web/SVG/Element/text#Specific_attributes 

Sorry, I don't understand what you meant about converting text to drawing. Do you mean filling a `path`  with the vectorized text?

issame:
Yes.

Chrisss50:
As this is labled as an unconfirmed bug, I want to give an update to this issue.
I can reproduce the bug by opening the exported svg-file with CorelDraw, Internet Explorer (okay, forget about that one) and Inkscape.

[...]

PR #1216, opened 2018-06-13
SVG dominant-baseline Vertical Alignment Workaround

### LINKED ISSUE #1146: PortableDocumentFont Multiline Text
`PortableDocumentFont` has no support for multi-line text.

### Steps to reproduce

Using a `TextAnnotation` with `SvgExporter` using the `PortableDocumentFont` text measuring (i.e. default parameters) results in an SVG with clearly mangled output.

            var model = new PlotModel { Title = "SVG Text Rendering" };
            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom });
            
            model.Annotations.Add(new OxyPlot.Annotations.TextAnnotation()
            {
                Text = "Hello1\r\nHello2\r\nHello3",
                TextPosition = new DataPoint(50, 50),
            });
            
            this.Refresh();
            
            SvgExporter svgExporter = new SvgExporter();
            using (var fs = System.IO.File.Open("test2.svg", System.IO.FileMode.Create))
            {
                svgExporter.Export(model, fs);
            }

Platform: Windows 7
.NET version: Framework 4.6.1

_(The code in  clearly has no support for mutli-line text, so external factors are unlikely to be relevant to this issue)_

### Expected behaviour

The box should bound the text. Permitting for the dodgy SVG rendering, the box should bound a hypothetical 3 lines of text. [OxyplotFiddle of Expected Behaviour](https://oxyplotfiddle.azurewebsites.net/?var%20model%20%3D%20new%20PlotModel%20%7B%20Title%20%3D%20%22OxyPlot%20on%20Azure!%22%20%7D%3B%0Amodel.Axes.Add(new%20LinearAxis%20%7B%20Position%20%3D%20AxisPosition.Left%20%7D)%3B%0Amodel.Axes.Add(new%20LinearAxis%20%7B%20Position%20%3D%20AxisPosition.Bottom%20%7D)%3B%0Amodel.Series.Add(new%20FunctionSeries(Math.Sin%2C%200%2C%2010%2C%201000))%3B%0A%0A%20%20%20%20%20%20%20%20%20%20%20%20model.Annotations.Add(new%20OxyPlot.Annotations.TextAnnotation()%0A%20%20%20%20%20%20%20%20%20%20%20%20%7B%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20Text%20%3D%20%22Hello1%5Cr%5CnHello2%5Cr%5CnHello3%22%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20TextPosition%20%3D%20new%20DataPoint(1%2C%200)%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%7D)%3B%0A%20%20%20%20%20%20%20%20%20%20%20%20%0A%0Areturn%20model%3B%0A)

### Actual behaviour

The bounding box bounds the text as if with `\r\n` removed. This is a consequence of the `TextAnnotation` measuring the text with the `SvgRenderContext`, which (in this instance) is using the `ProtableDocument` class for text measuring, and consequently `PortableDocumentFont`.

The lines of text being the wrong way round is a side-effect of issues with SVG Text Rendering (#459)

portdocfontlines

### Proposed Solution

Though this class is inherently dodgy, and essentially a fall-back, it would be simple to produce a better result by splitting measured text into lines, determining the width of the longest line, and multiplying the height by the number of lines, in the method `PortableDocumentFont.Measure`.

VisualMelon:
Fixes #1146, #459 (#1198).

### Checklist

- Multi-line text support for `PortableDocumentFont`, the default text measurer for `SvgExporter` (#1146)
- A workaround (enabled by Boolean flag) for manual vertical text alignment in `SvgExporter`, to accommodate viewers which do not support the `dominant-baseline` attribute, such as Microsoft Internet Explorer & Edge (#549).

### Outstanding concerns

 - Though a workaround is never ideal, the age and steady reoccurrence of this issue as well as the fact that no good solution exists (that I am aware of) suggests to me that this might be an acceptable change. If not, perhaps an alternative method of exposing the functionality could be agreed.

 - The Boolean flag on the static method `OxyPlot.SvgExporter.Export` is a binary breaking chance, but without it using the workaround would require manually assembling a `SvgRenderContext` and `SvgExporter`, which may be perfectly acceptable given this is a workaround. Other such default parameters also constitute binary breaking changes; if the default parameter for `OxyPlot.SvgExporter.Export` is undesirable, then these too should surely go. I'd be in favour of removing these, but I'd appreciate feedback to this effect.

 - Example outputs can be found in #549

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ All notable changes to this project will be documented in this file.
 - RectangleSeries (#1060)
 - InvalidNumberColor on Wpf.LinearColorAxis (#1087)
 - ContinuousHistogramSeries (#1145)
+- Multiline text support for PortableDocumentFont (#1146)
+- Workaround for text vertical alignment in SVG Export to accomodate viewers which don't support dominant-baseline (#459, #1198)
 
 ### Changed
 - Let Gtk# PlotView show up in Ui editor ToolBox of MonoDevelop and XamarinStudio (#1071)
--- a/Source/OxyPlot/Pdf/PortableDocumentFont.cs
+++ b/Source/OxyPlot/Pdf/PortableDocumentFont.cs
@@ -9,6 +9,8 @@
 
 namespace OxyPlot
 {
+    using System.Text.RegularExpressions;
+
     /// <summary>
     /// Represents a font that can be used in a <see cref="PortableDocument" />.
     /// </summary>
@@ -102,21 +104,35 @@ namespace OxyPlot
         /// <param name="height">The height of the text.</param>
         public void Measure(string text, double fontSize, out double width, out double height)
         {
-            int w = 0;
+            int wmax = 0;
 
-            for (int i = 0; i < text.Length; i++)
+            var lines = Regex.Split(text, "\r\n");
+
+            int lineCount = lines.Length;
+
+            foreach (string line in lines)
             {
-                var c = text[i];
-                if (c >= this.FirstChar + this.Widths.Length)
+                int w = 0;
+
+                for (int i = 0; i < line.Length; i++)
                 {
-                    continue;
+                    var c = line[i];
+                    if (c >= this.FirstChar + this.Widths.Length)
+                    {
+                        continue;
+                    }
+
+                    w += this.Widths[c - this.FirstChar];
                 }
 
-                w += this.Widths[text[i] - this.FirstChar];
+                if (w > wmax)
+                {
+                    wmax = w;
+                }
             }
 
-            width = w * fontSize / 1000;
-            height = (this.Ascent - this.Descent) * fontSize / 1000;
+            width = wmax * fontSize / 1000;
+            height = lineCount * (this.Ascent - this.Descent) * fontSize / 1000;
         }
     }
 }
--- a/Source/OxyPlot/Svg/SvgExporter.cs
+++ b/Source/OxyPlot/Svg/SvgExporter.cs
@@ -40,6 +40,11 @@ namespace OxyPlot
         /// Gets or sets a value indicating whether the xml headers should be included.
         /// </summary>
         public bool IsDocument { get; set; }
+        
+        /// <summary>
+        /// Gets or sets a value indicating whether to use a workaround for vertical text alignment to support renderers with limited support for the dominate-baseline attribute.
+        /// </summary>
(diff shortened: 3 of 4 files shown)


objorke:
sorry for the delay!
agree we can accept the workaround for text alignment here!

VisualMelon:
Are you sure you're happy with the overloading? The only reason I didn't change it in the first instance was because I knew atleast one person was using my fork, and didn't want to break it a hurry. I'd be perfectly happy to make the appropriate changes now that you've OK'd the workaround itself.

I do believe it is a breaking change, which is less than ideal.

VisualMelon:
Binary breaking change in the sense that code compiled against the new assembly will work fine, but the same code compiled against the old assembly (which uses this method) will not (i.e. to 'fix' it you need to recompile against the new assembly but nothing else; no semantic changes or anything scary).

I could be wrong, but my understanding of the way implicit operators are implemented in C# is that they are baked into the calling assembly at compile time; duly, there is no method without the parameter available for code compiled against an old assembly to find. See [this answer on StackOverflow](https://stackoverflow.com/a/23517490).

Guidemarcus:
Hi. I am pretty new to Oxyplot and .Net development. I have read the issue #459 and I absolutely need to use a stable version of the library (version 1.0.0). I understand that there is a way to fix the text alignment issue by changing the TextMeasurer property, but what do you changing it for? I have tried : 
            Dim tempCanvas = New Canvas()
            tempCanvas.VerticalAlignment = Windows.VerticalAlignment.Bottom
            exporter.TextMeasurer = New CanvasRenderContext(tempCanvas)
but it does not fix anything. Otherwise, I have tried this too : 
           exporter.TextMeasurer = New CanvasRenderContext(New Canvas())
Do you have anything else you can think of that I can try?

VisualMelon:
@Guidemarcus sorry, I've probably mislead you (as a result of confusing myself): a different text renderer only resolves the multi-line text issues; the workaround is still necessary to sort out alignment.

Until the next stable version, can you C&P the important SVG code (e.g. from these commits; there isn't much at all) and use it instead?

Guidemarcus:
Thank you for your help. For now, I will use the code from the last commits. That will do the job just fine.

PR #1254, opened 2018-10-26
Fix a problem with inifity size polyline

mennodeij:
Issue #1303 

The problem solved is illustrated in PlotViewTests.cs:110-136.
In short, when plotting large Y-ranges, the WPF measuring of a `System.Windows.Shapes.Polyline` may return an invalid size containing `double.Infinity`. This has been fixed by reimplementing the polyline class with a fix in `MeasureOverride` (the WPF implementation is `sealed` unfortunately, so no inheritance could be used).

### Checklist

- Replace the `System.Windows.Shapes.Polyline` with a fixed re-implementation (`Polyline2`). 

@oxyplot/admins


mennodeij:
If other platforms handle this without single precision overflow, the only fix should be in the WPF platform. Is there a way to test this in CI on all platforms? I do not have access to other platforms other than GTK.

objorke:
I would guess UWP and maybe Xamarin.Forms have similar issues. It probably needs to be tested manually.
I still think this could be solved within the axis classes - if the coordinates transformed into "screen space" is unreasonable large, they could be "clipped" or ignored?

objorke:
I suggest an alternative solution in #1365 

The WPF Polyline should not be modified, I think this can be solved in OxyPlot.Wpf.

objorke:
Merged #1365

PR #1272, opened 2019-01-25
Fixed bug in magnitude axis text rendering over plotarea bounds

NilesDavis:
Fixes # .
The texts along the magnitude axis were rendered right for 0° orientation. On other orientations the textelements happened to be too long or too short. Now all startangles are supported

### Checklist

-
-
-

@oxyplot/admins


objorke:
Can you create an issue we can link to?
Can you add an example where this can be tested? It is best if the example is committed first, then it is easier to verify that the change fixes the bug.

objorke:
I created #1363 as I understand this issue.

objorke:
This PR still doesn't solve the problem properly - labels below 1.0 shows outside the plot area:

image

objorke:
I also found another issue #1364

objorke:
I think this PR is an improvement, so let us merge it. But the two issues remains open.

PR #1282, opened 2019-02-12
Change TrackerControl to reuse existing ContentControls

### LINKED ISSUE #1281: Tracker controls are unnecessarily re-created for each tracker update
### Steps to reproduce

1. Call `ShowTracker()` with a new tracker hit result
2. Note that a new `ContentControl` is created even if the new hit result uses the same tracker template (default or series specific)

Platform: WPF
.NET version: 4.7

### Expected behaviour

If the new hit result uses the same template as the currently displayed tracker, it should simply update the data context of the existing tracker

### Actual behaviour

A new `ContentControl` is created every time the tracker hit result is updated, regardless of template.

It looks like there was an attempt to avoid this (or a similar) scenario, but because it uses a `ReferenceEquals` to compare an object new'd two lines above with a member variable, this test always fails and thus a new instance is always created

```
            var tracker = new ContentControl { Template = trackerTemplate };

            // ReSharper disable once RedundantNameQualifier
            if (!object.ReferenceEquals(tracker, this.currentTracker))
            {
                this.HideTracker();
                this.overlays.Children.Add(tracker);
                this.currentTracker = tracker;
            }
```

I'm currently testing a fix and will raise a PR for it once I've checked it works as expected

danpaul88:
Fixes #1281 .

### Checklist

- [ ] I have included examples or tests (*see notes*)

The tracker control has an additional private member variable to track the template which was assigned to the current content control. When `ShowTracker` is called whilst a tracker is currently active AND the template choice is the same as the currently active tracker (default or series specific) then it will simply update the data context of the existing content control rather than creating a new one.

I believe the changes are stylecop compliant but as I'm not a regular user of it I'm unsure if I'm using it correctly - please advise if any changes are required.

As this is an optimisation rather than a UI visible change, I'm not sure what to provide as an example?

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.Wpf/PlotBase.cs
+++ b/Source/OxyPlot.Wpf/PlotBase.cs
@@ -52,6 +52,11 @@ namespace OxyPlot.Wpf
         private FrameworkElement currentTracker;
 
         /// <summary>
+        /// The current tracker template.
+        /// </summary>
+        private ControlTemplate currentTrackerTemplate;
+
+        /// <summary>
         /// The grid.
         /// </summary>
         private Grid grid;
@@ -181,6 +186,7 @@ namespace OxyPlot.Wpf
             {
                 this.overlays.Children.Remove(this.currentTracker);
                 this.currentTracker = null;
+                this.currentTrackerTemplate = null;
             }
         }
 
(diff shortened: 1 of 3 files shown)


Source/OxyPlot.Wpf/PlotBase.cs
```diff
@@ -341,14 +347,15 @@ public void ShowTracker(TrackerHitResult trackerHitResult)
                 return;
             }
 
-            var tracker = new ContentControl { Template = trackerTemplate };
-
             // ReSharper disable once RedundantNameQualifier
-            if (!object.ReferenceEquals(tracker, this.currentTracker))
+            if (!object.ReferenceEquals(trackerTemplate, this.currentTrackerTemplate))
```
objorke: is this a breaking change? will old code not using a tracker template still work?

danpaul88: I don't believe so - in the cases where the content of the if block is now skipped (previously it was *always* evaluated since a brand new local variable created on the line above is never the same reference as a member variable) we still set a new DataContext on the control (line 363 in the branch), so even OneTime data bindings would be refreshed.

I tested various scenarios to ensure this is the case, but obviously it's impossible to test this against every possible scenario.

objorke: OK, I will merge this one. Sorry for the delay!

PR #1289, opened 2019-04-11
Fix Catmull-Rom Splines

### LINKED ISSUE #476: Performance of LineSeries with Smoothing is bad
The rendering performance and responsiveness of a plot with LineSeries and Smooth=true is very bad.
Already with a few points you see the issue.

Ideas
- don't re-calculate the smoothed points on every rendering (need a strategy when to re-calculate)
- add some collection changed event to the `Points` collection to mark series as dirty (needs re-calculate)

### LINKED ISSUE #1237: Extreme zoom on catmull-rom spline causes hang
### Steps to reproduce

1. Go to the LineSeries "Catmull-Rom interpolation" Example
2. Using Ctrl-RMB drag, zoom in on one of the points on the graph
3. Continue zooming in again and again on the same point
4. Eventually, observe that the chart hangs and stops responding

Platform: WPF, Windows 10 x64
.NET version: 4.5.2

### Expected behaviour
The chart should not take more than a few milliseconds to render three 2 point splines

### Actual behaviour
The application hangs and stops responding. In some cases an out of memory exception was thrown

I have investigated this and the cause is as follows:

In CatmullRomSpline.cs line 207-208
```
int iterations = Math.Min((int)((Math.Abs(pt1.X - pt2.X) + Math.Abs(pt1.Y - pt2.Y)) / tolerance), maxSegments);
for (double t = t1; t < t2; t += (t2 - t1) / iterations)
{...
```

When the zoom is too extreme, the point coordinates become extremely large, these are transformed screen coordinates of data points that can be effectively kilometers offscreen. This can lead to overflows. int iterations is capped at maxSegments, but that is always 1000 and in extreme zoom cases the line is almost straight making 1000 iterations a massive waste of resources.

Worse though, it is possible (even common) for iterations to evaluate to -2147483648 making this effectively an infinite loop.

Suggested fix:
A quick fix would be simply to check if iterations evaluations to a negative number and set it to 4 if it does.
A nicer fix would be to somehow determine if extreme zoom is currently in effect, e.g. if the points are really far away, and then just set the iterations to 4.

Dunkhan:
Catmull-Rom splines: Changed default segments to 100 from 1000, made segment count settable by user. Fixed integer overflows causing infinite iterations.

Fixes #1237 
Should also fix #476 in all cases where this spline type is used

### Checklist

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 - FontSize, FontWeight and FontFamily on Wpf.TextAnnotation (#1023)
 - RectangleSeries (#1060)
 - InvalidNumberColor on Wpf.LinearColorAxis (#1087)
+- ContinuousHistogramSeries (#1145)
 
 ### Changed
 - Let Gtk# PlotView show up in Ui editor ToolBox of MonoDevelop and XamarinStudio (#1071)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,7 @@ All notable changes to this project will be documented in this file.
 - OverflowException when zoomed in on logarithmic axis (#1090)
 - ScatterSeries with DateTimeAxis/TimeSpanAxis (#1132)
 - Exporting TextAnnotation with TextColor having 255 alpha to SVG produces opaque text (#1160)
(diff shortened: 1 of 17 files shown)


Dunkhan:
Argh I screwed that up and forgot to sync my fork before editing. I will kill this PR and make a new one its the simplest way

PR #1294, opened 2019-04-27
Brushdisposal

### LINKED ISSUE #1230: OxyPlot.WindowsForms.GraphicsRenderContext leaks SolidBrush objects
### Steps to reproduce

1. Plot a lot of filled in polygons
2. Use a machine which doesn't finalize often
3. Watch the memory usage hike

leakanon

Platform: WindowsForms (only observed on Windows 10, Win7 machine seems to be finalising more aggressively)
.NET version: .NET Framework (not sure)

### Partial Proposed Solution

Some parts of `GraphicsRenderContext` clearly produce `SolidBrush` objects without disposing them: [`GraphicsRenderContext.DrawPolygon`](https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot.WindowsForms/GraphicsRenderContext.cs#L174)

Replacing all calls to `ToBrush(this OxyColor)` with `GetCachedBrush(OxyColor)` (excepting those within `GetCachedBrush`, and `PngExporter`) seems an easy and reasonable solution; there is no apparent pattern to which methods explicitly cache their brushes and which do not.

Cached `Brushes` are correctly disposed by `PngExporter`, __but seemly not when used as part of a `PlotView`__ (which warrants discussion). The `GraphicsRenderObject` is also not disposed by the winforms `SvgExporter`, which it should be, though this won't be leaking `SolidBrush` objects as it is only employed for text-measuring.

### Other Related

Currently the `GraphicsRenderContext` only caches `Brush` objects. With some care, it could cache `Pen` objects also. I'm also not convinced by the method [`CleanUp`](https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot.WindowsForms/GraphicsRenderContext.cs#L353), which is the only place that removes elements from `imagesInUse`, which I think means it will only clean up images the _second_ time it is called, meaning single-use instances (such as in `PngContext`) will not dispose `Image` objects.

I don't understand the rationale behind the existing code here, so I won't attempt to suggest an alternative, but `imagesInUse` appears to at the centre of the problem.

VisualMelon:
Fixes #1230 concerning Brush disposal in `GraphicsRenderContext` and related issues with Pens and Images.

### Checklist

- [x] I have cleaned up the commit history (use rebase and squash) (maintained 2 separate commits, one for brushes, one for pens)

- Unify usage of `GetCachedBrush` so that all brushes are cached
- Adds the `GraphicsPenDescription` class, which describes a GDI+ solid-color pen
- Introduce corresponding `GetCachedPen`, and use everywhere so that all pens are cached
- Correctly dispose Brushes, Pens, and Images in `GraphicsRenderContext`
- Correctly dispose `GraphicsRenderContext` in `Oxyplot.WindowsForms.PlotView`

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -55,6 +55,7 @@ All notable changes to this project will be documented in this file.
 - Candle overlap each candle (#623)
 - CandleStick is overlapped when item.open == item.close in the CandleStickAndVolumeSeries (#1245)
 - Out of memory exception and performance issue with Catmull-Rom Spline (#1237)
+- Cache and Dispose Brush and Pen objects used by GraphicsRenderContext (#1230)
 
 ## [1.0.0] - 2016-09-11
 ### Added
--- a/Source/OxyPlot.WindowsForms/GraphicsPenDescription.cs
+++ b/Source/OxyPlot.WindowsForms/GraphicsPenDescription.cs
@@ -1,0 +1,163 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OxyPlot.WindowsForms
+{
+    /// <summary>
+    /// Describes a GDI+ Pen.
+    /// </summary>
+    public class GraphicsPenDescription
+    {
+        /// <summary>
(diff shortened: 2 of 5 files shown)


objorke:
Looks good!

Issue #475, opened 2015-05-12
Move data binding (ItemsSource support) out of core series and DataPoint


tibel:
The data binding feature is (mostly) used when defining a plot in XAML only.
But the binding logic adds much complexity to these classes (series and datapoint).

I think it would be better to move the data binding logic into some adapter classes.
Thus the core model will become easier to understand and optimize.
Also the whole databinding logic is in one place and can be optimized separately.

objorke:
Good idea, I agree this could make the core implementation less complex.

PR #1305, opened 2019-06-20
Three color histogram

rusamentiaga:
Support for three colors in histogram series:

image

### Checklist

- Support for three colors in histogram series (Issue #1330)

@oxyplot/admins


VisualMelon:
I'm glad someone else is using this; if you have any feedback on the Histogram API generally, I'd be interested to hear it.

I'm not sure I like adding code specifically for handling 3 colours (and I don't like the names `Hi` and `Lo`), but I can see how custom colouring would be useful. It's also not ideal having 2 inter-dependent properties; if the threshold and colour information were packaged in another object, this (minor) concern would go away.

Would it be better to provide a simple colouring API which the Histogram Series can consume, which allows user-specified colouring behaviour?

Something like

    `hs.Colorizer = new ThreeColourColorizer(low, high, lowColor, highColor)`

(Colorizer probably isn't a good name)

And internally the HistorgramSeries defers color choice to the colorizer based on the HistogramItem. The existing color properties can be defaults (if no colourizer is provided/it returns 'Automatic').

It might also be useful for the colorizer to be computed from the data, that way a standalone `ConfidenceIntervalColorizer` could be provided which doesn't require the user to update the colorizer with the data (it can be done by the model). I can think of a couple of ways this could be done.

This would obviously add complexity, but it shouldn't mess with the existing usage. I'd be more than happy to work with you (or anyone) on working out a sensible API _soon_ (it wouldn't make sense to merge this if it will become redundant with another more generic feature). The API would take into consideration how it might map to other series, and how it might class with colour-axes, which the `Histogram` series currently ignores.

In terms of performance, a virtual call per box will obviously be slower than a pair of comparisons, but the cost will only be present for those with a non-`null` colorizer, and should be insignificant compared to the cost of rendering.

rusamentiaga:
> 
> 
> I'm glad someone else is using this; if you have any feedback on the Histogram API generally, I'd be interested to hear it.
> 

I find it very easy to use. I was thinking about introducing some more options to HistogramHelpers, such as heuristics for determining the number of bins (example: Accord.net, BinAdjustmentRule)

> I'm not sure I like adding code specifically for handling 3 colours (and I don't like the names `Hi` and `Lo`), but I can see how custom colouring would be useful. It's also not ideal having 2 inter-dependent properties; if the threshold and colour information were packaged in another object, this (minor) concern would go away.

I agree, but I as trying to mimic the same API used in ThreeColorLineSeries, series already in the main branch.

> 
> Would it be better to provide a simple colouring API which the Histogram Series can consume, which allows user-specified colouring behaviour?
> 
> Something like
> 
> ```
> `hs.Colorizer = new ThreeColourColorizer(low, high, lowColor, highColor)`
> ```
> 
> (Colorizer probably isn't a good name)

Yes, but should this be used in ThreeColorLineSeries too?

> And internally the HistorgramSeries defers color choice to the colorizer based on the HistogramItem. The existing color properties can be defaults (if no colourizer is provided/it returns 'Automatic').
> 

I like the idea. The truth is that I was not sure about the if-else/if-else structure in RenderBins, but I did not want to break compatibility with NegativeFillColor.

> It might also be useful for the colorizer to be computed from the data, that way a standalone `ConfidenceIntervalColorizer` could be provided which doesn't require the user to update the colorizer with the data (it can be done by the model). I can think of a couple of ways this could be done.
> 

This could be problematic when the user requires absolute thresholds (my case), the confidence interval is a particular scenario. Limits could be also established base on percentiles. This could be an API that calculates absolute thresholds based on data.

> This would obviously add complexity, but it shouldn't mess with the existing usage. I'd be more than happy to work with you (or anyone) on working out a sensible API _soon_ (it wouldn't make sense to merge this if it will become redundant with another more generic feature). The API would take into consideration how it might map to other series, and how it might class with colour-axes, which the `Histogram` series currently ignores.
> 
> In terms of performance, a virtual call per box will obviously be slower than a pair of comparisons, but the cost will only be present for those with a non-`null` colorizer, and should be insignificant compared to the cost of rendering.

I would say performance should not be an issue, since the number of bins is generally low.

VisualMelon:
> I was thinking about introducing some more options to HistogramHelpers, such as heuristics for determining the number of bins (example: Accord.net, BinAdjustmentRule)

I would personally welcome any such additions!

> Yes, but should this be used in ThreeColorLineSeries too?

It's probably a bit late now for `ThreeColorLineSeries`, but could make a sensible addition to the 'normal' line series and other plots (Scatter, rectangle, etc.). Obviously these all have a different idea of what a data-sample is, so if that was a goal it would need to be generic over the data-type (which would be fine by me).

> but I did not want to break compatibility with NegativeFillColor.

Yeah... I'm kind of regretting adding that piece of 'functionality'. With the colorizer, it would just be another 'default' color should there be no colorizer or it requests the 'automatic' color.

> This could be problematic when the user requires absolute thresholds (my case), the confidence interval is a particular scenario

That should fit in just fine either way: we can provide an implementation which takes absolute thresholds and ignores the data. Having the option to compute from data would just (I think) make it a little easier to 'throw' an adaptive colorizer into an existing application (e.g. in the work I do, a histogram could appear from anywhere, and the original data may no longer once it gets to the formatting stage). That said, computing metrics from binned data is just asking for trouble, so perhaps updating with the data shouldn't be a goal at the moment.

> I would say performance should not be an issue, since the number of bins is generally low.

Personally, I often produce histograms with thousands of bins (which I then zoom), but I think you are probably right.

VisualMelon:
Further regarding `ThreeColorLineSeries`, it does more than just color the samples: it also colors the line, which would be incompatible with a per-sample colorizer.

objorke:
Thanks for the PR, @rusamentiaga !
1. Please create an issue before making a PR :-)
2. I think this could be implemented with a general `Func<double, OxyColor>` or `Func<object,OxyColor>` property, which could support many more ways of colouring the bars. Similar to https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Series/ScatterSeries%7BT%7D.cs#L81. It would also mean less custom properties in the class.

rusamentiaga:
I have updated the code following the suggestion. Definitely much better solution.

objorke:
I had a second thought about this. Maybe the core model in this library should be as simple as possible, e.g. no support for data binding, mapping functions etc. @tibel suggested something similar in #475. E.g. no mapping functions and data binding functionality in any of the series classes. There could be a `Color` property in the `HistogramItem` class. Which could be overriden by the color in the `HistogramSeries` if it is not set. What do you think?

VisualMelon:
I wouldn't object to adding a `Color` property to `HistogramItem` (and I'd be happy to do that work if no-one is in a hurry).

<hr />

Concerning bindings and mappings, It hasn't been my experience that they get in the way of other work on the library. They can be useful for doing work at the 'presentation' end of things without increasing the complexity of whatever process is accumulating the data (e.g. having to intercept it and recolor it, or try to hook an update event to recolor it then, etc.). They can feel a bit hacky, but I don't think the library suffers because of them. This is not to suggest that 'they are cheap' is a good reason for introducing more, I just don't think simple mappings like this cost too much in practice.

I would also argue that presentation stuff should probably be independent of the data stuff. This has implications for performance: e.g. if a series can determine that a point won't be rendered (out of bound, spline reduction), then it shouldn't have to determine how it will be rendered. If a unified mechanism for data binding is arranged, there should be one also for data-presentation.

objorke:
I added #1347, I think this is the way to go

VisualMelon:
With `HistogramSeries.ColorMapping` and `HistogramItem.Color` added, I think this can now be closed.

PR #1307, opened 2019-06-21
Fix AxisChangedEventArgs.DeltaMaximum in Axes.Reset

### LINKED ISSUE #1306: Reset axis - possible issue
While working with oxyplot library, am trying to reset the axe and make it so that the last data are shown, I keep getting a problem doing that, so I tried checking the source code, 
I am not sure if this is the reason behind my issue, but by instinct I think this should be changed to keep old maximum and not old minimum twice

https://github.com/oxyplot/oxyplot/blob/2c254dc8352f59b927e7f6600071fcfa5fe764e7/Source/OxyPlot/Axes/Axis.cs#L967

VisualMelon:
Fixes #1306

### Checklist

- Change the computation of the `AxisChangedEventArgs.DeltaMaximum` value provided by `Axis.Reset` to depend upon the previous maximum, rather than previous minimum.

@oxyplot/admins


objorke:
Thanks @VisualMelon

PR #1346, opened 2019-09-04
Add Support for transposed XYAxisSeries (#1334)

### LINKED ISSUE #1334: Add Support for transposed XYAxisSeries
Related to #46 .

I recently hacked together a `HorizontalBoxplotSeries` to make plots like this possible:
histogram with boxplot

I think it would be really nice to have this in the main library and I'd be willing to make that contribution.

I haven't quite yet decided what would be the best way of doing this. I see three possibilities:

1. Make `HorizontalBoxPlotSeries` inherit from `BoxPlotSeries` and override the relevant stuff (which is a lot: rendering, tracker, updatemaxmin etc.)
That's the quick and ugly way I'm using until now. We probably want to do it a bit more elegant.

2. Make `HorizontalBoxPlotSeries` completely independent from BoxPlotSeries
We could reuse the existing `BoxPlotItem`, but there would be a lot of duplicated code between `BoxPlotSeries` and `HorizontalBoxPlotSeries`.

3. Introduce abstract `BoxPlotSeriesBase` and make both `BoxPlotSeries` and `HorizontalBoxPlotSeries` inherit from it
This is the solution I'm leaning towards at the moment. This would minimize code duplication.
Possibly we would want to rename the properties of `BoxPlotItem` to more generic terms (e.g. `BoxMaximum` instead of `BoxTop`) while keeping the old properties as obsolete.

What do you think? Is there a better solution I haven't thought of?

Jonarw:
Fixes #1334 .

### Checklist

-
-
-

@oxyplot/admins


VisualMelon:
Are there any exciting changes since the version discussed in #1334?

Jonarw:
No, not really. I implemented most of your feedback and fixed some minor documentation stuff. Other than that I just rebased and squashed before making the pull request.

objorke:
Very good! Impressive contribution :-) And I like that you have removed and cleaned up a lot of code!

PR #1353, opened 2019-09-09
Added new LineLegendPositions.

SecondShiftEngineer:
Fixes # .

### Checklist

- Added some line legend position options.

@oxyplot/admins


CHANGELOG.md
```diff
@@ -270,6 +270,9 @@ All notable changes to this project will be documented in this file.
 - Add overridable Axis.FormatValueOverride (#181)
 - PngExporter text formatting (#170)
 
+### Added
+  Added additional line legend positions to line series. It was previously only Start and End.
```
VisualMelon: Should probably be in the `Unreleased` section at the top.

Source/Examples/ExampleLibrary/Series/LineSeriesExamples.cs
```diff
@@ -206,7 +206,7 @@ public static PlotModel CustomLineLegendPosition()
             var model = new PlotModel { Title = "LineSeries with LineLegendPosition" };
             model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom, MinimumPadding = 0.1, MaximumPadding = 0.1 });
             model.Axes.Add(new LinearAxis { Position = AxisPosition.Left });
-            var s1 = CreateExampleLineSeries();
+            var s1 = CreateExampleLineSeries(246);
```
VisualMelon: Little more white-space before `s1` and `s3` would be nice.

Source/OxyPlot/Series/LineLegendPosition.cs
```diff
@@ -24,9 +24,33 @@ public enum LineLegendPosition
...
+
+        /// <summary>
+        /// Render legend below the midpoint of the line inside the plot area.
+        /// </summary>
+        HalfWayBelow,
+
+        /// <summary>
+        /// Render legend below the midpoint of the line inside the plot area. This will follow the slope of the line.
+        /// </summary>
+        HalfWayBelowSloped,
```
VisualMelon: Addition whitespite would be nice.

Source/OxyPlot/Series/LineSeries.cs
```diff
@@ -672,25 +672,70 @@ protected void RenderLegendOnLine(IRenderContext rc)
...
+                    va = VerticalAlignment.Top;
+                    break;
+                case LineLegendPosition.End:
                     // end position
+                    dy = 4;
                     point = this.ActualPoints[this.ActualPoints.Count - 1];
-                    dx = 4;
                     break;
+                default:
+                    //was end previously
```
VisualMelon: Note: this is a breaking change.

VisualMelon:
I've never used this feature, but it looks OK to me. Would an `EndInside` be useful also?

I think the change to throw on `default` is probably a good one, but changing the position of `End` in `LineLegendPosition` is a binary breaking change, and I think that should be avoided.

Source/OxyPlot/Series/LineLegendPosition.cs
```diff
@@ -24,9 +24,33 @@ public enum LineLegendPosition
...
         /// <summary>
         /// Render legend at the end of the line.
         /// </summary>
         End
```
VisualMelon: Binary breaking change (change of constant)

CHANGELOG.md
SecondShiftEngineer: I moved this to the unreleased section.

Source/Examples/ExampleLibrary/Series/LineSeriesExamples.cs
SecondShiftEngineer: I added some white space in this section to make it cleaner.

Source/OxyPlot/Series/LineLegendPosition.cs
SecondShiftEngineer: I added the missing white space.

Source/OxyPlot/Series/LineSeries.cs
SecondShiftEngineer: I removed the comment. Originally the End case would go to the default as if it was an if else statement. I added the end case so there should be no backwards compatibility issues.

SecondShiftEngineer:
I need to double check to see if my commits are up to date. Something looks like it might be missing.

Jonarw:
I see a couple minor issues with this - some of these were already present before this PR, but it would be nice to address them anyway :)

1. The LineLegend text is not clipped (it is rendered outside the plot area, but shouldn't be). Easily fixed by using `rc.DrawClippedText(...)` instead of `rc.DrawText(...)`.
2. The rotation angle for the `HalfWayAboveSloped` and `HalfWayBelowSloped` positions is determined based on the `DataPoint`s, making it more or less misaligned with the line depending on transformation. The angle should rather be determined from the transformed `ScreenPoint`s.
3. The implementation probably breaks when using reversed and/or transposed axes. 

Points 1 and 3 are already present with the implementation as it is currently, but are addressed by #1346. So maybe it would make sense to wait with this PR until #1346 is merged.

Source/OxyPlot/Series/LineLegendPosition.cs
SecondShiftEngineer: I've moved the "End" to the third position in the enumerator. Is this the change you had in mind?

SecondShiftEngineer:
I'm realizing there are some issues with how I've chosen to expanded this feature. The most notable is how I'm calculating slope. It's beginning to looks like I may need to make the ability to angle the line legend a separate property rather than combine it with the line legend position. There is also some lack of flexibility in the line legend text justification that should also be addressed.

I'll go ahead and close the pull request while I make the improvements. Thanks for all of the feedback as well as your efforts to continuously develop this project.

PR #1391, opened 2019-10-18
Make PNG exporters consistent

### LINKED ISSUE #1390: Make the PngExporters consistent
The PngExporter for WPF, Winforms and Core.Drawing should be consistent.
I suggest ot use the WinForms exporter as a template:

Include
- a static method to write to a file ("convenience"/one-liner method)
- a method to write to a stream
- a method to write to a platform dependent bitmap 

Do not add additional methods (=confusing)
Do not use native brush as parameter (making your code less portable)
Use `double` for resolution parameter
The implementation should be in the export to stream instance method

objorke:
Fixes #1390.

### Checklist

- Change Core.Drawing and WPF PngExporters to be consistent with WindowsForms
- Remove methods with Brush parameter
- Change resolution parameter to double

@oxyplot/admins
@VisualMelon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ All notable changes to this project will be documented in this file.
 - Overhaul HistogramHelpers (#1345)
 - Remove unused LabelFontSize property from HistogramSeries (#1309)
 - OxyPlot.Windows (UWP) moved to oxyplot-uwp repository (#1378)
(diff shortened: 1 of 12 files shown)


VisualMelon:
We should probably be disposing the brushes in Winforms/Core.Drawing versions, since they aren't managed by the `GraphicsRenderContext` (I really should have thought of this earlier)

PR #1397, opened 2019-10-19
Fix Legend font size is not affected by DefaultFontSize (#1396)

### LINKED ISSUE #1396: Legend font size is not affected by DefaultFontSize
### Steps to reproduce

1. Create a new `PlotModel`
2. Set `LegendFontSize `to `double.NaN`
3. Set `DefaultFontSize `to something noticeable, e.g. 20

Platform: WPF
.NET version: 4.5

### Expected behaviour

The Legend is rendered using the `DefaultFontSize`.

### Actual behaviour

The legend is still rendered with font size = 12 (at least in WPF, could be different for other Renderers).

### Proposed Fix

- Make Legend rendering use the `DefaultFontSize` if the `LegendFontSize` is `double.NaN` (same for `LegendTitleFontSize`).
- Set `LegendFontSize` and `LegendTitleFontSize` to `double.NaN` by default

This can be a breaking change under certain circumstances, but I think most users won't even notice. If we want to avoid the breaking change, we can leave the default values of `LegendFontSize` and `LegendTitleFontSize` at 12 for now and just implement the first point.

If there are no objections I will make a PR later.

Jonarw:
Fixes #1396.

### Checklist

- Use the `DefaultFontSize` for Legend if `LegendFontSize` is not set

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,7 @@ All notable changes to this project will be documented in this file.
 - OxyPlot.Core.Drawing PngExporter background when exporting to stream (#1382)
 - Windows Forms clipping last line of rendered text (#1124, #1385)
 - Dispose background brush in OxyPlot.Core.Drawing PngExporter (#1392)
+- Legend font size is not affected by DefaultFontSize (#1396)
 
 ## [1.0.0] - 2016-09-11
 ### Added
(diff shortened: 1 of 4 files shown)


objorke:
Thanks!

PR #1414, opened 2019-10-27
Check for empty lists in RenderingExtensions (#1410)

### LINKED ISSUE #1410: IndexOutOfRangeException when drawing line / area with no points
### Steps to reproduce

1. View 'Temperatures ver3' example in ExampleBrowser

Platform: WPF
.NET version: 4.5

### Expected behaviour

Example should be shown normally.

### Actual behaviour

`IndexOutOfRangeException` occurs in [`RenderingExtension.ReducePoints(...)`](https://github.com/oxyplot/oxyplot/blob/1190bece79ad52cba740e6f51adee11765e2732b/Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs#L1112).

We should probably add a check for empty points here.

VisualMelon:
Fixes #1410 .

### Checklist

- [ ] I have included examples or tests (ExampleLibrary, TwoColorAreaSeries/Temperature ver3)

- Add checks to `RenderingExtensions.DrawClippedPolygon` and `RenderingExtensions.ReducePoints` to exit early if no points are supplied.

Either addition would have address the `IndexOutOfRange` exception, but both seem reasonable additions given the similar `DrawClippedLine` method. The only behaviour change should be to produce no output and continue when previously their would have been an exception. Based on `DrawClippedLine`, attempting to render nothing should not be treated as exceptional.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ All notable changes to this project will be documented in this file.
 ### Removed
 - Copy to text report Ctrl+Alt+R (#1403)
 
+### Fixed
+- Exception when rendering polygon with no points (#1410)
+
 ## [2.0.0] - 2019-10-19
 ### Added 
 - WindowsForms and Wpf support .NET Core 3.0 (#1331)
--- a/Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs
+++ b/Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs
@@ -350,6 +350,12 @@ namespace OxyPlot
             LineJoin lineJoin = LineJoin.Miter,
             bool aliased = false)
         {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,11 +9,13 @@ All notable changes to this project will be documented in this file.
 ### Changed
 - Moved reporting functionality to separate projects (#1403)
 - Moved reporting functionality to oxyplot-reporting repository (#1403)
-- Moved WPF Plot component to Oxyplot.Contrib.Wpf (#1399)
```
Jonarw: That is probably an accident right?

VisualMelon: Thanks.

objorke:
thanks!

Issue #1081, opened 2017-04-19
Annotation tooltip not displayed


neandersen:
### Steps to reproduce

I upgraded from oxyplot .core/.wpf 1.0.0-unstable2100 to 2.0.0-unstable0968, no other changes.
My annotation tooltips no longer show, but they did with 1.0.0-unstable2100

1. Create plotmodel and add an annotation to it
2. Set the tooltop for the annotation.
3. Mouse over the annotation

Platform: Windows 10 x64 VS 2017 WPF
.NET version: 4.6

### Expected behaviour

Tooltip shown when mouse hovered over annotation.

### Actual behaviour

Nothing happens

silviubogan:
I see the same bug in the ExampleBrowser, in [this example](https://github.com/oxyplot/oxyplot/blob/develop/Source/Examples/ExampleLibrary/Annotations/AnnotationExamples.cs).

If I create a sample application to test this, I see the same bug, although the property Window.ToolTip works well in the same application, and it is not overriden by another tooltip when the mouse is over the annotation.

I use Windows 10 x64 with VS 2019 and the latest preview version of Oxyplot.Wpf from NuGet at the moment of this writing.

silviubogan:
I wish to solve this issue but I am a beginner. I saw that the ToolTip is actually already set on the WPF elements at the end of the OxyPlot's rendering process so I think it has to do with a detail about how WPF renders elements. If I set the ToolTip property of the Plot element, that ToolTip works. I will analyse this more tomorrow.

VisualMelon:
It seems to be a typical transparent background problem. This can be 'fixed' by removing the background on `mouseGrid` which is presumably eating all the mouse-over events (or setting `HitTestVisible = false`)

The `mouseGrid` was added in #1055 in response to #382 (which notes that the change breaks tooltips): https://github.com/oxyplot/oxyplot/blob/2c254dc8352f59b927e7f6600071fcfa5fe764e7/Source/OxyPlot.Wpf/PlotBase.cs#L292

I'm not a WPF guru, so I don't know a way to resolve the two issues off hand.

silviubogan:
[...]

Issue #111, opened 2014-08-14
Handle mouse events on PlotController instead of PlotModel (and elements)


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10132](http://oxyplot.codeplex.com/workitem/10132)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-02-17 08:08
We can achieve the same functionality by using a <code>PlotController</code>

<a href="https://oxyplot.codeplex.com/workitem/9625" rel="nofollow">https://oxyplot.codeplex.com/workitem/9625</a>

<ul>
<li>Better separation of controller logic and model </li><li>Less code </li><li>Better performance of PlotController </li></ul>

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-03-19 22:57
It seems to be quite convenient to use the events. Or is that valid only for the simple demos?

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-03-21 09:02
An alternative could be to add an <code>EnableMouseEventsInModel</code> property on the
<code>PlotController</code>. The default value could be false. When the value is false, the mouse events should not be raised on the
<code>PlotModel</code> and <code>PlotElement</code>s.

**[tibel](http://www.codeplex.com/site/users/view/tibel)** 2014-04-12 08:43
Hmm, still don't like it. The PlotModel and its child elements should not know anything about a mouse, keyboard or touch. Also it is easy to create memory leaks by events on child elements of the PlotModel when your plot is more dynamic (e.g. adding and
 removing annotations).

Therefore I would prefer following:

<ul>
<li>mouse/keyboard events exist only on <code>PlotController</code> and the event will tell you what element of the plot was e.g. clicked
</li><li>similar the <code>AxisChanged</code> should be on <code>PlotModel</code> and tell you what axis changed
</li><li>Ultimately no child element of the plotmodel should expose any events

</li></ul>

With this approach I don't have problems with adding/removing elements from the plotmodel and my event handlers.

tibel:
I like the suggested solution.

malambika:
hello tiebl im new to this xamairn..how to handle mouse evnt in oxyplot using  plotcontroller?

VisualMelon:
[...]

PR #1415, opened 2019-10-27
PREVIEW: Crossplatform tooltip system

objorke:
this is a preview for #1081


Source/OxyPlot.Wpf/PlotBase.cs
```diff
@@ -245,6 +245,11 @@ public void ResetAllAxes()
         /// <param name="updateData">The update Data.</param>
         public void InvalidatePlot(bool updateData = true)
         {
+            if (this.ActualWidth <= 0 || this.ActualHeight <= 0)
```
objorke: I think this should be reverted

Source/OxyPlot/Annotations/ArrowAnnotation.cs
```diff
@@ -183,6 +183,11 @@ public override void Render(IRenderContext rc)
         /// </returns>
         protected override HitTestResult HitTestOverride(HitTestArguments args)
         {
+            if (!this.PlotModel.PlotArea.Contains(args.Point))
```
objorke: move this into calling method?

objorke:
Can we remove https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Rendering/RenderContext/IRenderContext.cs#L187 with these changes?

Source/OxyPlot/Utilities/IToolTipView.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <summary>
+        /// Gets or sets the maximum time between the display of two tooltips where the second tooltip appears without a delay.
+        /// </summary>
+        int BetweenShowDelay { get; set; }
+
+        /// <summary>
+        /// Shows the tooltip.
+        /// </summary>
+        /// <param name="str">The string to show as tooltip.</param>
+        void ShowToolTip(string str);
```
objorke: Wouldn´t it be simpler if this was in https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Graphics/IView.cs instead?

Source/Examples/ExampleLibrary/Misc/ToolTipsExamples.cs
```diff
@@ -0,0 +1,197 @@
...
+                ToolTip = "Test",
+                Text = "Hover the rectangle to see the fixed tooltip.\n" +
+                    "Hover the flashing series to see a new random number each time you hover it.",
+            };
+            plot.Annotations.Add(ra);
+
+            var rnd = new Random();
+            _ = WaitThenChangeColorAndToolTip(plot, rnd, fs);
+
+            plot.MouseDown += (s, e) =>
```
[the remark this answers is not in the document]
silviubogan: @VisualMelon I am not sure how to do this. My first impression is to use the extension method `IController.BindMouseDown` with a parameter of type `DelegateViewCommand` but I do not know how code-gen handles this, and, more important to me, I do not know what `IController` method to use instead of the model's `MouseUp` event.

Thank you.

Source/OxyPlot/Annotations/ArrowAnnotation.cs
silviubogan: @VisualMelon Please, can you tell me how can I access the `(PlotModel.)PlotArea` from the calling method of the code above, [`UIElement.HitTest`](https://github.com/silviubogan/oxyplot/blob/5bc747c1eda59d55953ac70bf1b8a37af87bb29f/Source/OxyPlot/Graphics/UIElement.cs#L69)? Should there be a new `UIElement.PlotModel` property? Thank you.

silviubogan:
@objorke 

> Can we remove https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Rendering/RenderContext/IRenderContext.cs#L187 with these changes?

It is done in a new commit: 192c643.

Thank you.

Source/OxyPlot/Utilities/IToolTipView.cs
silviubogan: @objorke I do not have a strong opinion on this. I think that:

* there are some configuration variables in `IToolTipView` that would pollute `IView` implementations
* loosely coupled objects are good on the long term

Thank you.

Source/OxyPlot/Annotations/ArrowAnnotation.cs
VisualMelon: I'm not sure. `UIElement.PlotModel` would couple `UIElement` to `PlotModel` unnecessarily. I'm not sure if objorke means that the decision to test agains the PlotArea should be handled higher up (though currently it cannot, because `Model` and not `PlotModel` do the work), or if the idea is just to reduce redundancy.

I am not totally convinced that the elements should be filtered at this layer, and I wonder if `Model.HitTest` should give more control to the inheriting class. That's all assuming the change is desirable, which I don't think is entirely clear with annotations.

Source/Examples/ExampleLibrary/Misc/ToolTipsExamples.cs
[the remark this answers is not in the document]
VisualMelon: I'm note sure, I've never used them myself. However, take a look at `ManipulatorBase`: you should be able to bind a `MouseDownGesture`, and implement the started/delta/completed events to handle the entire interaction.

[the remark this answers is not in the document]
silviubogan: @VisualMelon I posted my unsuccessful attempt on this in a gist. The relevant lines with comments are [here](https://gist.github.com/silviubogan/b9eeac18ac4db3ccf0c7b5e534df2f31#file-using-oxyplot-controller-events-in-examples-cs-L62-L64). I think I will try more tomorrow. (I think that I must also find how to use the `OxyMouseDownGesture` class). It would be great if you could help me more. Thank you for all the help.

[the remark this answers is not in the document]
VisualMelon: Manipulators are, I believe, a transient thing. Lots of examples in `PlotCommands.cs` (e.g. [`PanAt'](https://github.com/oxyplot/oxyplot/blob/13c686502ebba28eeca1011ad77eb25ea5788723/Source/OxyPlot/PlotController/PlotCommands.cs#L28). I'd expect the code to be more like this:

    plot.PlotView.ActualController.BindMouseDown(OxyMouseButton.Left, OxyModifierKeys.None, new DelegateViewCommand<OxyMouseDownEventArgs>((v, c, e) =>
    {
        plot.PlotView.ActualController.AddMouseManipulator(plot.PlotView, new CustomMouseManipulator(plot.PlotView, rnd), e);
    }));

However, this can never work, because we just created the model, so it can't have a plotview or controller, so my gut feeling is that answer to @objorke question is that it is not possible. I think I'm going to have to try to get my head around #111...

silviubogan:
@VisualMelon A few unimportant questions, only if you have time for them:

1. What should be the order of the priorities of the tooltip string, the tooltip formatter and the tooltip format string? I would do it like this but I am not sure how good this is: if the tooltip string is available, take it, else if the tooltip format string is available take it, else take the tooltip formatter.

2. What variables should be passed to the format string if they should be? Currently the `HitTestResult` object is passed to the tooltip formatter. I think that in the next week I will take a better look at how this is done for the tracker format string.

3. If it is the case, is it possible to do a rebase onto `upstream/develop` without removing the existing comments on GitHub?

Thank you.

VisualMelon:
@silviubogan 
1. I'd go with (in preferential order) Formatter, Format String, just the String. I think that's what is done elsewhere.
2. I'm not sure. I'd forgotten the tracker was handled per-element, otherwise that would have been my suggestion. I might have another look later when I have more time.
3. Sorry, I don't know. I'd expect they just become 'outdated', but couldn't say for sure.

trovialdo:
I'm wondering if we can just drop the ToolTip string property all together and only provide the formatter function and string. I do not think you would lose any functionality, you can also use the format string to show a static message (or formatter function for that matter).

Be careful if you want to pass more arguments to the formatter function since, if you do it like the tracker (possible custom # arguments per element), it will become the responsibility of the element to generate the tooltip. Hence you might need to add `string GetToolTip(HitTestResult)` on the `element ` API, which i would try to avoid.

(whatever we decide upon we should do the same for the tracker to be consistent in API imo, so probably add a formatter function there as well. But this can be done later on and in another issue.)

silviubogan:
@VisualMelon In your opinion is it good to start working on a global tooltip functionality? What I want to say is  a new property `IPlotView.GlobalToolTip` of type `string` that is shown, if it is not `null`, when the user moves the mouse over a portion of the plot that has no specific tooltip (e.g. a `PlotElement` that has no `ToolTip`(`Formatter`) set). Thank you.

Jonarw:
Closing this according to the discussion in #2012. If there is interest to continue with this topic, feel free to let us know in here or start a new PR.

PR #1416, opened 2019-10-30
#1402 Remove ColumnSeries

### LINKED ISSUE #1402: Remove ColumnSeries
This is kind of a continuation of #1334.
In the discussion for that issue I briefly considered the point that a transposed `BarSeries` could more or less replace a `ColumnSeries`. The main differentiation between these two series seems to be the label placement. This however could be easily resolved by a similar approach as was introduced in #1334 for the `HistogramSeries`: https://github.com/oxyplot/oxyplot/blob/0f703e1ad9464a69c6222b14e4044acb6e38297f/Source/OxyPlot/Series/HistogramSeries.cs#L409

As a proof of concept, I modified `BarSeries` accordingly and tested in in transposed mode:

image
image

So this makes me wonder: Is there a compelling reason to keep the `ColumnSeries` as a separate class when we can get the same plot using a transposed `BarSeries`?

Immediate Benefits would be:

- Less code duplication
- Simpler API
- Derived Series would automatically also be transposable (e.g. `ErrorColumnSeries`)
- Easier switching between horizontal and vertical Bars

The main downside is that this would involve quite some breaking API changes. These are the main changes I can think of right now:

- Remove `BarSeriesBase` and `BarSeriesBase<T>`, move functionality to `BarSeries`
- Remove `BarItemBase`, move functionality to `BarItem`
- Remove `ColumnSeries`, `ColumnItem`
- Rename `ErrorColumnSeries` to `ErrorBarSeries` and make it inherit from `BarSeries`
- Fix `BarSeries` to work in transposed mode (as shown above)
- Remove `ColumnSeries` examples
- Add transposed versions of `BarSeries` examples

What do you think? Did I miss something? Is it worth it?

Jonarw:
Fixes #1402.

### Checklist

- Remove `ColumnSeries` (replaced by transposed `BarSeries`)
- Replace `ErrorColumnSeries` by transposable `ErrorBarSeries`
- Add the ability to display transposed plots to WPF ExampleBrowser
- Fix remaining `XYAxisSeries` to work in transposed mode

@oxyplot/admins


Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
```diff
@@ -10,16 +10,48 @@
...
+
+        /// <summary>
+        /// Returns a value indicating whether a plot model is transposable.
+        /// </summary>
+        /// <param name="model">The plot model.</param>
+        /// <returns>True if the plot model in transposable; false otherwise.</returns>
+        public static bool IsTransposable(this PlotModel model)
+        {                   
+            return (model.Axes.Count > 0 || model.Series.Count > 0)
+                && model.Annotations.Count == 0
```
objorke: could annotations also be transposed?

objorke:
Good job! Thanks for improving this code!

Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
Jonarw: Interesting question! I will open a separate issue for it.

PR #1433, opened 2019-11-21
Add Legends class for extensible Legends

### LINKED ISSUE #644: Make legends open for extension
- Add a property `Legends` to the `PlotModel`. Type `ElementCollection<Legend>` similar to `Axes` and `Series`.
- Move the current legend code into a new `Legend` class
- Refactor the layout (measure/arrange) engine to support multiple legends
- Include some examples

franktore:
Fixes #644

### Checklist

- Add Legends namespace
- Add Legends Base abstract class
- Add Legends default implementation
- Allow Legends to be affected by HitTest
- Allow Legend items to be grouped
- Allow multiple Legends in plot

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,18 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+- Made legends open for extension (#644)
+- Added Oxyplot.Legends namespace
+- Added abstract base class LegendBase
+- Added default implementation Legend and Legend.Rendering
+- Added Legends ElementCollection to PlotModel
+- PlotModel now accommodate multiple Legends, although same positions as before. Legends sharing position are placed on top of eachother
+- Made Legend Items clickable to toggle series visibility
+- Added properties LegendKey and SeriesGroupName to Series, allowing grouping series between multiple legends and/or within same legend
+- Remove PlotModel.Legends
+- Default behaviour is now plot without Legend.
+
 ### Added
 - OxyPlot.ImageSharp (#1188)
 - WPF ExampleBrowser can display transposed versions of examples (#1402)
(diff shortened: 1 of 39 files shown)


Jonarw:
First of all thank you for your contribution!
I only had time for a quick look today, most of it looks pretty good. I just came across two issues:

- Some of the example in WPF example browser look 'weird', e.g. here the 'LegendItemSpacing (only for horizontal orientation)' example:
image
This is probably not intended?

- It would be nice if the new Legend system would be supported by the `CodeGenerator`, so the generated code would also include the legend definition

franktore:
Thanks for the feedback Jonarw. It is pretty quick work and there might indeed be some quirks I've missed. I will look into both issues at first chance.

franktore:
@Jonarw, I've had a look. Adding CodeGenerator support seems straight forward and will be done. The other issue is more difficult. The behaviour you see is an effect of the difference in series title lengths while the LegendItemSpacing is fixed. In order to get the LegendSymbols aligned, as I presume you want, we would have to work with dynamic LegendItemSpacing, which would be possible although complicated since it would require us to keep track of the position of each LegendSymbol on the line above. But, since the legend titles could be anything and vary greatly in length, this could easily lead to very ridiculous looking legends. I feel that the Legend class now gives a great deal of freedom to customize depending on number of series and different titles. The user will have to configure the legend and find the layout that best suits his/her purpose. And should none of the current layouts or renderings be satisfactory it is now possible to write you own legend class based on the LegendBase.

Jonarw:
Oh, I think we are talking about different things, sorry for not being clear enough. In my first point I am not talking about the Legend entries not being aligned, but about the legend extending over the bounds of the plot area. For reference here a screenshot with the result I would have expected:
image

franktore:
Alright, I've made the appropriate corrections.

objorke:
Good job! Is this ready to merge now? I am not able to test all this myself right now, but code looks good. Are examples with multiple legends and grouping included?

franktore:
This is ready to merge. Examples of both multiple legends and grouping are included in examples library.

objorke:
Thank you for the contribution!

PR #1435, opened 2019-11-23
Transposable annotations

### LINKED ISSUE #1417: Add support for transposable Annotations
### Feature description

Now that we have support for transposable `XYAxisSeries`, it would be nice if annotations also supported this. Some initial thoughts:

- To avoid code duplication, we could introduce a `ITransposablePlotElement` interface which `XYAxisSeries` and `Annotation` would both implement. This would expose the `XAxis` and `YAxis`. We could then move the `Transform` and `Orientate` functions to extension methods, something like:
```
public static ScreenPoint Orientate(this ITransposablePlotElement element, ScreenPoint point)
{
    return element.IsTransposed() ? new ScreenPoint(point.Y, point.X) : point;
}
```

- Should `TextRotation` be influenced by transposition? For now I don't think it should be - consistently transposed text would be mirrored and that is hardly what anyone would want. So maybe for text only position and alignment should be affected by transposition?

I will start working on this, if there are any additional considerations, please let me know.

Jonarw:
Fixes #1417 .

### Checklist

- Add Support for transposable Annotations
- To avoid code duplication, introduce interface `ITransposablePlotElement` (implemented by `XYAxisSeries` and `Annotation`)

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 - OxyPlot.ImageSharp (#1188)
 - WPF ExampleBrowser can display transposed versions of examples (#1402)
+- Support for transposable Annotations (#1417)
 
 ### Changed
 - Moved reporting functionality to separate projects (#1403)
--- a/Source/Examples/ExampleLibrary/Annotations/ArrowAnnotationExamples.cs
+++ b/Source/Examples/ExampleLibrary/Annotations/ArrowAnnotationExamples.cs
@@ -6,6 +6,8 @@
 
 namespace ExampleLibrary
 {
+    using System;
+
     using OxyPlot;
     using OxyPlot.Annotations;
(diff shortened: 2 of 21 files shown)


objorke:
Thanks for the contribution! This looks very good!

objorke:
Is it worth explaining how this works in the documentation?

objorke:
I was also thinking if we really need the new interface - could it simply be merged with `IPlotElement`? I am aware of the interface-segregation principle ISP, but also worrying about creating "too small" interfaces - which could be YAGNI.. Keep it simple :-)

Jonarw:
I think there might be some issues with merging the new interface with `IPlotElement`. First of all, just semantically not all `IPlotElement`s are transposable, think of `PieSeries` or polar plots for example. Also not all `IPlotElement`s even have `XAxis` and `YAxis` properties that could implement the interface - for example all Axes classes themselves.

PR #1444, opened 2019-12-10
Remove exporter Background properties (#1409)

### LINKED ISSUE #1409: Most portable exporters ignore `PlotModel.Background`, but SVG does not.
The two portable PNG Exporters and the portable PDF exporter both ignore the `PlotModel` background property, but the portable SVG Exporter does not.

Since it doesn't seem to be the job of the `PlotModel` to render its own background, this should surely be the job of the exporter?

Each of these exporters (except SVG) has its own `Background` property. I would suggest that if this is left undefined, then it should use the `PlotModel` background property. I'll be happy to implement these changes if they are deemed desirable.

VisualMelon:
Fixes #1409 .

This is a breaking change presently and in my opinion warrants further discussion now that it is implement. Hopefully having a graft pull request makes it easy to spot the difference.

 - The WPF exporters have some unique behaviour which is lost in this change, and causes the 'default' background for any exporter to be none because `PlotModel.Background` is `Undefined` by default. This causes the [`OxyPlot.Wpf.Tests.ExampleLibraryTests.ExportPngAndCompareWithBaseline`](https://github.com/oxyplot/oxyplot/blob/4999e591c09464f68bd201fce7f9097e8d12980f/Source/OxyPlot.Wpf.Tests/ExampleLibraryTests.cs#L27) test to fail due to the different background (the CI doesn't notice because it doesn't have an old baseline).

 - The `WindowsForms` `PngExporter` presently defaults to `OxyColors.White`, but now does not.

 - There may be more breaking change outside of the core library (not checked yet).

It makes sense to me that export should be transparent by default, but there was clearly a deliberate decisions to avoid this.

### Checklist

- Remove `Background` properties from `Exporters` and relevant export method overloads
- Removes behaviour of WPF exporters using the `PlotView` background when the `PlotModel` background is invisible.
- Removes behaviour of WinForms exporters using the `OxyColors.White` when the `PlotModel` background is invisible.
- One test is updated to modify `PlotModel.Background`.

@oxyplot/admins


Jonarw:
To me this all looks pretty sensible. I am also not too concerned about the breaking changes, because in the end it contributes to the consistency in the library.

objorke:
I think we can merge this!

PR #1460, opened 2020-02-04
Fix #1455 Custom tracker strings can cause exception in histogram chart

### LINKED ISSUE #1455: Custom tracker strings can cause exception in histogram chart
### Steps to reproduce

1. Create a histogram chart using mapped objects in the ItemsSource
2. Set the TrackerString to use a property of the mapped objects
3. Observe that when mousing over the chart an exception is thrown

Platform:
.NET version:

### Expected behaviour

No exception should be thrown

### Actual behaviour

The problem is in HistogramSeries.GetNearestPoint() 

`return new TrackerHitResult
                        {
                            Series = this,
                            DataPoint = p,
                            Position = point,
                            Item = null,
                            Index = -1,
                            Text = StringHelper.Format(
                            this.ActualCulture,
                            this.TrackerFormatString,
                            null,
                            this.Title,
                            this.XAxis.Title ?? DefaultXAxisTitle,
                            this.XAxis.GetValue(p.X),
                            this.YAxis.Title ?? DefaultYAxisTitle,
                            this.YAxis.GetValue(p.Y),
                            item.RangeStart,
                            item.RangeEnd,
                            item.Value,
                            item.Area,
                            item.Count),
                        };`

The Item = null is wrong, as is the Index = -1 and the StringHelper being passed null.
Both null objects should instead be set to the item from the ItemsSource (or GetItem(index)) and the index should be set. This appears to be merely unfinished code that someone forgot to come back to

Jonarw:
Fixes #1455 .

### Checklist

- Fix the issue
- Remove some unnecessary code from HistogramSeries, most likely copy&paste remainders from initial implementation

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ All notable changes to this project will be documented in this file.
 - Exception when rendering polygon with no points (#1410)
 - ErrorBarSeries, IntervalBarSeries and TornadoBarSeries work correctly in transposed mode (#1402)
 - OxyPlot.WindowsForms package description (#1457)
+- Custom tracker strings can cause exception in histogram chart (#1455)
 
 ## [2.0.0] - 2019-10-19
 ### Added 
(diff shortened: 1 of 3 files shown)


objorke:
thanks!

PR #1475, opened 2020-02-08
Reintroduce support for NET40

### LINKED ISSUE #1452: Again support .NET 4.0 for WinForms and WPF
### Again support .NET 4.0 for WinForms and WPF

As discussed by email, I believe really good libraries should not be limited to the latest bells and whistles, as not all applications can keep up with Microsoft release cycles and support deadlines. .NET 4.0 is the last version of .NET supported on WinXP, and a tool e.g. used on an autonomous machine in a production environment may be tied to the OS version used on that machine. So I ask for maintaining support for .NET 4.0 for WinForms and WPF, even knowing that Microsoft officially ended support for .NET 4.0. (Note that if the Microsoft dates were strictly followed, .NET 3.5 SP1 would still have to be supported.  .NET 3.5 SP1 is officially supported until 2023/2028.)

objorke:
Fixes #1452 .

### Checklist

- Reintroduce support for NET40 for the core library, WindowsForms and WPF

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Foundation/CodeGenerator/CodeGenerator.cs
+++ b/Source/OxyPlot/Foundation/CodeGenerator/CodeGenerator.cs
@@ -137,7 +137,11 @@ namespace OxyPlot
         private string Add(object obj)
         {
             var type = obj.GetType();
+#if NET40
+            var hasParameterLessCtor = type.GetConstructors().Any(c => c.GetParameters().Length == 0);
+#else
             var hasParameterLessCtor = type.GetTypeInfo().DeclaredConstructors.Any(ci => ci.GetParameters().Length == 0);
+#endif
 
             if (!hasParameterLessCtor)
             {
@@ -346,7 +350,12 @@ namespace OxyPlot
             var listsToAdd = new Dictionary<string, IList>();
             var arraysToAdd = new Dictionary<string, Array>();
 
+#if NET40
+            var properties = instanceType.GetProperties(BindingFlags.Public | BindingFlags.Instance);
+#else
             var properties = instanceType.GetRuntimeProperties().Where(pi => pi.GetMethod.IsPublic && !pi.GetMethod.IsStatic);
+#endif
+
             foreach (var pi in properties)
             {
                 // check the [CodeGeneration] attribute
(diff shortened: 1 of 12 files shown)


VisualMelon:
Huh... I had issues with [`IReadOnlyList<T>`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.ireadonlylist-1?view=netframework-4.8) not existing in 4.0 in [my branch](https://github.com/VisualMelon/oxyplot/tree/WinformsNet40Support). I'll submit those changes separately, since there is yet another bug in the histogram helpers that needs addressing.

Source/OxyPlot/Utilities/TypeExtensions.cs
```diff
@@ -26,12 +26,20 @@ public static class TypeExtensions
         /// <returns>An object that represents the specified property, or null if the property is not found.</returns>
         public static PropertyInfo GetRuntimeProperty(this Type type, string name)
         {
+#if NET40
+            var source = type.GetProperties();
```
VisualMelon: I think this should be `var source = type.GetProperties(BindingFlags.NonPublic | BindingFlags.Public | BindingFlags.Static | BindingFlags.Instance);` to be equivalent to `var source = typeInfo.AsType().GetRuntimeProperties();`

Source/OxyPlot/PlotModel/PlotElement.cs
```diff
@@ -147,10 +147,15 @@ protected CultureInfo ActualCulture
         /// <remarks>This method creates the hash code by reflecting the value of all public properties.</remarks>
         public virtual int GetElementHashCode()
         {
+#if NET40
+            var properties = this.GetType().GetProperties(BindingFlags.Public|BindingFlags.Instance);
```
VisualMelon: Formatting: spaces around the `|`?

Source/OxyPlot/Foundation/CodeGenerator/CodeGenerator.cs
```diff
@@ -137,7 +137,11 @@ public string ToCode()
         private string Add(object obj)
         {
             var type = obj.GetType();
+#if NET40
+            var hasParameterLessCtor = type.GetConstructors().Any(c => c.GetParameters().Length == 0);
```
VisualMelon: `DeclaredConstructors` retrieves static constructors, which `Type.GetConstructors` does not (without suitable `BindingFlags`). This code probably shouldn't be picking up static constructors at all, so I think the NET4.0 code is fine, but the NET Standard code needs an additional check so it only counts non-static constructors.

Example of counting static constructors:

     typeof(System.Collections.Generic.List<int>).GetTypeInfo().DeclaredConstructors.Count(ci => ci.GetParameters().Length == 0) // returns 2

objorke: Thanks! It is likely to be other bugs like that...

I think OxyPlot.Tests and ExampleLibrary should also target net40, so we could run all tests on all framework versions...

PR #1477, opened 2020-02-08
Fix bug in HistogramHelpers.Collect when using out-of-order bin breaks (#1476)

### LINKED ISSUE #1476: HitogramHelpers.Collect produces incorrect results when bin breaks are not ordered.
When the supplied bin breaks are not ordered, the `HistogramItem`s produced by `HistogramHelpers.Collect` are incorrect.

The following test demonstrates the incorrect behaviour:

    [Test]
    public void Collect_OutOfOrderBinBreaks()
    {
        // binBreaks should be ordered by Collect 
        var sample = new double[] { 2, 3, 9, 8, 7 };
        var outOfOrderBreaks = new double[] { 0.0, 10.0, 5.0, 15.0 };
        var expectedCounts = new int[] { 2, 3, 0 };
        var expectedAreas = new double[] { 0.4, 0.6, 0.0 };

        foreach (var binningMode in new[] { BinningOutlierMode.CountOutliers, BinningOutlierMode.IgnoreOutliers })
        {
            var binningOptions = new BinningOptions(binningMode, BinningIntervalType.InclusiveLowerBound, BinningExtremeValueMode.ExcludeExtremeValues);
            TestCollect(sample, outOfOrderBreaks, expectedCounts, expectedAreas, binningOptions);
        }
    }

VisualMelon:
Fixes #1476.

### Checklist

- Fix `HistogramHelpers.Collect`, so that the output items are constructed from the internally ordered list of bin breaks
- Add a test for this condition
- Minor fixes to similar tests so that they cover a wider range of inputs as intended

In addition, this commit makes the following change to allow access to the `HistogramHelper` methods from .NET Framework 4.0
- Change the type of `binBreaks` to `IEnumerable<double>` (`IReadOnlyList<T>` is not available in .NET Framework 4.0)
- Change the return type of `Collect` to `List<HistogramItem`> (`IReadOnlyList<T>` is not available in .NET Framework 4.0)
- Remove compiler directories which remove the relevant APIs in .NET Framework 4.0 builds

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ All notable changes to this project will be documented in this file.
 - Support for transposed (X and Y axis switched) plots with XYAxisSeries (#1334)
 - Color property on HistogramItem (#1347)
 - Count property on HistogramSeries (#1347)
+- Bug in HistogramHelpers.Collect when using out-of-order bin breaks (#1476)
 
 ### Changed
 - OxyPlot.Core changed to target netstandard 1.0 and net45 (#946, #1147)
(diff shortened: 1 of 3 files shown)


objorke:
Great!

PR #1484, opened 2020-02-23
Added high DPI support for WPF

### LINKED ISSUE #149: Bad rendering in WPF when scaling is not 100%
**Imported from [http://oxyplot.codeplex.com/workitem/10221](http://oxyplot.codeplex.com/workitem/10221)**

**[ktemelkov](http://www.codeplex.com/site/users/view/ktemelkov)** 2014-06-15 20:58
The plot is not rendered as expected if the display scaling level is not set to 100% (see attached picture)

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-06-15 22:49
I think this should be fixed in the <code>ShapesRenderContext</code>. 

It can probably be solved in a similar way as for the Android render context!

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-07-11 14:08
I committed some TODOs in the ShapesRenderContext class where I think this should be corrected. I have also moved the pixel snapping methods into this class.

Mikant:
Fixes #149 

### Checklist

- [ x] I have included examples or tests
- [ x] I have updated the change log
- [ x] I am listed in the CONTRIBUTORS file
- [ x] I have cleaned up the commit history (use rebase and squash)

Added high DPI support

### DIFF
--- a/Source/OxyPlot.Wpf/Utilities/PixelLayout.cs
+++ b/Source/OxyPlot.Wpf/Utilities/PixelLayout.cs
@@ -1,0 +1,114 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="PixelLayout.cs" company="OxyPlot">
+//   Copyright (c) 2020 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Provides functionality to support non-default dpi scaling
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Wpf
+{
+    using System;
+    using System.Windows;
+    using System.Windows.Media;
+
+    /// <summary>
+    /// Provides functionality to support non-default dpi scaling
+    /// </summary>
+    public static class PixelLayout
+    {
+        /// <summary>
+        /// Determines a visual offset of one <see cref="Visual"/> relative to another
+        /// </summary>
+        /// <param name="visual">The <see cref="Visual"/> which offset is calculated</param>
+        /// <param name="root">The reference <see cref="Visual"/></param>
+        /// <returns>A point structure which contains X and Y offsets</returns>
+        public static Point GetVisualOffset(Visual visual, Visual root)
+        {
+            if (visual == null)
+                throw new ArgumentNullException(nameof(visual));
+
+            if (root == null)
+                return default;
+
+            return visual.TransformToAncestor(root).Transform(default);
+        }
+
+        /// <summary>
+        /// Determines visual dpi scales of a <see cref="Visual"/>
+        /// </summary>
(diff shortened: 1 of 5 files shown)


Source/OxyPlot.Wpf/Utilities/PixelLayout.cs
```diff
@@ -0,0 +1,114 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="PixelLayout.cs" company="OxyPlot">
+//   Copyright (c) 2020 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Provides functionality to support non-default dpi scaling
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Wpf.Utilities
```
objorke: we have not used namespaces like this, but I agree this is a better way

Mikant: Hmm, I was sure I checked it... Should be changed. Is it possible to fix it via github interface (2 file changes (commits?) + squash -> the same pull request)? (sry, 10 yrs of TFS made me somehow resistant to understanding GIT workflow) Or it's simpler close this request and open a new?

Source/OxyPlot/Rendering/RenderContext/IRenderContext.cs
```diff
@@ -186,6 +186,12 @@ public interface IRenderContext
         /// <param name="text">The text in the tool tip, or <c>null</c> if no tool tip should be shown.</param>
         void SetToolTip(string text);
 
+        /// <summary>
+        /// Updates render context state
+        /// </summary>
+        /// <remarks>This method is called at the beginning of each rendering.</remarks>
+        void BeginFrame();
```
Mikant: Despite of trying to avoid anything that could lead to inherent changes, I've added BeginFrame to IRenderContext. I felt almost free to do it because there had already been a method CleanUp which acts the same: it's invoked by the core, once per frame, has five empty implementations... Both of these methods could easily be removed from the interface and be implemented in CanvasRenderContext directly to be invoked from e.g. PlotView. But I'd like to rename CleanUp to EndFrame to make BeginFrame-EndFrame symmetric. As these methods are invoked by the core it would be nice to use them not only in WPF but in e.g. Xamarin or OxyPlot.SharpDX as well to manage platform specific resources. @objorke What do you think? Can anyone grant me permissions for such a breaking change? Or it's better to rollback the interface changes, leave everything as is and implement BeginFrame in WPF's CanvasRenderContext only?

Mikant:
Fixed PixelLayout namespace; removed BeginFrame from IRenderContext. It's not called directly from Wpf.PlotBase.

objorke:
Thanks for the contribution!

PR #1492, opened 2020-03-18
Fix NullReference in VolumeSeries if no data in Items list

### LINKED ISSUE #1491: NullReference in VolumeSeries if no data in Items list
**Steps to reproduce**
Create a new VolumeSeries object and add it to PlotModel.Series
```csharp
var plot = new PlotModel();
var volumeSeries = new VolumeSeries();
plot.Series.Add(volumeSeries);
// arrange Axes, add the plot to a form, etc.
```
Start the application
Platform:
Any

**Expected behaviour**
The plot is empty, only axes are shown 

**Actual behaviour**
NullReferenceException is thrown in VolumeSeries.UpdateData() method until you set VolumeSeries.Items property or add some data to it
oxyPlotVolumeSeriesNullRef

DShaurin:
Fixes #1491 .

### Checklist

- Add a check for item == null in VolumeSeries.GetNearestPoint and VolumeSeries.UpdateData methods

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Series/FinancialSeries/VolumeSeries.cs
+++ b/Source/OxyPlot/Series/FinancialSeries/VolumeSeries.cs
@@ -365,7 +365,7 @@ namespace OxyPlot.Series
         /// <returns>A TrackerHitResult for the current hit.</returns>
         public override TrackerHitResult GetNearestPoint(ScreenPoint point, bool interpolate)
         {
-            if (this.XAxis == null || this.YAxis == null || interpolate || this.data.Count == 0)
+            if (this.XAxis == null || this.YAxis == null || interpolate || this.data == null || this.data.Count == 0)
             {
                 return null;
             }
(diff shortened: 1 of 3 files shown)


Jonarw:
Thanks for your contribution!
Could you please update the changelog and list yourself in the CONTRIBUTORS file? Thanks!

Source/OxyPlot/Series/FinancialSeries/VolumeSeries.cs
```diff
@@ -423,6 +423,11 @@ protected internal override void UpdateData()
             base.UpdateData();
             this.winIndex = 0;
 
+            if (this.data == null || this.data.Count == 0)
```
Jonarw: Are we changing behaviour here in case `data != null` and `data.Count == 0`? Previously it that case line 447 would have been run...

DShaurin: Yes, the UpdateData behavior is changed for this case, but it will have no effect on the plot since minDx is a gap size between existing bars. But, we have no bars at all if data.Count is 0, so minDx is not used.

Jonarw:
Thanks!

PR #1500, opened 2020-03-28
Remove WpfExamples dependency on OxyPlot.Pdf (#1499)

### LINKED ISSUE #1499: WpfExamples references to a project with net452 when itself is netcoreapp3.1
### Steps to reproduce

1. Call 
```cmd
dotnet build .\Source\Examples\WPF\WpfExamples\WpfExamples.csproj
```
2. See following
`C:\Program Files\dotnet\sdk\3.1.201\Microsoft.Common.CurrentVersion.targets(1655,5): warning NU1702: ProjectReference '.\oxyplot\Source\OxyPlot.Pdf\OxyPlot.Pdf.csproj' was resolved using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v3.1'. This project may not be fully compatible with your project. [.\oxyplot\Source\Examples\WPF\WpfExamples\WpfExamples.csproj]`

### Expected behaviour

Build without warning NU1702

### Actual behaviour

Build with warning NU1702

VisualMelon:
Fixes #1499

### Checklist

- Remove useless and unused project reference in `WpfExamples` to `Oxyplot.Pdf` which was causing nuget to complain (#1499)

The dependency was unused (uses the Pdf support in `Oxyplot.Core` instead), and could not be used anyway, because the examples are compiled for .NET Core 3.1, but `OxyPlot.Pdf` only targets .NET Framework because PdfSharp only support .NET Framework (though there are efforts to port it to .NET Standard, e.g. https://github.com/ststeiger/PdfSharpCore and https://github.com/Didstopia/PDFSharp)

@oxyplot/admins

### DIFF
--- a/Source/Examples/WPF/WpfExamples/WpfExamples.csproj
+++ b/Source/Examples/WPF/WpfExamples/WpfExamples.csproj
@@ -6,7 +6,6 @@
       <OutputType>WinExe</OutputType>
   </PropertyGroup>
   <ItemGroup>
-    <ProjectReference Include="..\..\..\OxyPlot.Pdf\OxyPlot.Pdf.csproj" />
     <ProjectReference Include="..\..\..\OxyPlot.Wpf\OxyPlot.Wpf.csproj" />
     <ProjectReference Include="..\..\..\OxyPlot\OxyPlot.csproj" />
   </ItemGroup>


rstm-sf:
It seems to me that you can delete an unnecessary file: https://github.com/oxyplot/oxyplot/blob/27712b73571f9aa5606550004e1fc4ee70031ed4/Source/Examples/WPF/WpfExamples/app.config#L1-L3

VisualMelon:
The above change was slipped in t0 #1501. Is this ready to merge, or would an entry in the changelog be useful?

objorke:
This is just an example, I think there is no need to update the changelog for this change.

objorke:
Thanks for fixing this!

PR #1506, opened 2020-04-17
Fix wrong index from FindWindowStartIndex

avl:
Fix an error where FindWindowStartIndex in OxyPlot.Series.XYAxisSeries would return the wrong value.

This error made it so that parts of the graph were sometimes not drawn, since the index returned from FindWindowStartIndex could be too large and therefore cause the drawing of the graph to start in the middle of the window instead of at the first visible part.

I'm not a developer for Oxyplot, but thought I'd make this pull request anyway, since it's a very non-controversial fix.

### DIFF
--- a/Source/OxyPlot/Series/XYAxisSeries.cs
+++ b/Source/OxyPlot/Series/XYAxisSeries.cs
@@ -786,6 +786,7 @@ namespace OxyPlot.Series
                     {
                         return end;
                     }
+                    lastguess = curGuess;
                 }
                 else
                 {


avl:
Oh wait, this fix doesn't work! I'll be back!

VisualMelon:
@avl It would be useful if you could provide a reproducible example, so that the maintainers can verify the problem and solution.

avl:
I'm closing this, since it doesn't work. I'll be back with a solution which works.

PR #1508, opened 2020-04-17
Add more color palettes (#1505)

### LINKED ISSUE #1505: Add more color palettes
As the Jet color palette has become somewhat of a meme in the scientific community, I think it would be nice to have some alternatives built into oxyplot. matplotlib has some nice colormaps specifically designed for data visualization, which are under public domain. Specifically I think the following colormaps would be nice to integrate:

- viridis
- plasma
- magma
- inferno
- cividis

Sources for the RGB data of these color maps:
- https://github.com/BIDS/colormap/blob/master/colormaps.py
- https://github.com/pnnl/cmaputil/blob/master/colormaps/cividis.txt

Further reading:
- https://matplotlib.org/3.1.0/tutorials/colors/colormaps.html
- https://bids.github.io/colormap/
- https://www.youtube.com/watch?v=xAoljeRJ3lU
- https://arxiv.org/ftp/arxiv/papers/1712/1712.01662.pdf

Jonarw:
Fixes #1505.

### Checklist

- Add color maps Viridis, Plasma, Magma, Inferno and Cividis from matplotlib
- Change default color map of `LinearColorAxis` to Viridis

@oxyplot/admins


VisualMelon:
The existing palettes allow the user to specify the number of colors in the map. It isn't hard to create a new palette interpolated from an existing palette, but would it be useful to provide a consistent API?

Jonarw:
You do have a valid point, however I am not really sure how we could make this consistent because of the way these new color palettes are defined.

For palettes like `Jet` or `Hot` it is easy to define them by a small amount of colors and interpolate between them. On the other hand, palettes like `Viridis` cannot easily be determined by interpolation, because they follow fairly complex paths through color space - that's why they are usually defined by a 'baked-in' sequence of colors. In most cases, the best way to use the palette would be to just straight-up take it without any interpolation, which I hoped would be kind of implied by this choice of API.

VisualMelon:
Sorry, yes: I didn't mean interpolation for the purpose of having more colors (there are already duplicates, so it certainly won't help), but sampling (of the complex path) to have fewer.

It should still be the case that a reduced-count palette should provide a representative mapping ([matplotlib](https://matplotlib.org/3.1.3/tutorials/colors/colormap-manipulation.html#getting-colormaps-and-accessing-their-values) provides this provision through a consistent API). Just for example, lots of the work I do requires 'banding' for the purpose of interpretation.

There is also a boring concern that rendering the color-map involves plotting the all the colors (this could be changed, but would require knowledge about the render), which has overheads. 200 colors is arguably excessive for any purpose, and could result in rendering deficiencies depending on how sub-pixel rectangles are processed (not that I've observed this).

Jonarw:
Ok, you convinced me.
I think a good compromise would be to offer the same API as for the existing palettes, but with a default value for `numberOfColors` which gives the full palette without subsampling / interpolation.

UndisputedBasement33:
In my opinion it would be great if you could define the alpha value of the palette also. When creating for example heatmaps, I create my own palettes with transparency, because otherwise the gridlines are not visible. So I do something like that:
```

            List<OxyColor> jetTransparent = new List<OxyColor>();
            foreach (var item in OxyPalettes.Jet(1000).Colors)
            {
                jetTransparent.Add(OxyColor.FromAColor(220, item));
            }
            OxyPalette myPalette = new OxyPalette(jetTransparent);
```

This is a workaround, but it would be easier if you could do something like this: OxyPalettes.Jet(1000, 200), so that the first number is the amount of colors and the second number is the alpha value.

Jonarw:
I think a better approach would be to have an extension method like

```
public static OxyPalette WithAlpha(this OxyPalette palette, byte alpha)
{
...
}
```

Then you could write

```
var palette = OxyPalettes.Jet(1000).WithAlpha(220);
```
But I think this is its own topic. Feel free to open a separate issue about this.

PR #1512, opened 2020-04-19
Fix bug where FindWindowStartIndex did not correctly handle NaN-values

### LINKED ISSUE #1507: AreaSeries sometimes render incorrectly (with fix!)
Because there is an error in method FindWindowStartIndex in file XYAxisSeries.cs, on line 788, AreaSeries are sometimes not rendered correctly. Rather, some polygons which should be part of the series are never drawn.

There is a very simple fix:

https://github.com/oxyplot/oxyplot/pull/1506

avl:
Fixes #1507.


avl:
I think we ideally should have a fuzz-test that verifies that this function works.

VisualMelon:
@avl The Example Library has an `Issues` section containing example plots which expose issues: feel free to put together a 'fuzz-test' as a demonstration.

Jonarw:
Yes, please add an example that shows what is broken without your fix.
There are some issues with code style in your PR. We are not super-religious about this, but some things we keep consistent (braces around single-line ifs and loops and spacing around operators for example). If you could take another look at this it would be great :).

Also, could `FindWindowStartIndex(...)` be rewritten to use `List.BinarySearch(...)`?

Source/OxyPlot/Series/XYAxisSeries.cs
```diff
@@ -750,62 +750,65 @@ protected int UpdateWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter,
         /// <param name="targetX">target x.</param>
         /// <param name="initialGuess">initial guess index.</param>
         /// <returns>
-        /// index of x with max(x) &lt;= target x or -1 if cannot find
+        /// index of x with max(x) &lt;= target x or 0 if cannot find
         /// </returns>
-        protected int FindWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter, double targetX, int initialGuess)
+        public int FindWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter, double targetX, int initialGuess)
```
avl, line 755: Making it public to be able to call it from tests. Is there a better way?

avl: > Yes, please add an example that shows what is broken without your fix.
> There are some issues with code style in your PR. We are not super-religious about this, but some things we keep consistent (braces around single-line ifs and loops and spacing around operators for example). If you could take another look at this it would be great :).
> 
> Also, could `FindWindowStartIndex(...)` be rewritten to use `List.BinarySearch(...)`?

Sorry about the code style issues! I agree style should be consistent - my apologies!

I don't think List.BinarySearch would be a good fit, since we need to somehow handle the NaN-values. They can't be sorted on their own, we need to sort by their neighbors. But the IComparer in List.BinarySearch doesn't have access to the index, so it can't find the neighbors.

avl:
> @avl The Example Library has an `Issues` section containing example plots which expose issues: feel free to put together a 'fuzz-test' as a demonstration.

Unfortunately this really isn't a good candidate for such a test. Because the "initial guess" value has such a high impact on the the result, the error may appear to occur intermittently. In fact it is deterministic, but slightly chaotic.

On the other hand, the expected result from FindWindowStartIndex can be mathematically stated. I think my test probably does a better job. In fact, my fuzz test exposes the NaN-bug in the original implementation.

avl:
> Yes, please add an example that shows what is broken without your fix.
> There are some issues with code style in your PR. We are not super-religious about this, but some things we keep consistent (braces around single-line ifs and loops and spacing around operators for example). If you could take another look at this it would be great :).
> 
> Also, could `FindWindowStartIndex(...)` be rewritten to use `List.BinarySearch(...)`?

Sorry about the code style issues! I agree style should be consistent - my apologies!

I don't think List.BinarySearch would be a good fit, since we need to somehow handle the NaN-values. They can't be sorted on their own, we need to sort by their neighbors. But the IComparer in List.BinarySearch doesn't have access to the index, so it can't find the neighbors.

Source/OxyPlot/Series/XYAxisSeries.cs
```diff
@@ -750,62 +750,65 @@ protected int UpdateWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter,
...
+                        return guessX;
                 }
+                return xgetter(items[nominalEnd]);
+            }
 
-                double guessX = xgetter(items[curGuess]);
+            while (start < end)
+            {
+                double guessX = GetX(curGuess);
                 if (guessX.Equals(targetX))
```
VisualMelon, line 780: (Not an issue with this PR, but may be relevant to the issue). This logic is inconsistent with the assumption of monotonicity: it is possible for a previous `X` value to also equal `targetX`.

VisualMelon:
I'm somewhat dubious of the `FindWindowStartIndex` method and it's usage altogether. It strikes me as unnecessarily complicated. It's presumably very efficient in the case where the data is monotonic and evenly spaced, but a simpler binary search would guarentee logarithmic performance on any data set (e.g. with logarithmically distributed X values).

There is an edge-case where the data contains duplicate x values. `FindWindowStartIndex` is only employed when the dataset is monotonic (i.e. every x value is no less than a previous x values), but essentially assumes that every x value is greater than any previous x value. The documentation suggests that if there are duplicate X values, then it will return the last (highest index) of them, but this is not enforced. Since it used (exclusively?) to find the start of a window, I would suggest changing the contract and behaviour so it finds either the first entry with the given X value, or the last entry with an X value less than that.

Source/OxyPlot/Series/XYAxisSeries.cs
VisualMelon: This is the sort of thing I'd be tempted to move out of the class into Utilities somewhere as a `public static` method. There is no reason this logic could be used by e.g. annotations, one-dimensional series, pre-labelled axes, etc.. I'd be tempted to move everything to do with monotonicity optimisations into a new class.

Source/OxyPlot/Series/XYAxisSeries.cs
```diff
@@ -750,62 +750,65 @@ protected int UpdateWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter,
         /// <param name="targetX">target x.</param>
         /// <param name="initialGuess">initial guess index.</param>
         /// <returns>
-        /// index of x with max(x) &lt;= target x or -1 if cannot find
+        /// index of x with max(x) &lt;= target x or 0 if cannot find
```
VisualMelon, line 753: The change to `0` appears to accurately describe the new and old behaviour. A simple test for this would be nice.

In principle I'd prefer it returned -1, but existing code seems to depend on this behaviour.

Source/OxyPlot/Series/XYAxisSeries.cs
```diff
@@ -750,62 +750,65 @@ protected int UpdateWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter,
...
-            int end = items.Count - 1;
-            int curGuess = initialGuess;
+            int nominalEnd = items.Count - 1;
+            while (nominalEnd > 0 && double.IsNaN(xgetter(items[nominalEnd])))
+                nominalEnd -= 1;
+            int end = nominalEnd;
+            int curGuess = Math.Max(0, Math.Min(end, initialGuess));
 
-            while (start <= end)
+            double GetX(int index)
```
VisualMelon, line 764: Presently the method can return indexes to `NaN` values, which is inconsistent with the documentation, because this pretends the item at the current index is something that it is not. Returning the 'true' index might be a sufficient solution; however, this value could not be used for updating the `end`, only the start or returning the index when found.

Source/OxyPlot/Series/XYAxisSeries.cs
```diff
@@ -750,62 +750,65 @@ protected int UpdateWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter,
...
 
                 var m = (end - start + 1) / (endX - startX);
+                
                 curGuess = start + (int)((targetX - startX) * m);
+                curGuess = Math.Max(start + 1, Math.Min(curGuess, end));
             }
 
-            return lastguess;
+            while (start > 0 && (xgetter(items[start]) > targetX))
+                start -= 1;
```
VisualMelon, line 809: This condition is never fired in the tests. Is it necessary? Is it liable to obscure bugs in the main mechanism?

Source/OxyPlot/Series/XYAxisSeries.cs
VisualMelon: Not an issue with this PR, but the summary text should explain that the method assumes the non-NaN items are ordered.

Source/OxyPlot/Series/XYAxisSeries.cs
avl: Yeah, I think that equals-comparison is quite suspect. My thought was that it might be some kind of optimization - perhaps we end up there if the function is called with a 'correct' initial guess?

I think the idea of monotonicity in oxyplot is a graph where the Y-value is a function of the X-value?

In that case, what we really have is strict monotonicity - it shouldn't be allowed to have two different values with the same X-value, since the curve wouldn't correspond to a function of the X-axis in that case.

Then of course we have parametric graphs where both X and Y depend on an independent variable, in that case the X-axis values wouldn't necessarily have any monotonicity.

My proposal would be to change the interpretation of Monotonicity in Oxyplot and replace it with strict monotonicity.

Source/OxyPlot/Series/XYAxisSeries.cs
```diff
@@ -750,62 +750,65 @@ protected int UpdateWindowStartIndex<T>(IList<T> items, Func<T, double> xgetter,
...
-                }
-                else if (curGuess > end)
-                {
-                    return end;
+                    double guessX = xgetter(items[index]);
+                    if (double.IsNaN(guessX))
+                        index += 1;
+                    else
+                        return guessX;
                 }
```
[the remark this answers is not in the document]
avl: I think more than one NaN in a row would be rather unexpected. But yeah, I agree. This could be optimized. 

My 'fix' is just to consider the NaN values as if they had the value of a non-NaN neighbor. There are definitely more efficient ways to handle this, by making the algorithm in itself aware of NaN-values.

Source/OxyPlot/Series/XYAxisSeries.cs
avl: Yeah, I agree! I just wanted to keep my PR as minimal as possible, since it's meant as a bug fix. 

But I also noted that the function could/should have been a static method in some other class :-)

Source/OxyPlot/Series/XYAxisSeries.cs
avl: I thought about that. 

If the use case is that we want to know which is the first index we need to consider in order to draw the graph correctly, we could make the function return the first index that needs to be drawn. The could well be an index with X-value larger than the targetX.

An alternative is to define the function to return the largest non-NaN value with x < targetX.

I think I prefer the latter alternative. What do you think?

Source/OxyPlot/Series/XYAxisSeries.cs
avl: Sorry, I threw this together for fixing an semi-urgent issue in my day-job. I didn't have to time to convince myself that the code would always give the correct value so added that as a sort of fail safe.

It's a horrible kludge! :-) 

It should be cleaned up. Did you run a code coverage tool? And the test didn't trigger it? I be it could be removed in that case.

Source/OxyPlot/Series/XYAxisSeries.cs
VisualMelon: Yeah, that latter idea is how I would read the documentation (i.e. find the index with largest x value that is no more than the target, which arguably can't be a NaN entry).

Source/OxyPlot/Series/XYAxisSeries.cs
VisualMelon: I couldn't convince myself it was needed, and I wanted to check it wasn't always being over-used (e.g. something was wrong with binary search and it was just falling back on a linear search) so I replaced the insides with a `throw` and all the tests still passed.

VisualMelon:
Would anyone object to merging this at this point? I have complaints with the use of a non-trivial binary search, but that isn't an issue with this PR. The changes seem to solve the problem, and add additional tests. The test failures don't look like they are caused by this PR. @avl could you rebase and repush? That might gives us some more green ticks.

avl:
> Would anyone object to merging this at this point? I have complaints with the use of a non-trivial binary search, but that isn't an issue with this PR. The changes seem to solve the problem, and add additional tests. The test failures don't look like they are caused by this PR. @avl could you rebase and repush? That might gives us some more green ticks.

I've rebased upon latest commit in branch "develop".

Unfortunately I don't have access to a working development environment for Oxyplot right now, so I'm afraid I can't really do much work on this at the moment. 

I'll be happy to answer questions though.

Jonarw:
To be honest, I'm not 100% sure which problem this solves. I apologize if I overlooked this, but is there an example somewhere that shows an `AreaSeries` which is rendered wrongly before the changes and correctly after?

avl:
> To be honest, I'm not 100% sure which problem this solves. I apologize if I overlooked this, but is there an example somewhere that shows an `AreaSeries` which is rendered wrongly before the changes and correctly after?

Because the binary search in FindWindowStartIndex does not support NaN-values, it will basically fail if it ever encounters such a value. NaN-values are used to signal 'breaks' in series in Oxyplot.

When drawing any series containing NaN-values, the drawing may not start at the correct location on the screen. This means that large parts of the series may end up not being drawn. Panning will lead to new 'guesses' of start location, which may sometimes not hit a NaN-value, which means the series will partially flicker in and out of existence while panning/zooming. 

This PR fixes the problem.

I'm sorry I don't have the time to produce a working reproduction. I initially experienced the problem while drawing an AreaSeries, but I'm not certain it only affects AreaSeries.

avl:
I think this should be merged. Ok, I'm biased as the author, but it is a rather important fix.

It may be that there is nothing in the examples that demonstrates the problem, but the PR contains a test case which actually fails for the old code. The old code is simply broken, and this PR fixes it.

VisualMelon:
@avl if you could make an example to go in the Issues section of the example library that shows the issue (and fix), that would be very helpful. I assume that any line/area series with enough NaNs in it should present the issue easily enough. If you don't beat me to it, I'll probably (try to) put one together in a few hours time.

VisualMelon:
There seem to be plenty of issues with `AreaSeries` and NaN values to go around, but I can't seem to reproduce this one now. @avl can I throw this back to you? As jonarw says, we need to be sure this solves the problem, and I'm clutching at straws now ;)

avl:
@VisualMelon 
Ok, I found some time to look att this. It wasn't easy to reproduce in an example, but I finally did find a case where an areagraph was drawn wrong in an example I made (one which could be displayed in example browser).

````csharp
        [Example("Broken series")]
        public static PlotModel BrokenSeries()
        {
            var plotModel1 = new PlotModel { Title = "AreaSeries broken in time" };
            var axis = new LinearAxis {Position = AxisPosition.Left, MinimumPadding = 0, MaximumPadding = 0.06, AbsoluteMinimum = 0};
            var xAxis = new LinearAxis() {Position = AxisPosition.Bottom};
            
            plotModel1.Axes.Add(axis);
            plotModel1.Axes.Add(xAxis);

            var N = 15;
            var random = new Random(6);
            var currentValue = random.NextDouble() - 0.5;
            var areaSeries = new AreaSeries();
            for (int i = 0; i < N; ++i)
            {
                if (random.Next(4) == 0)
                {
                    areaSeries.Points.Add(DataPoint.Undefined);
                    areaSeries.Points2.Add(DataPoint.Undefined);
                }
                    
                currentValue += random.NextDouble();
                areaSeries.Points.Add(new DataPoint(currentValue, currentValue));
                areaSeries.Points2.Add(new DataPoint(currentValue, currentValue));
            }
            
            plotModel1.Series.Add(areaSeries);
          

            return plotModel1;
        }
````

The above series can be made to be drawn wrong by positioning the view exactly like this:

image

(note the barely visible data series in the lower hand left part)



And then using the keyboard to pan the graph to the right. It is possible to end up with a badly drawn graph, like this:

image

Just panning slightly back and forth restores the correct drawing:
image


Note, with this example it is actually quite tricky to reproduce. The initial position needs to be such that the last part of the graph is just barely drawn in its entirety (although linethickness makes it so that it appears clipped). With my customer's data, it happened all the time. It is a large dataset though, and one I don't think I can share.

I'm not sure there is much point in actually adding this example to the examplebrowser, since it's so tricky to reproduce the error.

**Comment by avl** [comment]
The above test data is data found by the fuzz-test that I actually included in the PR. So I'm pretty convinced the fuzz-test works and is valuable.

**Comment by VisualMelon** [comment]
@avl Thanks!

I have reproduced the problem without the fix, and it _not_ going wrong with the fix. I was actually able to reproduce it very easily with your example by just moving the plot with the cursor keys without zooming (slap left 4 times in either winforms or wpf example browser). Alternatively, setting the x-axis minimum to 4 (heh) means it open with the error present.

Thanks for taking the time to put that together.

**Comment by avl** [comment]
Are there any remaining hurdles preventing this from being merged at this time?

**Comment by VisualMelon** [comment]
@Jonarw did you want to take a look at this? I'll merge it over the weekend if there are no objections (feel free to beat me to it).

**Comment by Jonarw** [comment]
Oh I am sorry, I completely forgot to check back here.
Thanks for the example @avl !

Could you please:

- Add this example to the [issues section](https://github.com/oxyplot/oxyplot/blob/develop/Source/Examples/ExampleLibrary/Issues/Issues.cs) of the `ExampleLibrary`
- Leave a small note in the [changelog](https://github.com/oxyplot/oxyplot/blob/develop/CHANGELOG.md) about this fix
- Add yourself to the [list of contributors](https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTORS)

Then I'll be happy to merge this. Thanks!

**Comment by avl** [comment]
I have now added the example to the issues section in the example library, written a note in the changelog, and added myself to the list of contributors.

File: Source/Examples/ExampleLibrary/Issues/Issues.cs [hunk 7]
```diff
@@ -2283,6 +2283,38 @@ public static PlotModel EmojiText()
 
             return plot;
         }
+        [Example("#1512: FindWindowStartIndex.")]
+        public static PlotModel FindWindowsStartIndex()
+        {
+            var plotModel1 = new PlotModel { Title = "AreaSeries broken in time" };
+            var axis = new LinearAxis {Position = AxisPosition.Left, MinimumPadding = 0, MaximumPadding = 0.06, AbsoluteMinimum = 0};
+            var xAxis = new LinearAxis() {Position = AxisPosition.Bottom};
```
VisualMelon, line 2291: Could this have `Minimum = 4` set, so that the error is presented immediately? (Or otherwise, a subtitle explaining how to provoke the issue). In the future it won't necessarily be easy for people to find this PR and read the discussion relevant to the example.

PR #1514, opened 2020-04-21
Add renderer based on SkiaSharp (#1509)

### LINKED ISSUE #1509: Add renderer based on SkiaSharp
### Why would this be awesome?

- Performance: 
  - SkiaSharp offers GPU-accelerated rendering on [many platforms](https://github.com/mono/SkiaSharp/wiki/Views-Support-Matrix) (sadly not WPF at the moment, though this may [change](https://github.com/mono/SkiaSharp/issues/745#issuecomment-581581567))
  - Even the CPU-based Skia renderer seems to be faster on WPF than `CanvasRenderContext`

- Portability
  - SkiaSharp itself targets `netstandard1.3`
  - The same `SkiaRenderContext` could be used on all platforms where a SkiaSharp `View` [is available](https://github.com/mono/SkiaSharp/wiki/Views-Support-Matrix)
  - This would make the implementation of OxyPlot controls on new platforms easier, as you wouldn't have to implement the `IRenderContext` for each platform individually
  - All SkiaSharp-based platforms would produce similar rendering output, because they would be based on the same `IRenderContext` implementation

- Export functionality
  - SkiaSharp supports PNG and PDF export (including Unicode) without any further dependencies - so this would be a nice way of getting 'proper' portable PDF export

### Current State
I have a somewhat crude albeit working version of oxyplot with SkiaSharp in this branch: https://github.com/Jonarw/oxyplot/tree/skia

This still needs some more work and testing, but allows for viewing the examples as well as PDF and PNG export.

The central question for me is:
Would be want to have this in the main repository? Or should this be a separate repository like OxyPlot.Avalonia and OxyPlot.SharpDX?

Jonarw:
Fixes #1509.

### Checklist

- Add `SkiaRenderContext`, implementing `IRenderContext` using SkiaSharp
- Add PNG, PDF, SVG and JPEG exporters based on `SkiaRenderContext`

See also my notes in #1509.

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.SkiaSharp.Tests/OxyPlot.SkiaSharp.Tests.csproj
+++ b/Source/OxyPlot.SkiaSharp.Tests/OxyPlot.SkiaSharp.Tests.csproj
@@ -1,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+    <PropertyGroup>
+        <TargetFrameworks>net461;netcoreapp3.1</TargetFrameworks>
+        <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
+        <Description>OxyPlot SkiaSharp unit tests</Description>
+        <PackageLicenseExpression>MIT</PackageLicenseExpression>
+        <Copyright>OxyPlot contributors</Copyright>
+        <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+        <Platforms>AnyCPU;x86;x64</Platforms>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(TargetFramework)|$(Platform)'=='net461|AnyCPU'">
+        <PlatformTarget>x86</PlatformTarget>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(TargetFramework)|$(Platform)'=='net461|x86'">
+        <PlatformTarget>x86</PlatformTarget>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(TargetFramework)|$(Platform)'=='net461|x64'">
+        <PlatformTarget>x64</PlatformTarget>
+    </PropertyGroup>
+    <ItemGroup>
+        <ProjectReference Include="..\Examples\ExampleLibrary\ExampleLibrary.csproj" />
+        <ProjectReference Include="..\OxyPlot.SkiaSharp\OxyPlot.SkiaSharp.csproj" />
+        <ProjectReference Include="..\OxyPlot\OxyPlot.csproj" />
+    </ItemGroup>
+    <ItemGroup>
+        <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.6.0" />
+        <PackageReference Include="NUnit" Version="3.12.0" />
+        <PackageReference Include="NUnit3TestAdapter" Version="3.15.1" />
+    </ItemGroup>
+</Project>
--- a/Source/OxyPlot.SkiaSharp/SkiaExtensions.cs
+++ b/Source/OxyPlot.SkiaSharp/SkiaExtensions.cs
@@ -1,0 +1,36 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="SkiaExtensions.cs" company="OxyPlot">
+//   Copyright (c) 2020 OxyPlot contributors
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.SkiaSharp
+{
+    using global::SkiaSharp;
+
+    /// <summary>
+    /// Provides extension methods for conversion between SkiaSharp and oxyplot objects.
+    /// </summary>
+    public static class SkiaExtensions
+    {
+        /// <summary>
+        /// Converts a <see cref="OxyColor"/> to a <see cref="SKColor"/>;
+        /// </summary>
+        /// <param name="color">The <see cref="OxyColor"/>.</param>
+        /// <returns>The <see cref="SKColor"/>.</returns>
+        public static OxyColor ToOxyColor(this SKColor color)
+        {
+            return OxyColor.FromArgb(color.Alpha, color.Red, color.Green, color.Blue);
+        }
+
+        /// <summary>
+        /// Converts a <see cref="SKColor"/> to a <see cref="OxyColor"/>;
+        /// </summary>
+        /// <param name="color">The <see cref="SKColor"/>.</param>
+        /// <returns>The <see cref="OxyColor"/>.</returns>
+        public static SKColor ToSKColor(this OxyColor color)
+        {
+            return new SKColor(color.R, color.G, color.B, color.A);
+        }
+    }
+}
--- a/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
+++ b/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
@@ -1,0 +1,861 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="SkiaRenderContext.cs" company="OxyPlot">
+//   Copyright (c) 2020 OxyPlot contributors
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.SkiaSharp
+{
+    using global::SkiaSharp;
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// Implements <see cref="IRenderContext" /> based on SkiaSharp.
+    /// </summary>
+    public class SkiaRenderContext : IRenderContext, IDisposable
+    {
+        private readonly Dictionary<FontDescriptor, SKTypeface> typefaceCache = new Dictionary<FontDescriptor, SKTypeface>();
+        private SKPaint paint = new SKPaint();
+        private SKPath path = new SKPath();
+
+        /// <summary>
+        /// Gets or sets the DPI scaling factor. A value of 1 corresponds to 96 DPI (dots per inch).
+        /// </summary>
+        public float DpiScale { get; set; } = 1;
+
+        /// <inheritdoc/>
+        public bool RendersToScreen { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets the <see cref="SKCanvas"/> the <see cref="SkiaRenderContext"/> renders to. This must be set before any draw calls.
+        /// </summary>
+        public SKCanvas SkCanvas { get; set; }
+
+        /// <inheritdoc/>
+        public void CleanUp()
+        {
+        }
+
+        /// <inheritdoc />
+        public void Dispose()
+        {
+            this.Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        /// <inheritdoc/>
+        public void DrawEllipse(OxyRect extents, OxyColor fill, OxyColor stroke, double thickness, EdgeRenderingMode edgeRenderingMode)
+        {
+            if (!fill.IsVisible() && !(stroke.IsVisible() || thickness <= 0))
+            {
+                return;
+            }
+
+            var actualRect = this.Convert(extents);
+
+            if (fill.IsVisible())
+            {
+                var paint = this.GetFillPaint(fill, edgeRenderingMode);
+                this.SkCanvas.DrawOval(actualRect, paint);
+            }
+
+            if (stroke.IsVisible() && thickness > 0)
+            {
+                var paint = this.GetStrokePaint(stroke, thickness, edgeRenderingMode);
+                this.SkCanvas.DrawOval(actualRect, paint);
+            }
+        }
+
+        /// <inheritdoc/>
+        public void DrawEllipses(IList<OxyRect> extents, OxyColor fill, OxyColor stroke, double thickness, EdgeRenderingMode edgeRenderingMode)
+        {
+            if (!fill.IsVisible() && (!stroke.IsVisible() || thickness <= 0))
+            {
+                return;
+            }
+
+            var path = this.GetPath();
+            foreach (var extent in extents)
+            {
+                path.AddOval(this.Convert(extent));
+            }
+
+            if (fill.IsVisible())
+            {
+                var paint = this.GetFillPaint(fill, edgeRenderingMode);
+                this.SkCanvas.DrawPath(path, paint);
+            }
+
+            if (stroke.IsVisible() && thickness > 0)
+            {
+                var paint = this.GetStrokePaint(stroke, thickness, edgeRenderingMode);
+                this.SkCanvas.DrawPath(path, paint);
+            }
+        }
+
+        /// <inheritdoc/>
+        public void DrawImage(
+            OxyImage source,
+            double srcX,
+            double srcY,
+            double srcWidth,
+            double srcHeight,
+            double destX,
+            double destY,
+            double destWidth,
+            double destHeight,
+            double opacity,
+            bool interpolate)
+        {
+            if (source == null)
+            {
+                return;
+            }
+
+            var bytes = source.GetData();
+            var image = SKBitmap.Decode(bytes);
+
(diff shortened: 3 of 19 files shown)


Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -0,0 +1,762 @@
...
+            return scaledThickness;
+        }
+
+        /// <summary>
+        /// Gets a <see cref="SKPaint"/> containing information needed to render the fill of a shape.
+        /// </summary>
+        /// <param name="fillColor">The fill color.</param>
+        /// <param name="edgeRenderingMode">The edge rendering mode.</param>
+        /// <returns>The paint.</returns>
+        private SKPaint GetFillPaint(OxyColor fillColor, EdgeRenderingMode edgeRenderingMode)
```
VisualMelon, line 636: Could these methods have a little documentation explaining that they modify the local instance and then return the same object? They are private, and it's obvious from the code, but they look otherwise like any piece of interop code, and someone making sweeping changes to multiple render contexts might not realise this slightly unique behaviour.

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -0,0 +1,762 @@
...
+        {
+            this.paint.Color = fillColor.ToSKColor();
+            this.paint.Style = SKPaintStyle.Fill;
+            this.paint.IsAntialias = this.ShouldUseAntiAliasing(edgeRenderingMode);
+            return this.paint;
```
VisualMelon, line 641: It doesn't appear to make any difference in testing, and I can't work out from the documentation if `PathEffect` ever applies to Fills, but should the `PathEffect` be cleared here?

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -0,0 +1,762 @@
...
+        /// <summary>
+        /// Gets a <see cref="SKPaint"/> containing information needed to render text.
+        /// </summary>
+        /// <param name="fontFamily">The font family.</param>
+        /// <param name="fontSize">The font size.</param>
+        /// <param name="fontWeight">The font weight.</param>
+        /// <returns>The paint.</returns>
+        private SKPaint GetTextPaint(string fontFamily, double fontSize, double fontWeight)
+        {
+            var typeface = SKTypeface.FromFamilyName(fontFamily, new SKFontStyle((int)fontWeight, (int)SKFontStyleWidth.Normal, SKFontStyleSlant.Upright));
```
VisualMelon, line 740: Should these be disposed?

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -0,0 +1,762 @@
...
+            {
+                var paint = this.GetStrokePaint(stroke, thickness, edgeRenderingMode);
+                this.SkCanvas.DrawOval(actualRect, paint);
+            }
+        }
+
+        /// <inheritdoc/>
+        public void DrawEllipses(IList<OxyRect> extents, OxyColor fill, OxyColor stroke, double thickness, EdgeRenderingMode edgeRenderingMode)
+        {
+            if (!fill.IsVisible() && !stroke.IsVisible())
```
VisualMelon, line 69: Include `thickness > 0` check?

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -0,0 +1,762 @@
...
+            this.paint?.Dispose();
+            this.path?.Dispose();
+            this.paint = null;
+            this.path = null;
+        }
+
+        /// <inheritdoc/>
+        public void DrawEllipse(OxyRect extents, OxyColor fill, OxyColor stroke, double thickness, EdgeRenderingMode edgeRenderingMode)
+        {
+            if (!fill.IsVisible() && !(stroke.IsVisible() || thickness == 0))
```
VisualMelon, line 48: Should the thickness check be `<= 0`? Negative numbers don't make sense, but the later checks explicitly exclude them. Such checks also appear in e.g. `GraphicsRenderContext` (though that looks like a bit of mess in this regard).

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -0,0 +1,762 @@
...
+        private SKPath path = new SKPath();
+        private float dpiScale = 1;
+
+        /// <inheritdoc/>
+        public bool RendersToScreen { get; set; } = true;
+
+        /// <summary>
+        /// Gets or sets the <see cref="SKCanvas"/> the <see cref="SkiaRenderContext"/> renders to. This must be set before any draw calls.
+        /// </summary>
+        public SKCanvas SkCanvas { get; set; }
```
VisualMelon, line 29: Could this be set in a constructor to improve discoverability and minimise the opportunity to forget setting it? (Other renders accept a `null` canvas for the purposes of text measuring, which I'd want to be explicitly acknowledged as in `Oxyplot.Wpf`.

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -0,0 +1,762 @@
...
+    using System.Linq;
+
+    /// <summary>
+    /// Implements <see cref="IRenderContext" /> based on SkiaSharp.
+    /// </summary>
+    public class SkiaRenderContext : IRenderContext, IDisposable
+    {
+        private SKPaint paint = new SKPaint();
+        private SKPath path = new SKPath();
+        private float dpiScale = 1;
```
VisualMelon, line 21: Is there a reason this should be private and set through a method rather than just an auto property?

Source/OxyPlot.SkiaSharp/SkiaExtensions.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// Converts a <see cref="SKColor"/> to a <see cref="OxyColor"/>;
+        /// </summary>
+        /// <param name="color">The <see cref="SKColor"/>.</param>
+        /// <returns>The <see cref="OxyColor"/>.</returns>
+        public static SKColor ToSKColor(this OxyColor color)
+        {
+            if (color.IsAutomatic() || color.IsUndefined())
+            {
+                return SKColors.Empty;
+            }
```
VisualMelon, line 36: Would it be better to preserve the exact values so that these methods can round-trip?

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
Jonarw: Setting this via the constructor would be the obvious choice when we are only considering exporters.
With potential `PlotView` implementations however it's a slightly different story: we don't have the `SKCanvas` at the time of creating the `SkiaRenderContext`, rather we get it passed from the `OnRender` (or similar) event - potentially a different `SKCanvas` object every time. So in any case we need the setter.
We could initialize this from the constructor, but allow `null` to be passed.

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
Jonarw: Not really... You are right, probably an auto property would be neater.

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
VisualMelon: > We could initialize this from the constructor, but allow null to be passed.

That's what `CanvasRenderContext` and `GraphicsRenderContext` do, but if you reckon the setter is necessary than I'd be happy leaving it as is.

Source/OxyPlot.SkiaSharp/SkiaExtensions.cs
Jonarw: My reasoning behind this was that SkiaSharp does not know that the ARGB [0,0,0,1] value is somehow 'special' to us.
But thinking about it, it doesn't really matter... For rendering it should make no difference and the round-trip argument is valid.

Jonarw:
Thank you for your suggestions @VisualMelon! I will take a closer look tomorrow.

Jonarw:
The failed unit tests on Azure Pipelines could be related to the unit tests being run in 'Any CPU' configuration, this may conflict with SkiaSharp's native dll file. Running the tests in x64 should do the trick (as the `OxyPlot.SkiaSharp.Tests` project is compiled as x64), but I am not too sure where I could change that.

VisualMelon:
Could setting the target platform to x64 unconditionally help? I get the same exceptions when I set it to x86. Presumably the tests are run under Release rather than Debug, but I can't seem to find this information.

Jonarw:
I am not even sure how exactly the tests are run in x86 mode, considering `OxyPlot.SkiaSharp.Tests` is explicitly compiled in x64.

That the tests fail on the other hand is no surprise to me: Because `OxyPlot.SkiaSharp.Tests` is compiled in x64 mode, the 'nuget magic' places the native x64 libskiasharp.dll in the build folder, which can't be used by a x86 process. I will try and see what happens if I explicitly set the test platform to x64...

Jonarw:
OK it was totally me being dumb, I did set the `OxyPlot.SkiaSharp.Tests` project to x64, but only for the `Debug` configuration and not for `Release`.

Jonarw:
I agree it makes sense running the tests on both net461 and netcoreapp3.1.
I don't see the immediate benefit of supporting both x86 and x64 on the test project. But I guess if nothing else at least it documents how to set up a project using OxyPlot.SkiaSharp that supports both platforms, so that's also worth something.

Source/OxyPlot.SkiaSharp.Tests/OxyPlot.SkiaSharp.Tests.csproj
```diff
@@ -0,0 +1,30 @@
...
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(Platform)'=='AnyCPU'">
+        <PlatformTarget>x86</PlatformTarget>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(Platform)'=='x86'">
+        <PlatformTarget>x86</PlatformTarget>
+    </PropertyGroup>
+    <PropertyGroup Condition="'$(Platform)'=='x64'">
+        <PlatformTarget>x64</PlatformTarget>
+    </PropertyGroup>
```
VisualMelon, line 19: For reasons I do not understand, I had to change this further to

    <PropertyGroup Condition="'$(TargetFramework)|$(Platform)'=='net461|AnyCPU'">
        <PlatformTarget>x86</PlatformTarget>
    </PropertyGroup>
    <PropertyGroup Condition="'$(TargetFramework)|$(Platform)'=='net461|x86'">
        <PlatformTarget>x86</PlatformTarget>
    </PropertyGroup>
    <PropertyGroup Condition="'$(TargetFramework)|$(Platform)'=='net461|x64'">
        <PlatformTarget>x64</PlatformTarget>
    </PropertyGroup>

in order to get the tests to run on my WSL Ubuntu instance (i.e. don't set PlatformTarget under .NET Core; doesn't seem to be a problem on my windows box).

Jonarw: What was the issue otherwise? Did you get a `BadImageFormatException` or something else?

VisualMelon: Not really sure, never really needed to understand NUnit. Here's the [output in a gist](https://gist.github.com/VisualMelon/3ae85794600a8167c40c3ccf1d4b9774).

It reports a `BadImageFormatException` on the Tests file itself: the error seems rather uncharacteristic to be honest, and I'm not sure what is going on; I was under the impression that `PlatformTarget` was ignored by net core.

VisualMelon: Update: It works when `PlatformTarget` is set to x64.

VisualMelon: Sorry for all the confusion. I've been using .NET Core for years without realising that you can target platforms without publishing. The default is AnyCPU, and I believe there is no reason to have otherwise here.

Now I just need to work out why the x86 version targeting NetCore 3.1 runs on my Windows box when I don't have an x86 version of NetCore 3.1...

Jonarw: I'm similarly baffled as you that I was oblivious to this behavior of .NET Core until now. I'm still not completely sure I fully understand what is going on, but I think it is like this:

- .NET Core ignores the `PlatformTarget` and creates a cross-platform app that can run on all supported architectures and operating systems
- Depending on the environment the app is executed as either x86 or x64
- The test host is deployed as either x86 or x64 (but not both of them)
- When there is a mismatch between the test host and the (runtime) app architecture you get the `BadImageFormatException`

I think for the specific case of the Test project, the most sensible way is to force one specific architecture on both frameworks, for example x64. So we can be sure the x64 test host is deployed and has a x64 app to work with. To do that I think we should specify the `RuntimeIdentifier` for netcoreapp3.1. So we end up with the project file as in this gist: 
https://gist.github.com/Jonarw/f98dde40a50dbd8a30faf3dd9b11b85b

Do you agree? Did you come to the same conclusions as me? Or did I misunderstand something?

VisualMelon: It is irritating how little documentation there is about this. Based on my observations, my new understanding is this:
 - `PlatformTarget` is not ignored by .NET Core, and defaults to `AnyCPU`. Seems to have the same behaviour with .NET Framework
 - For `AnyCPU`, this means the runtime is determined by the host: if you run it will NET Core x64, it will run as x64; x86 for an x86 version of NET Core. In this case, it is able to pick the right native DLLs because it ships both
 - My Linux run failed because the current csproj was default to x86, because I wasn't asking for anything else. It worked when I unconditionally set `PlatformTarget` to `AnyCPU` or `x64`, because the x64 host can deal with either of these.
 - My Windows runs worked because I had an x86 .NET Core 3.1 host installed, a fact I didn't realise at the time. (I thought it was working because I thought `PlatformTarget` was ignored under core)

Because the image type is determined by the host, I think AnyCPU would be the best default for .NET Core.

Seeing as all the CIs are Windows, I don't think it's the biggest issue in the world (and I think this should be merged as soon as possible; this is the sort of thing that won't influence the design which we can resolve later). I would, however, prefer that it wasn't unconditionally be x64 given x86 should (and does) work: which is used for CI doesn't seem to matter. I'd also prefer the that the NET Core config wasn't windows specific in any case.

By luck more than anything else, I think my most recent suggestion would be the best option, as it allows the platform to be qualified under framework, defaulting to x86 which is the better-supported image format. It doesn't touch the config for .NET Core, defaulting to AnyCPU (which should work anywhere), and not interfering with anything that might be specified by a user. Changing the `AnyCPU` override to `x64` would equally be fine.

But again, these are just tests, so maybe I'm wasting my effort worrying about it. I've certainly learned a thing or two in the last couple of days... or perhaps you can tell me what I've got wrong ;)

Jonarw: > It is irritating how little documentation there is about this.
Yes, I was thinking that as well...

What you say sounds plausible, and I think is also consistent with my observations.

> But again, these are just tests, so maybe I'm wasting my effort worrying about it. I've certainly learned a thing or two in the last couple of days...

Yes indeed, me as well.

Jonarw:
From my side this would be ready to merge I think.
Any more comments from your side @objorke?

Jonarw:
Added a small fix related to dash arrays (SkiaSharp uses pixels, not multiples of stroke thickness).

Issue #1393, opened 2019-10-18
Rename OxyPlot project to OxyPlot.Core


objorke:
To be consistent with the `OxyPlot.Core` NuGet package.
Still inconsistent with namespace `OxyPlot`

This is not a breaking change for SDK style projects using PackageReference.
Breaking change for old style projects.

objorke:
do it for v2.0?

VisualMelon:
I would support Option A - and don't see any issue with the package/project not directly matching the namespace - but I think the benefit would be negligible. I'm not sure what you mean by it not being a breaking change: every SDK project (I don't think there are any non-SDK projects remaining) will need to be updated, though I think it's a one-click job in VS via the OxyPlot.CI solution. It shouldn't have any impact on users.

I don't think option B is worth the breaking change, and I would oppose the renaming on principle, since the package is a subset of functionality.

objorke:
Added alternative C

safecorndiet:
Hello everyone, I am interested in participating in open-source projects. Please inform me if there is any way I can help.

aiyervenkat:
Option A sounds like the safest path to achieving consistency.

sachPundir:
> Added alternative C

Still, Option C is best. I can do this if you still need a contributor. Please let me know, and I will start on it.

PR #1523, opened 2020-04-29
Rename OxyPlot to OxyPlot.Core (#1393)

VisualMelon:
Addresses #1393, but the more I look at it the less I like it.

This was not as trivial as I expected, because I'd neglected to perform the directory renaming in my first experiments.

### Checklist

 - Rename the `OxyPlot` project to `OxyPlot.Core`, and rename its directory to `OxyPlot.Core`
 - Make relevant updates to other 

### Outstanding concerns

- I've not added a changelog notice because this shouldn't affect users, but I feel I probably should add something...
- Renaming the `OxyPlot` directory is a big change, but it looks like git handles it well. Does anyone know of any issues with renaming? Will this invalidate endless references to code on GitHub?
- `OxyPlot.Core` sounds like `OxyPlot.Core.Drawing`: in one, `Core` means core component, in the other `Core` means .NET Core.
 - By default, VS will create new files with a `OxyPlot.Core` namespace, which is easy to resolve, but just an added annoyance
 - Should `OxyPlot.snk` be renamed?

### Summary

I'm not convinced this is a good idea. I would reject this and close the issue: it's not a problem for users, the devs can cope, and without renaming the namespaces (which would be a nightmare) will make it harder for people unfamiliar with the library to identify the right project file by namespace alone.

@oxyplot/admins


objorke:
yes, maybe this is not a good idea. Agree that this can be confused with .NET Core. I am adding option C to the issue.

PR #1533, opened 2020-05-01
Add WPF PlotView based on SkiaSharp (#1515)

### LINKED ISSUE #1515: Add WPF PlotView based on SkiaSharp
I already started working on this and have a basic working implementation. Stuff I still need to figure out (suggestions are welcome):

- Find a way to share code between `OxyPlot.WPF` and `OxyPlot.SkiaSharp.WPF`
  - Stuff like Tracker etc. would be duplicated otherwise
  - Introduce `OxyPlot.WPF.Shared` project?
- Find a way to have an example browser for `OxyPlot.SkiaSharp.WPF`
  - Integrate into WPF `ExampleBrowser` and offer a way to switch between Canvas and SkiaSharp renderer?
  - This would have the nice side effect to have a direct visual comparison between both renderers

Jonarw:
Fixes #1515.

### Checklist

- Add PlotView based on `SkiaRenderContext`
- To enable code sharing between `OxyPlot.Wpf` and `OxyPlot.SkiaSharp.Wpf`, introduce `OxyPlot.Wpf.Shared`. Most of the code from `OxyPlot.Wpf` ended up in here.
- Update WPF example browser so it supports both Canvas and SkiaSharp renderers to allow direct optical comparison between the two. 
  - Enabled DPI-aware mode for example browser, so it is possible to test the effect of different DPI settings without restart
  - Removed fps measurement from example browser (for now), as it would need some rework to also work with the SkiaSharp renderer

### Notes

- Getting pixel-perfect drawing of the bitmap rendered by SkiaSharp to the screen proved somewhat difficult. I got it to work on all configurations I can test, would be happy to hear if it works for other people as well!
- Direct optical comparison revealed some minor inconsistencies between Canvas and SkiaSharp renderers, I will address those in separate issues
- Should the new assemblies (`OxyPlot.SkiaSharp`, `OxyPlot.SkiaSharp.Wpf`, `OxyPlot.Wpf.Shared`) be signed? @objorke do you want to create the .snk files for those?

@oxyplot/admins


objorke:
I think we should be consistent with signing. 
Can we reuse the same .snk for all those libraries?

Jonarw:
Ah I wasn't aware that you could use the same .snk for multiple assemblies. As far as I can tell so far every assembly has its own .snk file.

So if we share the .snk between some of the projects, should we share it between all of them to be consistent?

For now I will just copy the .snk files to sign each of the new projects, but that might be something to be considered in a separate issue.

VisualMelon:
This seems to introduce a bug into the Wpf example browser, whereby it will crash if you:

1. Switch to the code tab
2. Open an example that is not already open

Partial stack-trace:

```
System.InvalidOperationException: The specified Visual is not an ancestor of this Visual.
   at System.Windows.Media.Visual.TrySimpleTransformToAncestor(Visual ancestor, Boolean inverse, GeneralTransform& generalTransform, Matrix& simpleTransform)
   at System.Windows.Media.Visual.TransformToAncestor(Visual ancestor)
   at OxyPlot.Wpf.PlotView.UpdateDpi() in C:\Users\Murdo\source\repos\oxyplot\Source\OxyPlot.Wpf\PlotView.cs:line 113
   at OxyPlot.Wpf.PlotViewBase.RenderOverride() in C:\Users\Murdo\source\repos\oxyplot\Source\OxyPlot.Wpf.Shared\PlotViewBase.cs:line 381
   at OxyPlot.Wpf.PlotView.RenderOverride() in C:\Users\Murdo\source\repos\oxyplot\Source\OxyPlot.Wpf\PlotView.cs:line 94
   at OxyPlot.Wpf.PlotViewBase.Render() in C:\Users\Murdo\source\repos\oxyplot\Source\OxyPlot.Wpf.Shared\PlotViewBase.cs:line 373
```

[_More stack-trace_](https://gist.github.com/VisualMelon/605af123468ecf384cfca9a97f2ea561)

This appears to be new as of 458de6a9c2f92b1e932dc4d72ee3ea1bd07a25eb

Everything seems to work fine if I just wrap all of `PlotViewBase.RenderOverride()` in a `try...catch { }`, but I assume there is a better way of detecting that we shouldn't render because we aren't in the spatial tree.

VisualMelon:
> Removed fps measurement from example browser (for now), as it would need some rework to also work with the SkiaSharp renderer

Could this be included, but only enabled for the Canvas renderer?

Source/OxyPlot.Wpf.Shared/PlotViewBase.Events.cs
```diff
@@ -1,10 +1,7 @@
 ﻿// --------------------------------------------------------------------------------------------------------------------
-// <copyright file="PlotBase.Events.cs" company="OxyPlot">
-//   Copyright (c) 2014 OxyPlot contributors
+// <copyright file="PlotView.Events.cs" company="OxyPlot">
```
VisualMelon, line 2: Should be `PlotViewBase.Events.cs`

Source/OxyPlot.Wpf.Shared/PlotViewBase.Properties.cs
```diff
@@ -1,10 +1,7 @@
 ﻿// --------------------------------------------------------------------------------------------------------------------
-// <copyright file="PlotBase.Properties.cs" company="OxyPlot">
-//   Copyright (c) 2014 OxyPlot contributors
+// <copyright file="PlotView.Properties.cs" company="OxyPlot">
```
VisualMelon, line 2: `PlotViewBase`

Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
```diff
@@ -0,0 +1,434 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="PlotView.cs" company="OxyPlot">
```
VisualMelon, line 2: `PlotViewBase`

Source/OxyPlot.Wpf/PlotView.cs
```diff
@@ -1,5 +1,5 @@
 ﻿// --------------------------------------------------------------------------------------------------------------------
-// <copyright file="PlotView.cs" company="OxyPlot">
+// <copyright file="PlotBase.cs" company="OxyPlot">
```
VisualMelon, line 2: `PlotView`

Jonarw:
> This seems to introduce a bug into the Wpf example browser

Good catch! I've addressed it and will upload a fix and test soon.

> Could this be included, but only enabled for the Canvas renderer?

In principle yes, but I did not found a quick way of getting it to work consistently.
Also I did not find that fps measurement terribly useful before, as it shows the combined update+render time... It would be nice if we in the future could have a measurement that works for both Canvas and SkiaSharp and would display separate update and render times, but this would be out of the scope of this PR.

Jonarw:
Squashed the commits.
From my side this would be ready to merge. Any other concerns?

objorke:
Nice! I am looking forward to try this out!

Issue #1530, opened 2020-05-01
Change OxyPlot.ImageSharp to target netstandard1.3


Jonarw:
See discussion in #1518.
ImageSharp itself targets netstandard1.3, so it should be possible.

Optional:

- Implement `DrawImage(...)`
- DPI-Awareness

VisualMelon:
While we are at it, should we rename the `PngRenderContext` to `ImageRenderContext`, and add overloads for the various formats supported by ImageSharp? (e.g. png, jpg, gif). Additional exporters could be added for the corresponding formats, or we could find some other way of specifying the format (e.g. an enum, filename expection).

Jonarw:
Propably separate exporters for each export formats would be preferable, as there might be some additional settings relevant for each format (e.g. compression factor for JPEG).

Issue #1531, opened 2020-05-01
Improve vertical text alignment of OxyPlot.SvgExporter


Jonarw:
See discussion in #1518.
Output of SkiaSharp SVG export could give some inspiration.

VisualMelon:
It looks like the Skia svg exporter doesn't touch the `dominant-baseline` alignment, which is the svg parameter that has little browser support. It 'defaults' to baseline, which is a problem, because the Svg exporter has no idea where this is (`MeasureText` only tells it the height). This means the best we can do is to 'guess' where the baseline is (e.g. ~1/4 of the way up) unless we extend some interfaces.

An implementation of this 'just sort of guess where the baseline might be' approach is implemented in 679023676a0247b3e98cbd5da28ba78f48a27d71 and produces results like [these](https://gist.github.com/VisualMelon/946e6b3d5c71e90dbd1b5b28ea562fb0
) when using ImageSharp for text measurement.

![image](https://user-images.githubusercontent.com/5646475/81206281-c84b6380-8fc3-11ea-8311-4534bd32edc8.png)

![image](https://user-images.githubusercontent.com/5646475/81206170-a0f49680-8fc3-11ea-80b9-aa7d039be4d4.png)

That commit also does a few other things:
 - Adds an ImageSharp based Svg exporter
 - Changes everything to use the SVG vertical text workaround by default, since support for dominant-baseline is extremely limited.
 - Improves the multi-line text alignment for SVGs
 - Adds a multi-line aligned text example from #1538 
 - Changes multi-line support in various places to split on the regex "\r?\n" rather than "\r\n"

I'll wait until the nightmare inducing ImageSharp changes have been pulled/rejected/whatever before I add a PR for any of this, and will probably break it up into separate PRs (assuming the descent kludge is acceptable at least for the time being).

While I'm typing, I noticed that the PortableDocumentFont stuff doesn't add gaps between lines (which explains some of the deficiencies in the SvgRenderContext I lazily didn't fix last time round).

Jonarw:
> It 'defaults' to baseline, which is a problem, because the Svg exporter has no idea where this is (`MeasureText` only tells it the height). This means the best we can do is to 'guess' where the baseline is (e.g. ~1/4 of the way up) unless we extend some interfaces.

Could we improve this by overriding `DrawText` in an e.g. ImageSharp-based `SvgRenderContext` implementation? There we have access to more sophisticated text measurements and should be able to properly align the text using the baseline only, right?

[...]

PR #1534, opened 2020-05-03
Overhaul OxyPlot.ImageSharp (#1530)

VisualMelon:
Addresses #1530 

### Checklist

- [x] I have cleaned up the commit history

- Change OxyPlot.ImageSharp to target .NET Standard 1.3, bringing it inline with the base library
- Add `DrawImage` support to `OxyPlot.ImageSharp.PngRenderContext`
- Add Dpi awareness to `OxyPlot.ImageSharp.PngRenderContext` and `OxyPlot.ImageSharp.PngExporter`
- Add rotated text support to `Oxyplot.ImageSharp`
- Makes `Oxyplot.ImageSharp` font sizes conform to 96DPI standard
- Add clipping support to `Oxyplot.ImageSharp`
- Add dashed line support to `Oxyplot.ImageSharp`
- (Code breaking) Add a constructor to `OxyPlot.ImageSharp.PngExporter`, meaning there is no longer a default constructor
- Add basic tests for the `PngExporter` copied from `OxyPlot.SkiaSharp.Tests`
- Add a `JpegExporter` along with tests

Dpi awareness is based directly on the implementation for `OxyPlot.SkiaSharp`, and as such would appreciate a review by @Jonarw. There is a lot of shared logic, but type discrepancy mean there is relatively little shared code, and none of it is very interesting; it may be worth providing helper methods (within `RenderContextBase`?) for Dpi awareness to reduce bloat and ensure consistency between platforms going forward. The change undoes a hack to do with a now result issue with dpi aware text rendering in ImageSharp.

I am putting together [some notes](https://gist.github.com/VisualMelon/d1eb4bca4f52de86353421f96bcaeb58) on the implementation to explain why it is the way it is, so that when someone looks at it, proclaims "That's hideous!", and decides to rewrite it, they will hopefully know more going into it then I did.

There are a couple of TODOs left, but they are small things that won't have sweeping design consequences.

@oxyplot/admins


VisualMelon:
I can't see why snyk is failing, so I'm not sure how I can address that at this time.

objorke:
I am testing the snyk tool, was not aware that it was added as a required check here. Can remove it if it blocks the PR!

VisualMelon:
@objorke do you have any idea why Snyk is unhappy? It seems I don't have authorisation to view the report.

objorke:
Snyk says "Expected target framework netstandard2.0 is not present in proj file" with link to https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot.ImageSharp/OxyPlot.ImageSharp.csproj
To me this looks like a false positive error report.

objorke:
I have deactivated the Snyk webhook. I guess we need a change here to trig the checks again.

VisualMelon:
@objorke currently rewriting `DrawImage`, so should hopefully have a change soon...

VisualMelon:
The `DrawImage` method should now approximate the `SkiaSharp` renderer. I've added additional tests to both in order to verify this.

There is significantly complexity in the `DrawImage` method in order to provide sub-pixel resolution and edge-clamped sampling. I'm steadily coming to the conclusion it would be better to implement a more advanced `DrawImage` method in `ImageSharp` itself for us to use, rather than try to work around the existing API; however, this seems to work well enough at the moment.

Once someone has sanity checked the changes, I'll probably squash everything into one commit, because it's all relevant to #1530, though the additional SkiaSharp tests are pushing that a bit.

Source/OxyPlot.ImageSharp/OxyPlot.ImageSharp.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFramework>netstandard2.0</TargetFramework>
+        <TargetFramework>netstandard1.3</TargetFramework>
```
Jonarw, line 3: Should we add `<LangVersion>8</LangVersion>`? Then we could use those tasty switch expressions ;)

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
```diff
@@ -75,42 +87,286 @@ public void Save(Stream output)
...
+            if (edgeRenderingMode == EdgeRenderingMode.PreferSharpness)
+            {
+                scaledThickness = Snap(scaledThickness, 0);
+            }
+
+            return scaledThickness;
+        }
+
         /// <inheritdoc/>
         public override void DrawText(ScreenPoint p, string text, OxyColor fill, string fontFamily = null, double fontSize = 10, double fontWeight = 400, double rotation = 0, OxyPlot.HorizontalAlignment horizontalAlignment = OxyPlot.HorizontalAlignment.Left, OxyPlot.VerticalAlignment verticalAlignment = OxyPlot.VerticalAlignment.Top, OxySize? maxSize = null)
```
Jonarw, line 320: It would be nice to support text rotation, but from a quick search ImageSharp does not seem to support that... Or is there a way?

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
```diff
@@ -119,27 +375,154 @@ public override OxySize MeasureText(string text, string fontFamily = null, doubl
...
-                PointF[] mappedPoints = points.Select(p => new PointF((float)p.X, (float)p.Y)).ToArray();
+                var actualThickness = this.GetActualThickness(thickness, edgeRenderingMode);
+                var actualDashArray = dashArray != null
+                    ? this.ConvertDashArray(dashArray, actualThickness)
+                    : null;
+
+                var pen = actualDashArray != null
+                    ? new Pen(ToRgba32(stroke), actualThickness, actualDashArray)
+                    : new Pen(ToRgba32(stroke), actualThickness);
+                var actualPoints = this.GetActualPoints(points, thickness, edgeRenderingMode).ToArray();
```
Jonarw, line 522: I think you have to pass the `actualThickness` here.

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
```diff
@@ -148,14 +531,21 @@ public override void DrawPolygon(IList<ScreenPoint> points, OxyColor fill, OxyCo
...
+                var actualThickness = this.GetActualThickness(thickness, edgeRenderingMode);
+                var actualDashArray = dashArray != null
+                    ? this.ConvertDashArray(dashArray, actualThickness)
+                    : null;
 
-                var brush = Brushes.Solid(new Rgba32(fill.R, fill.G, fill.B, fill.A));
+                var pen = actualDashArray != null
+                    ? new Pen(ToRgba32(stroke), actualThickness, actualDashArray)
+                    : new Pen(ToRgba32(stroke), actualThickness);
+                var actualPoints = this.GetActualPoints(points, thickness, edgeRenderingMode).ToArray();
```
Jonarw, line 542: Also `actualThickness`.

Source/OxyPlot.ImageSharp/OxyPlot.ImageSharp.csproj
VisualMelon: Personally I deplore switch expressions, but scoped using is nice yo have ;)

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
VisualMelon: I've done rotated text with ImageSharp before, but it was only mutliples of 90 degrees. I'll dig out that old code and see what I did there - because I feel like I was using transforms in a way that my recent experience suggests don't exist - and get back to you.

VisualMelon:
@Jonarw yeah, just scaling and snapping is what I had in mind. I won't do that as part of this, because we may want to incorporate it into other exporters, and this is PR is already getting out of hand.

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
```diff
@@ -75,42 +87,286 @@ public void Save(Stream output)
...
+            {
+                case EdgeRenderingMode.PreferGeometricAccuracy:
+                case EdgeRenderingMode.PreferSpeed:
+                    return rects.Select(this.Convert);
+                default:
+                    var actualThickness = this.GetActualThickness(strokeThickness, edgeRenderingMode);
+                    var snapOffset = GetSnapOffset(actualThickness);
+                    return rects.Select(rect => this.ConvertSnap(rect, snapOffset));
+            }
+        }
```
Jonarw, line 300: These appear to be unused, so can be removed.

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
VisualMelon: Actually, `GetSnapOffset` inside `GetActualPoints` performs the transform itself. This is consistent with `SkiaSharp`

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
VisualMelon: (as above)

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
VisualMelon: Should probably override `DrawRectangles` at some point, but for now I will just remove these.

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
Jonarw: Oh ok sorry, my mistake then.

Source/OxyPlot.ImageSharp/PngRenderingContext.cs
VisualMelon: Nope: that old code is just rotating the whole image. I can't shake the feeling that I must be missing something about ImageSharp, because everything just feels wrong.

As I workaround, we could render the text to another image, rotate that image, and then draw it on top.

Jonarw:
A minor general point: It would be nice to have consistent use of `this` and `var`. I think most portions of the library use `this` and `var` wherever possible, so it would be a nice opportunity to bring the `PngRenderingContext` in line while we are modifying most of its code anyway.

Another small-ish point:
Missing rotation aside - vertical text alignment also seems slightly off:
TextAnnotation - TextAnnotation - Transposed
(notice that the 'Text annotation 1' text is a bit high in the box).
But we could also leave this for a future issue, this is already a huge improvement for the `OxyPlot.ImageSharp` library!

VisualMelon:
@Jonarw yep: sorting out `this` should be done in latests commits (though the jpeg stuff has gone wrong, so don't look at that); I'll check `var`, and do a StypeCop pass at some point.

Not sure what to make of the vertical text alignment... might just be ImageSharp's own take on fonts, but I'll double check it's not me being silly.

Thanks for your prompt feedback.

Jonarw:
> Not sure what to make of the vertical text alignment... might just be ImageSharp's own take on fonts, but I'll double check it's not me being silly.

In `SkiaRenderContext` I ended up just using the Ascent and Descent font metrics for vertical alignment to make text baselines consistent across runs of different text - but I don't know if you have these metrics available in ImageSharp.

VisualMelon:
@Jonarw I guess `Font.Ascender` and `Font.Descender` are the same thing: I'll take a good look at your code then and try to sort it out.

Source/OxyPlot.ImageSharp/OxyPlot.ImageSharp.csproj
VisualMelon: I caved: langversion 8 it shall be ;)

VisualMelon:
Slowly getting somewhere...

TextAnnotation - TextAnnotation - Transposed

VisualMelon:
What feels like a week later, I have text rendering in a vaguely stable state.

Skia
TextAnnotation - TextAnnotation - Transposed

ImageSharp
TextAnnotation - TextAnnotation - Transposed

The font size discrepency is soewhat disconcerting, and I'll be looking into that next but the real world may get in the way.

P.S. SkiaSharp doesn't seem to support multiline text. I've implemented it for ImageSharp, but not properly tested.

VisualMelon:
@objorke do you recall what the `FontScaleFactor` is achieving in the `GraphicsRenderContext`?

Comparing outputs to work, I think the WinForms and Skia renders have somewhat small text, but they just about match Word and the ImageSharp render when I make the following changes:
 - Change `GraphicsRenderContext` such that `FontScaleFactor = 1.0` (rather than `0.8`)
 - Change `SkiaRenderContext.GetTextPaint` such that `TextSize = this.Convert(fontSize / 0.75);` (convert points (1/72nds of an inch) to 1/96ths of an inch (nominal dpi))

@Jonarw does this make sense to you? I can't find any documentation for what `TextSize` really is (based on your code and these observations I am assuming it is pixels)

Jonarw:
> @Jonarw does this make sense to you? I can't find any documentation for what TextSize really is (based on your code and these observations I am assuming it is pixels)

With the SkiaSharp renderer I took the WPF `CanvasRenderContext` as a reference. On WPF text size is specified in 1/96 inch, so exactly 1 pixel at 100% UI scaling.

In SkiaSharp the text size is specified in pixels, and as neither SkiaSharp nor WPF do any scaling of the font size at the moment they give similar output.

In the end we should just define what unit the text size is, put that in the documentation and stick to it everywhere. Considering that WPF is probably the most used platform of oxyplot, should we define the unit of text size to 1/96 inch? If ImageSharp uses 1/72 inch, you would have to multiply font sizes by 0.75.

> P.S. SkiaSharp doesn't seem to support multiline text. I've implemented it for ImageSharp, but not properly tested.

Yes I also realized this recently, I plan to address that in the near future.

VisualMelon:
@Jonarw it would seem odd to use something other than points, but not knowing the history and the fact that it's been this way for a while, I'm hesitant to make a suggestion to change. If everything else is going for 1/96ths, then I'll change ImageSharp to agree for now. As you say, this needs documenting somewhere...

Should the WinForms scale factor be changed to 0.75 as well?

objorke:
yes, this library was first implemented for WPF, and simply used WPF device-independent units (1/96 inch) for font sizes, line thickness etc.

VisualMelon:
@objorke thanks for clarifying!

VisualMelon:
Rebased, squashed, and style-cop'd. There are a couple of TODOs in there, but they shouldn't motivate any changes in the design. I don't intend to do much more for this already bloated PR.

Most of this is truly horrendous code, and I'm putting a few [notes in a gist](https://gist.github.com/VisualMelon/d1eb4bca4f52de86353421f96bcaeb58) to go some way to explaining why this is.

Jonarw:
Wow, respect for going through all that trouble!
I can't claim that I've gone through every line of the code, but I have not seen any obvious issues.
Test output also seems fine, so to me this looks good!

VisualMelon:
(Just rebased to resolve merge conflict in changelog)

objorke:
Fixed the snyk check

VisualMelon:
@objorke thanks. I'm going to divert my effors to #1531 now that this font stuff is fresh in my mind. Is there anything else you wanted done with this, or is it mergeable? When I run out of fun things to do I might come back and try to tidy this a little, maybe looking at reducing redundancy between Dpi aware renders while I am at as (as in the discussion somewhere above).

objorke:
This looks good to me, but I am not able to follow all the details... 
I am very happy to see all the development/improvements of this project right now!

VisualMelon:
Merged. Feel free to shout at me if anything ImageSharp related goes wrong.

PR #1546, opened 2020-05-07
Add RenderTarget property to SkiaRenderContext (#1539)

### LINKED ISSUE #1539: SkiaRenderContext should not snap pixels when rendering to vector graphic
Pixel-Snapping only really makes sense when rendering to a pixel graphic. So for SVG and PDF rendering, we should disable this.

`SkiaRenderContext` already has the `RendersToScreen` property, which controls subpixel text rendering. I propose to replace that by a `RenderTarget` enum, with the possible values 

- `Screen`: pixel-snapping enabled, subpixel text rendering enabled
- `PixelGraphic`: pixel-snapping enabled, subpixel text rendering disabled
- `VectorGraphic`: pixel-snapping disabled, subpixel text rendering disabled

Jonarw:
Fixes #1539.

### Checklist

- [ ] I have included examples or tests (can be verified by existing examples)

- Add `RenderTarget` property to `SkiaRenderContext`
- Update SkiaSharp-based exporters to define an appropriate `RenderTarget`

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ All notable changes to this project will be documented in this file.
 - Extract most of the functionality from OxyPlot.Wpf into OxyPlot.Wpf.Shared to allow code sharing with other WPF PlotViews (#1515)
 - WPF ExampleBrowser can switch between Canvas and SkiaSharp renderers (#1515)
 - OxyPlot.ImageSharp now targets .NET Standard 1.3 (#1530)
+- SkiaRenderContext does not apply pixel snapping when rendering to vector graphic (#1539)
 
 ### Removed
 - Remove PlotModel.Legends (#644)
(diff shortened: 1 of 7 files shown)


VisualMelon:
Looks good to me. I'm just wondering if `RenderTarget` could be in the main namespace, since this is potentially a concern for other renderers? (E.g. if we do add an ImageSharp based SVG renderer, then it will need to make the same choices).

Jonarw:
Interesting point.

So far the `SkiaRenderContext` is the only `IRenderContext` that potentially outputs to any of screen, pixel or vector graphic, and I don't think this will change in the near future.

Even if we add an ImageSharp-based SVG exporter, ImageSharp itself would only be responsible for text measurement, correct? So it wouldn't be concerned by this I think.

VisualMelon:
> Even if we add an ImageSharp-based SVG exporter, ImageSharp itself would only be responsible for text measurement, correct? 

Of course you're right (and I'm kind of glad I don't have to go through and check them all... but that means the SVG exporter (if it lives) will need refitting). Lacking an immediate practical consideration, I wouldn't worry about it (for the moment at least).

VisualMelon:
Maybe I'm missing it, but should `RenderTarget` be set somewhere for the WPF provision?

Jonarw:
Not necessarily, because `SkiaRenderContext.RenderTarget` defaults to `RenderTarget.Screen`.

VisualMelon:
I'd prefer it was explicit, but doesn't really matter.

<hr />

Good to merge in my opinion.

PR #1548, opened 2020-05-07
Add consistent support for unix line endings (#1545)

### LINKED ISSUE #1545: Unify support for non-CRLF (i.e. just LF) line endings
### Feature description

Presently line splitting is done by splitting on `\r\n`, either with [regex](https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Svg/SvgRenderContext.cs#L162), [string.Split](https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs#L565), or is left to some other library (e.g. in the case of WPF and WinForms/GDI+).

The later two examples support `\n` line endies out-of-the-box, so I suggest replacing all other instances of splitting strings with the following regex based method:

    var lines = Regex.Split(text, "\r?\n");

This should probably be stuffed into an method somewhere to foster consistency.

VisualMelon:
Fixes #1545 .

### Checklist

- Adds an example with mixed LF and CRLF line endings
- Adds a `SplitLines` method to `StringHelper` so that string splitting is done all in one place
- Call the aforementioned `SplitLines` methods in `OxyPlot.ImageSharp`, `OxyPlot.Svg`, `OxyPlot.Pdf`, and the `DrawMultilineText` helper so that they all will provide consistent support for line endings
- Remove some embarrassing micro-optimisation in `OxyPlot.ImageSharp.ImageRenderContext`

The method is not used in e.g. WinForms because the renderer handles the line-endings as intended already.

Originally the plan was to put the method in `RenderContextBase` since this is only a concern for rendering; however, `OxyPlot.Pdf` support would have been tricky in this case, and I don't see any reason why this method shouldn't be generally available. My main concern is that it is a general enough method that it may conflict with user's code, and as such shouldn't be an extension method.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@ All notable changes to this project will be documented in this file.
 - JpegExporter for OxyPlot.ImageSharp
 - Multi-line aligned text rendering example
 - WindowsForms ExampleBrowser can display transposed versions of examples (#1535)
+- Example for Issue #1545 showing the use of different line endings
+- Support for unix line endings in OxyPlot.ImageSharp, OxyPlot.Svg, and OxyPlot.Pdf (#1545)
 
 ### Changed
 - Legends model (#644)
--- a/Source/Examples/ExampleLibrary/Issues/Issues.cs
+++ b/Source/Examples/ExampleLibrary/Issues/Issues.cs
@@ -2265,6 +2265,20 @@ namespace ExampleLibrary
             return plot;
         }
 
+        [Example("#1545: Some render contexts do not support unix line endings.")]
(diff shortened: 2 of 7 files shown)


Jonarw:
Looks good to me in general!
I kind of share your concern about it being an extension method, mainly because `StringHelper` is in the main namespace (literally everyone using oxyplot will have that imported) and operates on a very common type. 

I would say, either make it internal or just use a regular static method, leaning heavily towards the latter option.

VisualMelon:
Changed to a `public static` non-extension method.

PR #1549, opened 2020-05-08
Add support for multi-line text to SkiaRenderContext (#1538)

### LINKED ISSUE #1538: Add multi-line text support to SkiaRenderContext
WPF and Windows Forms do kind of support this, but I would argue that their horizontal text alignment of multiline text is not optimal. This is the output of `CanvasRenderContext` for a corresponding example in my experimental branch:
image

I think it should rather look like this:
image

Do you agree?
In a first step, I plan to implement this for `SkiaRenderContext`, but we should also consolidate the different platforms at some point.

Jonarw:
Fixes #1538.

### Checklist

- Add support for multi-line text to `SkiaRenderContext` using the `StringHelper.SplitLines(...)` method
- Add an additional example showing multi-line text in combination with rotation and alignment
- Add an additional SkiaSharp-based SVG export test based in the new example to also cover the code path without text shaping 

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
+++ b/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
@@ -354,12 +354,14 @@ namespace OxyPlot.SkiaSharp
             var x = this.Convert(p.X);
             var y = this.Convert(p.Y);
 
-            var metrics = paint.FontMetrics;
+            var lines = StringHelper.SplitLines(text);
+            var lineHeight = paint.GetFontMetrics(out var metrics);
+
             var deltaY = verticalAlignment switch
             {
                 VerticalAlignment.Top => -metrics.Ascent,
-                VerticalAlignment.Middle => -(metrics.Ascent + metrics.Descent) / 2,
-                VerticalAlignment.Bottom => -metrics.Descent,
+                VerticalAlignment.Middle => -(metrics.Ascent + metrics.Descent + lineHeight * (lines.Length - 1)) / 2,
+                VerticalAlignment.Bottom => -metrics.Descent - lineHeight * (lines.Length - 1),
                 _ => throw new ArgumentOutOfRangeException(nameof(verticalAlignment))
             };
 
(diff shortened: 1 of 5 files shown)


Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -403,9 +410,11 @@ public OxySize MeasureText(string text, string fontFamily = null, double fontSiz
                 return new OxySize(0, 0);
             }
 
+            var lines = StringHelper.SplitLines(text);
             var paint = this.GetTextPaint(fontFamily, fontSize, fontWeight, out var shaper);
-            var width = this.MeasureText(text, shaper, paint);
-            var height = paint.GetFontMetrics(out _);
+            var height = paint.GetFontMetrics(out _) * lines.Length;
```
VisualMelon, line 415: This feels wrong, but I can't tell from looking at the examples (e.g. Issue 1545 example. It reads looks like it is adding extra line spacing for the 'last' line (i.e. `lineHeight` is the sum of the line height and spacing). Sorry if I've just got myself confused (this is something that `OxyPlot.Pdf` gets wrong, and I think `Svg` does as well but will be fixed with #1531)

Jonarw: Well `paint.GetFontMetrics(out _)` returns the suggested line height. To get the total height of the text block we multiply this by the number of lines, so for a text with one line nothing changes, for two lines we get double the height etc. Sounds right to me? Am I missing something?

VisualMelon: Now I am confused...

My reading of the `GetFontMetrics` method documention is that it returns the line height (i.e. `|descender| + |ascender|`) plus the [line gap](https://docs.microsoft.com/en-us/typography/opentype/spec/os2#stypolinegap). This would mean that the height of a box with 1 line of text should be less than half of the height of a box with 2 lines of text. If the line gap is small then this won't be noticeable (and will dimish further with more lines). However, the opposite seems to happen with WinForms (more lines means shorter boxes by number of lines), so clearly I'm missing something.

VisualMelon: I guess this is just part and parcel of not really having a specification for the `MeasureText` method.

Jonarw: Yes I think your last point is probably the key issue: It is not really specified what height the `MeasureText` method should actually return. I interpreted this as the line height in `SkiaRenderContext`, as that happened to coincide fairly well with what `CanvasRenderContext` gives, but other interpretations are also valid and possibly make more sense.

VisualMelon: Given I can't tell the difference, I'd suggest we pretend I never said anything, and instead we open a new issue discussing the possibility of qualifying `MeasureText`. We are going to have to write custom implementations of multi-text support for WPF/WinForms anyway to support text-alignment properly, so it would be good to nail down the requirements, and perhaps put the text-alignment code in a common location. That way, the individual render contexts only have to worry about measuring and drawing single lines of text.

VisualMelon:
Good to merge in my opinion.

Issue #72, opened 2014-08-14
Zooming LineSeries with Smooth="True" raises exception


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10064](http://oxyplot.codeplex.com/workitem/10064)**

**[Strevlik](http://www.codeplex.com/site/users/view/Strevlik)** 2013-07-08 16:25
In C# buid of library, huge zoomimg in at graph with oxy:LineSeries raises System.OutOfMemoryException.

Each change of zoom raises recomputation of whole spline (not just a part). One serie allocated about 200mb of memory for a short time (probably for computation). Few more zooms raise exception.

Part of the stracktrace: 

OxyPlot.CanonicalSplineHelper.Segment(IList<code>1 points, ScreenPoint pt0, ScreenPoint pt1, ScreenPoint pt2, ScreenPoint pt3, Double t1, Double t2, Double tolerance)

OxyPlot.CanonicalSplineHelper.CreateSpline(IList</code>1 points, Double tension, IList`1 tensions, Boolean isClosed, Double tolerance)

Issue #42, opened 2014-08-14
Index out of bound in ContourSeries


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10004](http://oxyplot.codeplex.com/workitem/10004)**

**[mmickael](http://www.codeplex.com/site/users/view/mmickael)** 2012-09-08 15:26
Version 4.13.1 

Not sure about the issue. Conrec.cs / public static void Contour(...) crashes with index out of bound in the following section.

// The indexing of im and jm should be noted as it has to

<pre><code>                                    // start from zero
                                    h[m] = d[i + im[m - 1], j + jm[m - 1]] - z[k];
                                    xh[m] = x[i + im[m - 1]];
                                    yh[m] = y[j + jm[m - 1]];
</code></pre>

Specific line is 

xh[m] = x[i + im[m - 1]]; 

Seems like x and y coming into the function are reversed but not sure. 

Did not encounter this problem with the same data in 4.12.1 or previous versions.

Will try to create a clear test case and post it here.

AlexeiScherbakov:
This is happened when count(x points)!=count(y points)
Example browser Peaks have equals point count x and y

AlexeiScherbakov:
I added example https://github.com/AlexeiScherbakov/oxyplot/commit/3d211cf177292b597d6ad3095b557606cc455726

Isolocis:
After cleaning up the example code a bit, the problem seems to be that an irregular grid is passed to the Conrec implementation. Conrec does not support irregular grids, so the best course of action seems to be to throw an exception when an irregular grid is recognised.

MGBergweiler:
There is nothing wrong with the source code, Conrec does support an irregular grid. The error is in the documentation, where the supplied example wrongly attributes yy to ColumnCoordinates and xx to RowCoordinates, whereas it should be the other way round: ColumnCoordinates are the abscissas (x-values), RowCoordinates are the ordinates (y-values) of the square grid of points. A consistency check of the dimensions of the arrays could be added. 

There is, however, an error in CalculateContours. In both for-loops shown below the "<" should be changed to "<=", otherwise a grid consisting of only one cell throws an exception. I suspect that a totally flat surface but with many cells would also trigger the same error.

![image](https://user-images.githubusercontent.com/50400171/80227551-14a2b500-864e-11ea-8b2c-8a1949ab72ef.png)

PR #1550, opened 2020-05-08
Reduce dependence on default axes in ExampleLibrary (#1547)

### LINKED ISSUE #1547: Examples should not rely on default axes
Some examples in the example library depend on the default axes (e.g. LineSeries > MarkerTypes). #992 says we want to remove default axes for 3.0. This requires adding axes to any examples that need them.

This should be as simple as pasting the following into the appropriate methods:

    model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom });

It would not make sense to do this automatically in the example library somehow, because some examples (e.g. those showing rendering capabilities) should not have axes at all.

Related: #1544

VisualMelon:
Fixes #1547 (just about).

### Checklist

- Introduce the `PlotModelUtilities.CreateDefaultPlotModel` method, which creates a `PlotModel` with 2 'default' linear axes.
- Change all examples that are not explicitly about default axes so that they no longer depend on default axes

### Outstanding concerns

 - I'm not entirely happy with the name `CreateDefaultPlotModel`.
 - Visual Studio kindly put all the `using ExampleLibrary.Utilities` statements in... but I'm not sure how we would want those formatted. StyleCop didn't have any strong feelings, but we don't use that for examples/tests anyway.

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
+++ b/Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
@@ -602,6 +602,7 @@ namespace ExampleLibrary
                     Position = AxisPosition.Left,
                     MinimumRange = 400
                 });
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom });
 
             return model;
         }
@@ -616,6 +617,7 @@ namespace ExampleLibrary
                     Position = AxisPosition.Left,
                     MaximumRange = 40
                 });
+            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom });
 
             return model;
         }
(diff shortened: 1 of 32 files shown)


VisualMelon:
I found a couple of interesting issue examples while I was scanning through them all to check nothing was obviously wrong which I might follow up:
 - **ScatterSeries and LinearColorAxis on the same plot (Closed)**
    Fails because it can't find the color axis (key should be `null` not `string.Empty`, and even if it could the values are all NaN so there is nothing to see
 - #42: ContourSeries not working for not square data
    Bit of a shame this isn't resolved yet
 - #72: Smooth
    Fails for some reason; title shows the old way of smoothing rather than new

Jonarw:
> I'm not entirely happy with the name `CreateDefaultPlotModel`

`CreateModelWithDefaultAxes` would be slightly more descriptive, but also a bit more clunky. You decide ;).

> Visual Studio kindly put all the using ExampleLibrary.Utilities statements in... but I'm not sure how we would want those formatted. 

I think in the `ExampleLibrary` we are generally a bit more relaxed about code style stuff, so I don't think we have to worry about that too much :)

Other than that, looks good to me!

VisualMelon:
Closing for now; may re-open in future if/when we decide to remove/stop supporting default axes.

Issue #992, opened 2016-09-09
Remove support for default axes


objorke:
It is currently possible to create a plot without axes.
If no axes are defined, default linear or category axes will be created.

To make the code simpler, remove this feature.

VisualMelon:
Is the idea to throw an exception in any scenario where previously a default axis would be created, and otherwise keep the `Default*Axis` properties?

Will the various examples that explicitly concern this feature simply be removed, replaced with tests to ensure `IPlotModel.Update` throws when expected?

objorke:
This was marked as a breaking change, so I think it is ok to remove properties not in use and throw exception for models that are OK in the current version.
 
I think most examples without axes should be updated to include axes, and there should be one test for the exception.

I think this issue was created in order to reduce the complexity of the update pass. I am not 100% if it is really worth changing this... We still have similar "undefined" situations when there are multiple axes and axis keys are not specified.

PR #1552, opened 2020-05-09
Do not transpose annotation only plots (#1544)

### LINKED ISSUE #1544: IsTransposed property in example browser is too generous
# Steps to reproduce

1. Open your example browser of choice with 'transpose' support OR look at the example library texts for any of the exporters
2. Look at any of the rendering examples (i.e. things without axes)
3. Set the 'Transpose' checkbox

Platform: Any
.NET version: Any

### Expected behaviour

Plots without axes should not be transposable: the `IsTransposable` property should detect and reject these cases, and consequently the `Transpose` checkbox should be disabled.

image

### Actual behaviour

We are allowed to transpose the plot, and it just complains about a lack of axes.

image

# Resolution

What should be the considered transposable? Is it sufficient to change the condition to always require axes, given that without axes you can't transpose any axes?

            return (model.Axes.Count > 0 || model.Series.Count > 0 || model.Annotations.Count > 0)
                && model.Axes.All(a => a.Position != AxisPosition.None)
                && model.Series.All(s =>
                   {
                       var type = s.GetType();
                       return s is XYAxisSeries
                              && type.GetTypeInfo().Assembly == typeof(PlotModel).GetTypeInfo().Assembly
                              && !NonTransposableSeriesTypes.Contains(type);
                   });

To

            return (model.Axes.Count > 0 && (model.Series.Count > 0 || model.Annotations.Count > 0))
                && /* snip */

PR to follow if nothing jumps out at me.

VisualMelon:
Fixes #1544 .

### Checklist

- Don't assume that examples with annotations but not axes or series are transposable
- Don't transpose plots with TileMapAnnotations

Since `PlotModel.EnsureDefaultAxes` doesn't observe annotations, and annotations don't influence axes, this should be robust.

Looking through the example library, the only example I noticed that was obviously broken were the `TileMapAnnotation` examples which are a special case to say the least, and is handled as such.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ All notable changes to this project will be documented in this file.
 - Possible out-of-bounds exception in HeatMapSeries HitTest (#1524)
 - WPF CanvasRenderContext draws ellipses too small by half stroke thickness (#1537)
 - Text measurement and rendering in OxyPlot.ImageSharp
+- ExampleLibrary reporting annotation-only PlotModels as transposable (#1544)
 
 ## [2.0.0] - 2019-10-19
 ### Added 
--- a/Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
+++ b/Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
@@ -36,21 +36,34 @@ namespace ExampleLibrary.Utilities
         };
 
         /// <summary>
+        /// Lists all Annotations that need axes and are NOT transposable
+        /// </summary>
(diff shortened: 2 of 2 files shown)


Jonarw:
Looks sensible to me. 
In general this really depends on what we want to do with #992. I don't really have a strong opinion about this issue, when I have some time to think about it I will drop a comment there.

Regardless, I think we could also just merge this, if we decide to go through with #992 not much would be lost.

VisualMelon:
@Jonarw I would agree: I don't think merging 992 would invalidate the value of this change, this would be easy to undo if it later seemed prudent, and the exception for `TileMapAnnotation` is probably necessary however we proceed.

PR #1553, opened 2020-05-09
Improve ImageSharp.ImageRenderContext text alignment

VisualMelon:
Makes multi-line text alignment in `OxyPlot.ImageSharp` conform to the ideal described in #1554

### Checklist

- Perform horizontal text alignment on a per-line basis, so that the block of text is aligned internally as expected
- Significantly simplify the `DrawText` method, hopefully making it significantly less resource intensive in the process
- Remove some default parameters on private methods in `ImageSharp.ImageRenderContext` which will only harbour bugs

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.ImageSharp.Tests/JpegExporterTests.cs
+++ b/Source/OxyPlot.ImageSharp.Tests/JpegExporterTests.cs
@@ -178,6 +178,16 @@ namespace OxyPlot.ImageSharp.Tests
             Assert.IsTrue(File.Exists(fileName));
         }
 
+        [Test]
+        public void TestMultilineAlignment()
+        {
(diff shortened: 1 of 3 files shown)


Jonarw:
Ah nice to see that rotated text on ImageSharp is also possible without that 'drawing on paths' stuff!

Looks good to me!

VisualMelon:
@Jonarw yep, it took so long to get that stupid method working that by the end I didn't realise I could do the transformation manually using the `Point` based overload that just generates the glyphs (I only discovered it by chance, and found I could use it to work out what the `Path` based overload was getting wrong with the string "aaaa", which it was putting through the basline, not on it).

Thanks for putting together 1554

JimBobSquarePants:
Text alignment will be fixed upstream by https://github.com/SixLabors/Fonts/pull/198

Issue #1554, opened 2020-05-10
Unify multi-line text alignment on all the platforms


Jonarw:
Benchmark should be the 'Rendering Capabilities > Multi-line Alignment' and 'Rendering Capabilities > Multi-line Alignment/Rotation' examples. 

- [ ] **WPF** - fix horizontal alignment
![image](https://user-images.githubusercontent.com/18097859/81493091-abea4800-929d-11ea-81fb-94404218d951.png)

- [ ] **Windows Forms** (and Core.Drawing) - fix horizontal alignment
![image](https://user-images.githubusercontent.com/18097859/81493107-de944080-929d-11ea-9654-76cea1b0c13d.png)

- [x] **SkiaSharp**
![image](https://user-images.githubusercontent.com/18097859/81493101-c7ede980-929d-11ea-9087-ca5e63cd3e0e.png)

- [ ] **ImageSharp** - fixed in #1553
![image](https://user-images.githubusercontent.com/18097859/81493319-c02f4480-929f-11ea-8be8-6222c0d3a7a8.png)

- [x] **SVG**
![image](https://user-images.githubusercontent.com/18097859/81493135-2fa43480-929e-11ea-8d4b-746d742a1d22.png)

- **OxyPlot.PdfExporter** - multiline text is not supported. Given we want to mark this as obsolete anyway, we don't have to worry about this too much I think
![image](https://user-images.githubusercontent.com/18097859/81493168-801b9200-929e-11ea-8680-901a41f15535.png)

- **OxyPlot.Pdf.PdfExporter** - multiline text is not supported. Given we want to mark this as obsolete anyway, we don't have to worry about this too much I think
![image](https://user-images.githubusercontent.com/18097859/81493197-d25cb300-929e-11ea-9b47-31b639b95710.png)

VisualMelon:
If we reach a happy conclusion with #1551, then the code for performing this layout task could be provided by `RenderContextBase`. Adjustments will need to be made by the different platforms to account for the different rendering APIs (e.g. `OxyPlot.Svg` can only render relative to baseline), but this could reduce the individual complexity of each renderer, and generally make the code tidier.

If I can sit down for more than a few minutes I'll try to put a version of this together for `OxyPlot.Svg` implementation, which really needs #1551 to make sense at all.

Issue #1551, opened 2020-05-08
What is MeasureText


VisualMelon:
The `IRenderContext.MeasureText` method doesn't have a formal specification, and is implemented in different ways all over the place.

We should consider formalising its behaviour so that we can make everything consistent before attempting to address inconsistencies in multi-line text rendering between platforms (see Issue #1538 and discussion in PR #1549).

Jonarw:
There appear to be some different nomenclatures around this topic. To be sure we are talking about the same thing, I will use the same nomenclature as in this blog post:
http://www.cyrilchandelier.com/understanding-fonts-and-uifont

Some thoughts:

### Height
- Similar to what we do in `OxyPlot.SkiaSharp` and `OxyPlot.ImageSharp` for vertical text alignment, I would argue that the vertical text measurement should not actually be dependent on the content. I think it is more important that the baselines of multiple text runs are guaranteed to be aligned regardless of content.
- A possible definition could be: 
`[text-height] = (|[ascent]| + |[descent]| + [leading]) * [number-of-lines] - [leading]`
which would be equivalent to:
`[text-height] = [line-height] * [number-of-lines] - [leading]`

### Width
- This should take into account text shaping, if applicable
- A possible definition could be:
`[text-width] = MAX([line] => SUM([advancement]))`

VisualMelon:
Thanks for putting that so clearly.

It's not immediately obvious how to implement the width computing manually in some APIs, but in practice the width doesn't seem to be much of an issue.

Jonarw:
Yes, probably most of the APIs do exactly this computation under the hood, and if there is no issue there would be no need to change anything. The definition should only serve as a reference in case of doubts / inconsistencies.

Issue #1516, opened 2020-04-27
Add WindowsForms PlotView based on SkiaSharp


Jonarw:
As it has been years since I last worked with WindowsForms, if someone else feels like taking on this task, please go ahead.

Things to consider:

- Should this use GPU-accelerated rendering or not? Or both?
  - There are [controls for both](https://github.com/mono/SkiaSharp/tree/master/source/SkiaSharp.Views/SkiaSharp.Views.WindowsForms) available on SkiaSharp side. 
  - Do we need more dependencies with GPU acceleration?
- Can code be shared with `OxyPlot.WindowsForms`?
- Add an example browser / integrate into WindowsForms example browser?

VisualMelon:
Given this would have to write to a buffer and copy/swap, the rendering engine could be abstracted away completely. There is no reason e.g. the ImageSharp renderer could be swapped in as well. I had a go with Skia and ran into a couple of rendering issues (apart from hating the abstraction I built), but if this would be an acceptable design I would be happy to do the leg work. My first attempt (a bit out of date) can be found here: https://github.com/oxyplot/oxyplot/compare/develop...VisualMelon:WinFormsSkia

I'd envisage adding an `IPlotPainter` interface to WinForms which writes to a bitmap (though, this could even be mediated through OxyImage, and be platform independent), and can be passed to the `PlotView` as an alternative to the 'direct' renderer. There is no reason rendering can't be done off-thread to improve UX. This wouldn't be owned by the `PlotView`.

The advantages of such a design would be:
 - Easy to add more renderers (but would you want to?)
 - Minimal changes needed with respect to code reuse/sharing
 - Painter interface could be used elsewhere (same abstraction could work for any UI platform, or indeed for any PNG exporter)
 - Easy for user to swap out renderer:
    - Painters can be their own (tiny) NuGet package
    - Doesn't need to swap out controls in the UI (can continue to use their own extensions to `PlotView`)
    - If they can't use their renderer of choice for some reason (e.g. library not audited or w/e), they can use the 'native' provision as a fallback, and this can be easily configured at runtime

Obvious concerns
 - Someone has to dispose the painters
 - Abstractions add complexity and overheads
 - Abstractions get in the way of custom performance and functionality

If we aren't sure this is bad idea yet, I'm equally happy to do the leg work so we can decide how much we hate it once it is working.

VisualMelon:
[...]

Issue #391, opened 2015-02-23
Refactor layout code in `PlotModel`


objorke:
Currently the layout "engine" is in the `RenderOverride` method.
Move the layout code to a new method, e.g. `UpdateLayout`

VisualMelon:
Here is a (probably not comprehensive) list of issues that may benefit from separating the layout and rendering passes:
 - #392 would be nice to have a swappable layout engine
 - #55 subplots
 - #388, #1621 issues with interactions with non-rendered (and so non-layouted) plot models
 - #1322 no events for layout (updated events don't provide the information people expect)
 - #1320 plot model synchronisation: a total nightmare at the moment; would be easier still if the layout engine were swappable

Other related issues:
 - #1556 an `ITextMeasurer` interface would allow layout without a dependency on `IRenderContext`, which would be nice

PR #1556, opened 2020-05-10
Overhaul Text Arrangement and Trimming

### LINKED ISSUE #1531: Improve vertical text alignment of OxyPlot.SvgExporter

VisualMelon:
Fixes #1531, #1559, and #1554

### Checklist

- [x] I have updated the change log (will do so if/when this is striped back and becomes a set of meaningful commits)

- Adds the `ITextMeasurer` interface, and implementations for all RenderContexts that can support it
- Adds the `ITextTrimmer` interface, and `SimpleTextTrimmer` implementation which uses 'character ellipsis' trimming mode as its default
- Adds the `TextArranger` class, which arranges multiline text as per #1554 and #1551 
- Changes all platforms in this repo to use `TextArranger` for text arrangement (thereby resolving #1531 and #1559 )
- Adds an SvgExporter to `OxyPlot.ImageSharp` with additional tests
- Removes the `UseVerticalAlignmentWorkaround` property from `OxyPlot.Svg` as it is redundant and will only get in the way

### Big Issues

- API for `ITextTrimmer` is weak: the render context determines the type of trimming that is done, which may not be ideal

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.Tests/Rendering/MockTextMeasurer.cs
+++ b/Source/OxyPlot.Tests/Rendering/MockTextMeasurer.cs
@@ -7,12 +7,12 @@
 // </summary>
 // --------------------------------------------------------------------------------------------------------------------
 
-using OxyPlot.Rendering;
-using System.Collections.Generic;
-using System.Linq;
-
 namespace OxyPlot.Tests
 {
+    using OxyPlot.Rendering;
+    using System.Collections.Generic;
+    using System.Linq;
+
     public class MockTextMeasurer : ITextMeasurer
     {
         public MockTextMeasurer()
--- a/Source/OxyPlot.Tests/Rendering/MockTextMeasurer.cs
+++ b/Source/OxyPlot.Tests/Rendering/MockTextMeasurer.cs
@@ -1,0 +1,59 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="MockTextMeasurer.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Provides a predicatble implementation of ITextMeasurer.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+using OxyPlot.Rendering;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace OxyPlot.Tests
+{
+    public class MockTextMeasurer : ITextMeasurer
+    {
+        public MockTextMeasurer()
+        {
+        }
+
+        public double Ascent { get; set; } = 0.6;
+        public double Descent { get; set; } = 0.3;
+        public double Leading { get; set; } = 0.1;
+        public double CharacterWidth { get; set; } = 0.8;
+
+        public HashSet<char> CharsWithWidth { get; } = new HashSet<char>();
+
+        public FontMetrics GetFontMetrics(string fontFamily, double fontSize, double fontWeight)
+        {
+            return new FontMetrics(fontSize * this.Ascent, fontSize * this.Descent, fontSize * this.Leading);
+        }
+
+        public double MeasureTextWidth(string text, string fontFamily, double fontSize, double fontWeight)
+        {
+            return fontSize * this.CharacterWidth * text.Count(CharsWithWidth.Contains);
+        }
+
+        public void AddBasicAlphabet()
+        {
+            for (char c = 'a'; c <= 'z'; c++)
+            {
+                this.CharsWithWidth.Add(c);
+                this.CharsWithWidth.Add(char.ToUpperInvariant(c));
+            }
+        }
+
+        public void AddBasicWhitespace()
+        {
+            this.CharsWithWidth.Add(' ');
+        }
+
+        public void AddEllipsisChars()
+        {
+            this.CharsWithWidth.Add(SimpleTextTrimmer.AsciiEllipsis[0]);
+            this.CharsWithWidth.Add(SimpleTextTrimmer.UnicodeEllipsis[0]);
+        }
+    }
+}
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -9,6 +9,7 @@
 
 namespace OxyPlot.Wpf
 {
+    using OxyPlot.Rendering;
     using System;
     using System.Collections.Generic;
     using System.IO;
@@ -27,7 +28,7 @@ namespace OxyPlot.Wpf
     /// <summary>
     /// Implements <see cref="IRenderContext" /> for <see cref="System.Windows.Controls.Canvas" />.
     /// </summary>
-    public class CanvasRenderContext : ClippingRenderContext
+    public class CanvasRenderContext : ClippingRenderContext, ITextMeasurer
     {
         /// <summary>
         /// The images in use
@@ -84,7 +85,16 @@ namespace OxyPlot.Wpf
             this.TextFormattingMode = TextFormattingMode.Display;
             this.TextMeasurementMethod = TextMeasurementMethod.TextBlock;
             this.RendersToScreen = true;
+            this.BalancedLineDrawingThicknessLimit = 3.5;
+            this.DefaultFontFamily = "Segoe UI";
+
+            this.TextArranger = new TextArranger(this, new SimpleTextTrimmer());
         }
+
+        /// <summary>
+        /// Gets or sets the <see cref="TextArranger"/> used by this instance.
+        /// </summary>
+        public TextArranger TextArranger { get; set; }
 
         /// <summary>
         /// Gets or sets the text measurement method.
@@ -97,6 +107,23 @@ namespace OxyPlot.Wpf
         /// </summary>
         /// <value>The text formatting mode. The default value is <see cref="System.Windows.Media.TextFormattingMode.Display"/>.</value>
         public TextFormattingMode TextFormattingMode { get; set; }
+
+        /// <summary>
+        /// Gets or sets the thickness limit for "balanced" line drawing.
+        /// </summary>
+        public double BalancedLineDrawingThicknessLimit { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to use stream geometry for lines and polygons rendering.
+        /// </summary>
+        /// <value><c>true</c> if stream geometry should be used; otherwise, <c>false</c> .</value>
+        /// <remarks>The XamlWriter does not serialize StreamGeometry, so set this to <c>false</c> if you want to export to XAML. Using stream geometry seems to be slightly faster than using path geometry.</remarks>
+        public bool UseStreamGeometry { get; set; }
+
+        /// <summary>
+        /// Gets or sets the default font familiy.
+        /// </summary>
+        public string DefaultFontFamily { get; set; }
 
         ///<inheritdoc/>
         public override void DrawEllipse(OxyRect rect, OxyColor fill, OxyColor stroke, double thickness, EdgeRenderingMode edgeRenderingMode)
@@ -306,127 +333,56 @@ namespace OxyPlot.Wpf
             string fontFamily,
             double fontSize,
             double fontWeight,
-            double rotate,
-            HorizontalAlignment halign,
-            VerticalAlignment valign,
+            double rotation,
+            HorizontalAlignment horizontalAlignment,
+            VerticalAlignment verticalAlignment,
             OxySize? maxSize)
         {
-            var tb = this.CreateAndAdd<TextBlock>();
-            tb.Text = text;
-            tb.Foreground = this.GetCachedBrush(fill);
-            if (fontFamily != null)
+            this.TextArranger.ArrangeText(p, text, fontFamily, fontSize, fontWeight, rotation, horizontalAlignment, verticalAlignment, maxSize, HorizontalAlignment.Left, TextVerticalAlignment.Top, out var lines, out var linePositions);
+
+            for (int i = 0; i < lines.Length; i++)
             {
+                var line = lines[i];
+                var linePosition = new Point(linePositions[i].X, linePositions[i].Y);
+
+                var tb = this.CreateAndAdd<TextBlock>();
+                tb.Text = line;
+                tb.Foreground = this.GetCachedBrush(fill);
                 tb.FontFamily = this.GetCachedFontFamily(fontFamily);
-            }
 
-            if (fontSize > 0)
-            {
-                tb.FontSize = fontSize;
-            }
-
-            if (fontWeight > 0)
-            {
-                tb.FontWeight = GetFontWeight(fontWeight);
-            }
-
-            TextOptions.SetTextFormattingMode(tb, this.TextFormattingMode);
-
-            double dx = 0;
-            double dy = 0;
-
-            if (maxSize != null || halign != HorizontalAlignment.Left || valign != VerticalAlignment.Top)
-            {
-                tb.Measure(new Size(1000, 1000));
-                var size = tb.DesiredSize;
-                if (maxSize != null)
+                if (fontSize > 0)
                 {
-                    if (size.Width > maxSize.Value.Width + 1e-3)
-                    {
-                        size.Width = Math.Max(maxSize.Value.Width, 0);
-                    }
-
-                    if (size.Height > maxSize.Value.Height + 1e-3)
-                    {
-                        size.Height = Math.Max(maxSize.Value.Height, 0);
-                    }
-
-                    tb.Width = size.Width;
-                    tb.Height = size.Height;
+                    tb.FontSize = fontSize;
                 }
 
-                if (halign == HorizontalAlignment.Center)
+                if (fontWeight > 0)
                 {
-                    dx = -size.Width / 2;
+                    tb.FontWeight = GetFontWeight(fontWeight);
                 }
 
-                if (halign == HorizontalAlignment.Right)
+                TextOptions.SetTextFormattingMode(tb, this.TextFormattingMode);
+
+                var transform = new TransformGroup();
+                if (Math.Abs(rotation) > double.Epsilon)
                 {
-                    dx = -size.Width;
+                    transform.Children.Add(new RotateTransform(rotation));
                 }
 
-                if (valign == VerticalAlignment.Middle)
+                transform.Children.Add(new TranslateTransform(linePosition.X, linePosition.Y));
+                tb.RenderTransform = transform;
+                if (tb.Clip != null)
                 {
-                    dy = -size.Height / 2;
+                    tb.Clip.Transform = tb.RenderTransform.Inverse as Transform;
                 }
 
-                if (valign == VerticalAlignment.Bottom)
-                {
-                    dy = -size.Height;
-                }
+                tb.SetValue(RenderOptions.ClearTypeHintProperty, ClearTypeHint.Enabled);
             }
-
-            var transform = new TransformGroup();
-            transform.Children.Add(new TranslateTransform(dx, dy));
-            if (Math.Abs(rotate) > double.Epsilon)
-            {
-                transform.Children.Add(new RotateTransform(rotate));
-            }
-
-            transform.Children.Add(new TranslateTransform(p.X, p.Y));
-            tb.RenderTransform = transform;
-            if (tb.Clip != null)
-            {
-                tb.Clip.Transform = tb.RenderTransform.Inverse as Transform;
-            }
-
-            tb.SetValue(RenderOptions.ClearTypeHintProperty, ClearTypeHint.Enabled);
         }
 
         ///<inheritdoc/>
         public override OxySize MeasureText(string text, string fontFamily, double fontSize, double fontWeight)
         {
-            if (string.IsNullOrEmpty(text))
-            {
-                return OxySize.Empty;
-            }
-
-            if (this.TextMeasurementMethod == TextMeasurementMethod.GlyphTypeface)
-            {
-                return MeasureTextByGlyphTypeface(text, fontFamily, fontSize, fontWeight);
-            }
-
-            var tb = new TextBlock { Text = text };
-
-            TextOptions.SetTextFormattingMode(tb, this.TextFormattingMode);
-
-            if (fontFamily != null)
-            {
-                tb.FontFamily = new FontFamily(fontFamily);
-            }
-
-            if (fontSize > 0)
-            {
-                tb.FontSize = fontSize;
-            }
-
-            if (fontWeight > 0)
-            {
-                tb.FontWeight = GetFontWeight(fontWeight);
-            }
-
-            tb.Measure(new Size(1000, 1000));
-
-            return new OxySize(tb.DesiredSize.Width, tb.DesiredSize.Height);
+            return this.TextArranger.MeasureText(text, fontFamily, fontSize, fontWeight);
         }
 
         ///<inheritdoc/>
@@ -494,7 +450,66 @@ namespace OxyPlot.Wpf
             this.clip = null;
         }
 
-        ///<inheritdoc/>
+        /// <inheritdoc/>
+        public FontMetrics GetFontMetrics(string fontFamily, double fontSize, double fontWeight)
+        {
+            var typeface = new Typeface(
+                new FontFamily(fontFamily), FontStyles.Normal, GetFontWeight(fontWeight), FontStretches.Normal);
+
+            if (!typeface.TryGetGlyphTypeface(out var glyphTypeface))
+            {
+                throw new InvalidOperationException("No glyph typeface found");
+            }
+
+            var lineHeight = Math.Abs(glyphTypeface.Height) * fontSize;
+            var ascender = Math.Abs(typeface.FontFamily.Baseline) * fontSize;
+            var descender = lineHeight - ascender;
+            var leading = (Math.Abs(typeface.FontFamily.LineSpacing) * fontSize) - lineHeight;
+
+            return new FontMetrics(ascender, descender, leading);
+        }
+
+        /// <inheritdoc/>
+        public double MeasureTextWidth(string text, string fontFamily, double fontSize, double fontWeight)
+        {
+            if (string.IsNullOrEmpty(text))
+            {
+                return 0.0;
+            }
+
+            if (this.TextMeasurementMethod == TextMeasurementMethod.GlyphTypeface)
+            {
+                return MeasureTextByGlyphTypeface(text, fontFamily, fontSize, fontWeight).Width;
+            }
+
+            var tb = new TextBlock { Text = text };
+
+            TextOptions.SetTextFormattingMode(tb, this.TextFormattingMode);
+
+            if (fontFamily != null)
+            {
+                tb.FontFamily = new FontFamily(fontFamily);
+            }
+
+            if (fontSize > 0)
+            {
+                tb.FontSize = fontSize;
+            }
+
+            if (fontWeight > 0)
+            {
+                tb.FontWeight = GetFontWeight(fontWeight);
+            }
+
+            tb.Measure(new Size(1000, 1000));
+
+            return tb.DesiredSize.Width;
+        }
+
+        /// <summary>
+        /// Cleans up resources not in use.
+        /// </summary>
+        /// <remarks>This method is called at the end of each rendering.</remarks>
         public override void CleanUp()
         {
             // Find the images in the cache that has not been used since last call to this method
@@ -760,11 +775,11 @@ namespace OxyPlot.Wpf
         }
 
         /// <summary>
-        /// Converts an <see cref="OxyRect" /> to a <see cref="Rect" />.
+        /// Converts an <see cref="OxyRect" /> to a <see cref="System.Windows.Rect" />.
         /// </summary>
         /// <param name="r">The rectangle.</param>
-        /// <returns>A <see cref="Rect" />.</returns>
-        protected static Rect ToRect(OxyRect r)
+        /// <returns>A <see cref="System.Windows.Rect" />.</returns>
+        private static Rect ToRect(OxyRect r)
         {
             return new Rect(r.Left, r.Top, r.Width, r.Height);
         }
@@ -780,8 +795,8 @@ namespace OxyPlot.Wpf
         {
             switch (edgeRenderingMode)
             {
-                case EdgeRenderingMode.Adaptive when IsStraightLine(points):
-                case EdgeRenderingMode.Automatic when IsStraightLine(points):
+                case EdgeRenderingMode.Adaptive when RenderContextBase.IsStraightLine(points):
+                case EdgeRenderingMode.Automatic when RenderContextBase.IsStraightLine(points):
                 case EdgeRenderingMode.PreferSharpness:
                     return points.Select(p => PixelLayout.Snap(p.X, p.Y, strokeThickness, this.VisualOffset, this.DpiScale));
                 default:
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -338,6 +338,7 @@ namespace OxyPlot.Wpf
             VerticalAlignment verticalAlignment,
             OxySize? maxSize)
         {
+            fontFamily ??= this.DefaultFontFamily;
             this.TextArranger.ArrangeText(p, text, fontFamily, fontSize, fontWeight, rotation, horizontalAlignment, verticalAlignment, maxSize, HorizontalAlignment.Left, TextVerticalAlignment.Top, out var lines, out var linePositions);
 
             for (int i = 0; i < lines.Length; i++)
@@ -779,7 +780,7 @@ namespace OxyPlot.Wpf
         /// </summary>
         /// <param name="r">The rectangle.</param>
         /// <returns>A <see cref="System.Windows.Rect" />.</returns>
-        private static Rect ToRect(OxyRect r)
+        protected static Rect ToRect(OxyRect r)
         {
             return new Rect(r.Left, r.Top, r.Width, r.Height);
         }
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -92,9 +92,9 @@ namespace OxyPlot.Wpf
         }
 
         /// <summary>
-        /// Gets or sets the <see cref="TextArranger"/> used by this instance.
+        /// Gets or sets the <see cref="ITextArranger"/> used by this instance.
         /// </summary>
-        public TextArranger TextArranger { get; set; }
+        public ITextArranger TextArranger { get; set; }
 
         /// <summary>
         /// Gets or sets the text measurement method.
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -460,7 +460,13 @@ namespace OxyPlot.Wpf
 
             if (!typeface.TryGetGlyphTypeface(out var glyphTypeface))
             {
-                throw new InvalidOperationException("No glyph typeface found");
+                var defaultTypeface = new Typeface(
+                    new FontFamily(this.DefaultFontFamily), FontStyles.Normal, GetFontWeight(fontWeight), FontStretches.Normal);
+
+                if (!defaultTypeface.TryGetGlyphTypeface(out glyphTypeface))
+                {
+                    throw new InvalidOperationException("No glyph typeface found");
+                }
             }
 
             var lineHeight = Math.Abs(glyphTypeface.Height) * fontSize;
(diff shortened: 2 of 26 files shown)


VisualMelon:
MaxSize support is probably pretty slow, and will be caught out by strings with combining characters (need to do some research on that), but gives pretty reasonable results. It could be performed more quickly by render contexts with access to glyph information (e.g. SkiaSharp), where cropping the line would be logarithmic (i.e. irrelevant) rather that log-linear.

WPF (the bottom-most line is drawn clipped)

image

Skia on WPF (the bottom-most line is skipped completely)

image

VisualMelon:
@Jonarw would you mind taking a gander at the changes in Skia, and let me know if this looks too abhorrent to you to be an acceptable change?

Having done the provisional implementation of `maxSize`, I think putting all this stuff in one place is a good idea, but it may be inappropriate to clip by text rather than pixels for e.g. SkiaSharp if it can do clipping like occurs with WPF. That might motivates adding additional functionality to `ArrangeText` so that it can arrange according to `maxSize` but not perform the clipping itself.

Jonarw:
Generally I think the `ArrangeText()` function is a very good idea to make text rendering more consistent between the platforms. However I think we should think a bit more about what exactly this should do.

I am a bit sceptical if clipping on character level is really the way to go for implementing `maxSize`. I think this could be unreliable for certain character combinations, and performance will probably fairly bad (though this should not be our main concern).

I am more concerned that we might be trying to do something very complex that we don't actually really need (see also my comment in #1559) - after all, I haven't missed this feature on SkiaSharp so far.

Jonarw:
I am also not too sure how about I feel about guessing the font metrics (ascent, descent, leading) by doing multiple text measurements. At the current state, the exact algorithm for text height measurement should be an implementation detail of the render context, which we shouldn't rely on. If we effectively move the responsibility for vertical text arrangement out of the render context and into the `ArrangeText()` function, we should also adapt the API accordingly to make the responsibilities clear. This could mean:

- `IRenderContext.MeasureText()` explicitly only works for single lines of text and only measures the width
- `IRenderContext.DrawText()` explicitly only works for single lines of text
- `IRenderContext` offers some way to get the relevant font metrics (ascent, descent, leading) the `ArrageText()` function needs to do its job properly

VisualMelon:
@Jonarw yeah, it's pretty horrendous, and I think what you describe is actually close to how OxyPlot was designed: `DrawText` is even documented as not working with multi-line text.

To the end, I would suggest keeping the current API as it is (i.e. `MeasureText` measures whole texts, `DrawText` _should_ support multi-line text), but add additional methods which provide what you describe (i.e. measure width of text, retrieve metrics for a given font/size/weight combination). It could be that these are not made available through `IRenderContext` - because axes and series and what-not shouldn't need them - but through another `ITextMeasurer` interface. That would avoid any breaking changes, and make it easy to push all the text-arranging stuff into it's own class.

Jonarw:
You are right of course, from the perspective of `Axes`, `Series` etc. the `IRenderContext` interface should not change.
_Or maybe it could change, and we replace the removed/changed APIs by extensions, similar to the `RenderingExtensions.DrawClippedXXX()` stuff?_

I haven't completely wrapped my mind around this, I need to think about it some more...

VisualMelon:
Thanks for the comments on what must look a complete mess at the moment. I'll probably next implement your suggestions about abstracting out the 'basic' measurement stuff (so we don't have to infer metrics), and put the whole thing in its own class. Rough plan:
 - add `ITextMeasurer` interface with methods
    - `GetFontMetrics(family, size, weight)` returning ascender, descender, and leading (probably in a struct if I can be bothered to add another new file)
    - `Measure(text, family, size, weight)` returning the width of the text
 - add `TextArranger` class with members
    - `ITextMeasurer Measurer` (or would we rather pass it as a parameter as I have been doing so far? All DPI awareness is in the `ITextMeasurer`: should the `RenderContext` have to know about this?)
    - `MeasureText(text, family, size, weight)` retuning the size of the (potentially multiline) text as an `OxySize` according to jonarw's proposal in #1551 
    - `ArrangeText(...)` doing what it does now, but leaning on the `ITextMeasurer` to arrange (potentiall multiline) text however it is asked to do so
 - patch `ImageSharp`, `SkiaSharp`, and managed-only SVG RenderContexts as necessary

The additional interface will duly provide the information the managed only `SvgRenderContext` needs to do probably text aligning because it won't have to guess the descent anymore.

VisualMelon:
Recent changes:
 - Moved `ArrangeText` into it's own class
 - Moved text-trimming into its own class which implements ellipsises
 - Broke Winforms somewhat (can't seem to get `GetFontMetrics` right, any input welcome)
 - Moved Baseline offset into `ArrangeText`, so the `DrawText` methods are even tidier than ever

Next tasks:
 - Fix winforms
 - Refit RenderContexts so that they lean on `TextArranger.MeasureText` and move the measuring logic into `MeasureTextWidth` (so they don't get into an infinite loop)

Any feedback (esspecially negative) welcome as always.

VisualMelon:
Refitting done, so `ImageRenderContext`, `SvgRenderContext`, and `SkiaRenderContext` shouldn't change any more. @jonarw all else aside, does the interface to `TextArranger` look half-sane?

Also refitted the TextTrimmer so it supports Unicode and adding additional tests. This of course has made it slower... but I don't think that should be a concern: it is rarely used, and would be providing a capability that is presently unavailable almost everywhere. I doubt it is perfect (i.e. there will be strings that it gives nonsense results on), but it should be fairly easy to extend by making the central regexes more complicated. Annoyingly we can't compile the regexes because we started .NET Standard 1.0; however, the architecture means that if someone really cared they can swap out the text trimmer for their own; whether this is a good architecture is another issue.

WinForms is still broken; I'm having difficultly extracting the font metrics (i.e. `Graphics.RenderContext.GetFontMetrics` is broken).

VisualMelon:
Fixed WinForms.

I think that means everything is working as I understand it should. I've not StyleCop'd this, but I'm going to open for review because there is nothing else I'm intending to do at the moment.

Jonarw:
Looks very promising!
I did not get into this too deep, but the general concept looks sensible to me. Over the weekend I will have a closer look.

Just for my understanding: This adapts ImageSharp, SkiaSharp, WinForms and Pdf, at the moment, right? So WPF would still have to be adjusted for this (but could also be its own issue and PR).

Also we should then add more sophisticated `ITextTrimmer`s where possible (i.e. for SkiaSharp) - but again, this does not have to be part of this PR in my opinion.

VisualMelon:
@Jonarw thanks. I'll eagerly await your feedback, and perhaps put my efforts into #1516 

It also adapts the built-in SVG renderer, which now depends on an `ITextMeasurer` rather than a whole `IRenderTarget`, which is something that could be exploited in the future.

I've not refitted WPF yet because frankly I'm frightened of it, but I can do that work for this PR or it can be done later if @objorke approves. For the purposes of consistent vertical alignment it is probably a good idea, but it may still be that we leave the trimming to it (though as I've discussed elsewhere, trimming by character seems a much better strategy to me) which could complicate matters with multi-line text

Regards advanced trimmers: yep, that was pretty much what I had in mind. The performance so far is perfectly good in my opinion, but there is plenty of scope for improvement. I won't claim the current trimmer is perfect, but it should be fairly easy to adapt if we find difficulties, and the architecture means it should be easy to support users without regular releases.

VisualMelon:
Added more tests and found a bug. Is this enough labels?

VisualMelon:
Refitted WPF in [another branch](https://github.com/VisualMelon/oxyplot/tree/WpfRefit). The only issue was the handling of the default font family being deferred to `TextBlock` which had to be changed.

Source/OxyPlot/Rendering/FontMetrics.cs
```diff
@@ -0,0 +1,50 @@
...
+
+        /// <summary>
+        /// The distance between the bottom of a line of text and the top of the next line of text.
+        /// </summary>
+        public double Leading { get; set; }
+
+        /// <summary>
+        /// The line height of the font.
+        /// </summary>
+        public double LineHeight => this.Ascender + this.Descender;
```
Jonarw, line 48: I know there are different definitions around these metrics, but all sources I could find referred to the line height as `ascent + descent + leading`. What is calculated here seems to be referred to as 'Font Height' or 'Cell Height'.

Source/OxyPlot/Rendering/RenderContext/RenderContextBase.cs
```diff
@@ -11,6 +11,7 @@ namespace OxyPlot
 {
     using System;
     using System.Collections.Generic;
+    using System.Linq;
```
Jonarw, line 14: Not needed

Source/OxyPlot/Rendering/SimpleTextTrimmer.cs
```diff
@@ -0,0 +1,217 @@
...
+// <summary>
+//   A simple trimmer that doesn't use glyph information.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Rendering
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq.Expressions;
```
Jonarw, line 14: Not needed

Source/OxyPlot/Rendering/SimpleTextTrimmer.cs
```diff
@@ -0,0 +1,217 @@
...
+        public static readonly string UnicodeEllipsis = "…";
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="SimpleTextTrimmer" /> class.
+        /// </summary>
+        public SimpleTextTrimmer()
+        {
+            this.TrimToWord = false;
+            this.Ellipsis = AsciiEllipsis;
+            this.AppendEllipsis = false;
```
Jonarw, line 39: Should we enable `AppendEllipsis` by default? Now that we have it in there, I think in most cases this would be preferable.
Also, I'm wondering whether the default ellipsis should be the `UnicodeEllipsis` - this will give the better appearance where supported, but could be a bit more problematic with font support.

Source/OxyPlot/Rendering/TextArranger.cs
```diff
@@ -0,0 +1,258 @@
...
+            var offsetDescender = new ScreenVector(-sin * descender, cos * descender);
+
+            // align to bounds
+            var offsetBoundsX = offsetBoundsWidth * 0.0; // keep centerline
+            var offsetBoundsY = offsetBoundsHeight * (ResolveOffset(verticalAlignment) - 0.5); // find the top of the top line
+
+            p += offsetBoundsX + offsetBoundsY;
+
+            // align lines within bounds
+            bool useBaslineOffset = targetVerticalAlignment == TextVerticalAlignment.Baseline;
```
Jonarw, line 151: Typo? (useBas**e**lineOffset)

Source/OxyPlot/Rendering/TextArranger.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// Center ->  0.0
+        /// Right  -> +0.5
+        /// </remarks>
+        private static double ResolveOffset(HorizontalAlignment horizontalAlignment)
+        {
+            var offset =
+                horizontalAlignment == HorizontalAlignment.Left ? -0.5
+                : horizontalAlignment == HorizontalAlignment.Center ? 0.0
+                : horizontalAlignment == HorizontalAlignment.Right ? 0.5
+                : throw new ArgumentException(nameof(horizontalAlignment));
```
Jonarw, line 232: Could this be a switch expression?
I know you don't like them too much, but I find this a bit hard to read ;).

Source/OxyPlot/Rendering/TextArranger.cs
```diff
@@ -0,0 +1,258 @@
...
+        /// Middle ->  0.0
+        /// Bottom -> -0.5
+        /// </remarks>
+        private static double ResolveOffset(VerticalAlignment verticalAlignment)
+        {
+            var offset =
+                verticalAlignment == VerticalAlignment.Top ? 0.5
+                : verticalAlignment == VerticalAlignment.Middle ? 0.0
+                : verticalAlignment == VerticalAlignment.Bottom ? -0.5
+                : throw new ArgumentException(nameof(verticalAlignment));
```
Jonarw, line 253: Same as above.

Source/OxyPlot/Svg/SvgRenderContext.cs
```diff
@@ -56,15 +55,10 @@ public SvgRenderContext(Stream s, double width, double height, bool isDocument,
...
         /// </summary>
-        /// <value>The text measurer.</value>
-        public IRenderContext TextMeasurer { get; set; }
-
-        /// <summary>
-        /// Gets or sets a value indicating whether to use a workaround for vertical text alignment to support renderers with limited support for the dominate-baseline attribute.
-        /// </summary>
-        public bool UseVerticalTextAlignmentWorkaround { get; set; }
+        /// <value>The text arranger.</value>
+        public TextArranger TextArranger { get; set; }
```
Jonarw, line 61: Could we add this to `RenderContextBase`?

Source/OxyPlot/Svg/SvgRenderContext.cs
```diff
@@ -128,6 +122,20 @@ public override void DrawRectangle(OxyRect rect, OxyColor fill, OxyColor stroke,
...
+        /// <summary>
+        /// Determines the font descent for the given font.
+        /// </summary>
+        /// <param name="fontFamily">The font family.</param>
+        /// <param name="fontSize">Size of the font.</param>
+        /// <param name="fontWeight">The font weight.</param>
+        /// <returns>A positive value, that distance between the baseline and the bottom of any rendered text.</returns>
+        public virtual double GetFontDescent(string fontFamily, double fontSize, double fontWeight)
+        {
+            var lineHeight = this.MeasureText("Ag", fontFamily, fontSize, fontWeight).Height;
```
Jonarw, line 134: See comment in `FontMetrics.cs` - I think the name `lineHeight` might be misleading.

Jonarw (commented):
I think this has turned out really well!
That's a lot of labels 😆

Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -9,10 +9,13 @@
 
 namespace OxyPlot.Wpf
 {
+    using OxyPlot.Rendering;
     using System;
     using System.Collections.Generic;
     using System.IO;
     using System.Linq;
+    using System.Reflection;
+    using System.Security.Cryptography;
```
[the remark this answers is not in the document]
VisualMelon: Thanks. I finally worked out why these keep appearing: I type `Hash` for hash-set, and if I don't have `using System.Collections.Generic` already, VS decides I meant `Hash` instead ;)

Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -635,6 +638,33 @@ public void ResetClip()
...
+            {
+                throw new InvalidOperationException("No glyph typeface found");
+            }
+
+            var lineHeight = Math.Abs(glyphTypeface.Height) * fontSize;
+            var ascender = Math.Abs(typeface.FontFamily.Baseline) * fontSize;
+            var descender = lineHeight - ascender;
+            var leading = (Math.Abs(typeface.FontFamily.LineSpacing) * fontSize) - lineHeight;
+
+            // TODO: review DPI support here
```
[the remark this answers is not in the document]
VisualMelon: Good idea. Having done the trial refit of WPF, I'm confident it works, so that can certainly be replaced.

Source/OxyPlot/Rendering/FontMetrics.cs
VisualMelon: Good point. I've been getting this wrong for a while: `CellHeight` sounds the least confusing, so I'll run with that.

Source/OxyPlot/Rendering/TextArranger.cs
VisualMelon: Shall do. As long as there is an explicit default I don't mind them; It's the implicit throw that irritates me (so I have a dim view of the design altogether).

Source/OxyPlot/Svg/SvgRenderContext.cs
VisualMelon: Yep, though I think that method should be gone altogether.

Source/OxyPlot/Svg/SvgRenderContext.cs
VisualMelon: I'm on the fence about this. I'll change it, then we can see how we feel after that.

VisualMelon:
@Jonarw thank you kindly!

Source/OxyPlot/Pdf/PdfRenderContext.cs
```diff
@@ -410,6 +382,22 @@ public override void ResetClip()
             this.doc.RestoreState();
         }
 
+        /// <inheritdoc/>
+        public FontMetrics GetFontMetrics(string fontFamily, double fontSize, double fontWeight)
+        {
+            var font = PortableDocument.GetFont(fontFamily, fontWeight > 500, false);
+            return font.GetFontMetrics(fontSize / 96 * 72);
```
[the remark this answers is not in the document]
VisualMelon: Am I going mad, or should they not all be `* 96 / 72`? (including `DrawText`)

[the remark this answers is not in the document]
VisualMelon: No I am definitely going mad.

Source/OxyPlot/Svg/SvgRenderContext.cs
VisualMelon: Without pulling the `ITextMeasurer` chunk of each render context into a new class or abstractly implementing `ITextMeasurer` from `RenderContextBase`, this would be a tad unpleasant. Such a change would probably make sense, but I'd sooner keep the current design, and leave that for another PR.

Source/OxyPlot/Pdf/PdfRenderContext.cs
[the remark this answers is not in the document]
VisualMelon: In conclusion, it shouldn't be doing anything here: it's going from 1/96ths to 1/96ths. I could convert one way and back, but there is no point, so I've just removed the conversions.

VisualMelon:
I believe I've addressed everything except moving the `TextArranger` property to `RenderContextBase`. A few ways of doing this come to mind:
 1. Just move it there (and maybe the `MeasureText` method). This doesn't achieve a whole lot, and means that the extending classes will have to initialise a member they don't 'own', as it were, because it must be created in its own constructor body.
2. Move it there, and implement `ITextMeasurer` abstractly in `RenderContextBase`, so that it can create the `TextArranger` itself.
3. Change the methods in `TextArranger` so that they take the `ITextMeasurer` as a parameter.
4. Move it to `RenderContextBase`, but have it passed to a constructor. This requires extracting the font-handling code from each render context into a separate class. While this may make the most sense, it has the difficulty of supplying the font handling class to the extending class itself.

None of these fill me with joy. Given the low cost of having it repeated in each class, I'm gong to leave it as is. If anyone has a good idea for how this chould be done, or can make a strong argument that it should be done, then I'll be happy to do the leg work.

VisualMelon:
The latest set of commits include the WPF refit, which is why the tests are failing. I believe these are all a consequence of the change in the how text heights are determined: they are now computed as per jonarw's proposal in #1551, rather than by WPF itself.

The example which is most butchered by the refit is the `DrawText - MaxSize` example :

image

_If_ we keep the new trimming behaviour, then this example might want to be changed/replaced. If not, then there will be a bit of work involves in implementing the old behaviour via the `TextArranger`.

Jonarw:
> If we keep the new trimming behaviour, then this example might want to be changed/replaced.

Agree.

With the failed test, you could wait on #1562, which also adjusts the same unit tests. Some of the failed tests should sort themselves out, for the `Axis_Desired_X` tests we would just have to adapt the values.

VisualMelon:
@Jonarw waiting on #1562 sounds good to me. I'd be happy to merge that now if you are (any little issue with it we can deal with over time); I think this need Objorke's blessing since it is architectural.

Jonarw:
To me this looks very good now!
@VisualMelon is this ready from your side?
@objorke would you be ok with this?

VisualMelon:
(Rebased)

I don't believe that was anything more I intended to do with this.

Considerations for the future, perhaps:
 - Review text trimming API (should it be configurable through `IRenderContext`?)
 - More tests for the text trimmer(s), preferablly written by someone who knows a little more about unicode than I.
 - Should `TextArranger` be a member of `RenderContextBase`? (this would also help with consistency.)
 - Should the text measuring/font-handling code be pulled out of the render contexts? (This would be beneficial in a couple of places (e.g. making the various SVG exporters lighter), but may just be adding complexity.
 - A text trimmer that an use glyph information (e.g. as is available to Skia) for more efficient text trimming (this will be a concern for any change in a API, as glyph may not map to character (I don't know enough to really comment.)

Jonarw:
I just stumbled over this PR again and I still think it would be very nice to merge this!

I think after all this time we can only assume that there would be no objections from objorke's side.

@VisualMelon Would you rebase this to the current develop, so we could verify this with the latest changes?

VisualMelon:
@Jonarw my main reservation is that it changes the text trimming. I've not really solicited @objorke's feedback since this came to be in some sort of a stable state, so I've been keen to wait a little longer now that I've actually pinged him here.

I'll get on with rebasing and retesting anyhow, and I'd better make some new examples to show text trimming as it is intended.

objorke:
Sorry, I am not able to follow all the conversations here at the moment. But I am very happy to see all the improvements you are implementing on the library, it is way more thorough than the original design. I agree text trimming could be a separate PR, but wouldn't object if both are merged in this one! Keep going!
ps. I hope performance is not severely affected with these changes.

VisualMelon:
@objorke thanks for the comment.

The only performance implications should be when there are large numbers of text annotations/labels/etc. with a max size that actually constrains them, otherwise it's just a big refactoring that makes everything a little more consistent.

The trimming algorithm is `O(n log(n))` where `n` is the number of characters in a line of text. @Jonarw and I have discussed options of replacing this with a linear algorithm on platforms that provide per-glyph character information, which can be done without modifying the core library under with this architecture, but I've not seen a situation where any significant number of plot elements need trimming, so absolute performance hasn't been a priority. I'll put together some excessive examples to see how they behave.

VisualMelon:
Added two performance examples with ~100 items of text: trimmed in one case, not trimmed in the other. With WinForms and the WPF/Skia renderer, the trimming example is observably slower, but not significantly so.

With WPF, it is significantly slower, which is a regressions as previously the trimming was just as fast. However, I can't observe any degradation in responsiveness if `TextMeasurementMethod.GlyphTypeface` is used, and both examples are noticeably quicker.

The cost is a function of the speed at which text is measured by the platform, and the `TextMeasurementMethod.GlyphTypeface` seems to be much faster than deferring to TextBlock. Is there a reason it is not used by default? Since it does a scan through the glyphs themselves, it is an obvious place we could use a linear trimming method in the future.

Jonarw:
I think the performance should be good enough for now. We can tweak this in the future for render contexts where we have individual glyph information and the situation with WPF should already be much better once we can go through #1609.

Source/OxyPlot/Rendering/FontMetrics.cs
```diff
@@ -0,0 +1,67 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="TextVerticalAlignment.cs" company="OxyPlot">
+//   Copyright (c) 2020 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Contains metrics for a given font.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+using System;
```
Jonarw, line 10: Inside namespace

Source/OxyPlot/Rendering/FontMetrics.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            if (descender < 0)
+            {
+                throw new ArgumentException("Descender must be non-negative.", nameof(descender));
+            }
+
+            Ascender = ascender;
+            Descender = descender;
+            Leading = leading;
```
Jonarw, line 39: this

Source/OxyPlot/Rendering/TextArranger.cs
```diff
@@ -0,0 +1,259 @@
...
+// <summary>
+//   Arranges text.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Rendering
+{
+    using System;
+    using System.Linq;
+    using System.Xml.Linq;
```
Jonarw, line 14: Not needed

Source/OxyPlot.Tests/Rendering/MockTextMeasurer.cs
```diff
@@ -0,0 +1,59 @@
...
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Provides a predicatble implementation of ITextMeasurer.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+using OxyPlot.Rendering;
+using System.Collections.Generic;
+using System.Linq;
```
Jonarw, line 12: Inside namespace

Jonarw (commented):
Just came across some small nitpick-stuff

VisualMelon:
@Jonarw thanks. I'd better let StyleCop have another go at it soon.

Source/OxyPlot/Rendering/TextArranger.cs
```diff
@@ -0,0 +1,259 @@
...
+
+        /// <summary>
+        /// Gets or sets the <see cref="ITextTrimmer"/> used by this instance.
+        /// </summary>
+        public ITextTrimmer TextTrimmer { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to squash the bounds of trimmed text to the size the text.
+        /// </summary>
+        public bool SquashTrimmedTextBounds { get; set; }
```
Jonarw, line 47: I can't quite figure out what exactly this property does - could this be explained a bit more clearly?

Source/OxyPlot/Rendering/TextArranger.cs
```diff
@@ -0,0 +1,259 @@
...
+            // turn metrics into vectors
+            var offsetBoundsWidth = new ScreenVector(cos * bounds.Width, sin * bounds.Width);
+            var offsetBoundsHeight = new ScreenVector(-sin * bounds.Height, cos * bounds.Height);
+
+            var offsetLineHeight = new ScreenVector(-sin * lineHeight, cos * lineHeight);
+            var offsetLeading = new ScreenVector(-sin * leading, cos * leading);
+            var offsetDescender = new ScreenVector(-sin * descender, cos * descender);
+
+            // align to bounds
+            var offsetBoundsX = offsetBoundsWidth * 0.0; // keep centerline
```
Jonarw, line 146: `offsetBoundsX` is always a (0,0)-vector?! Also this is the only usage of `offsetBoundsWidth`, so its value has no meaning? Am I missing something?

VisualMelon: I don't think you're missing anything... I'd better take a proper look at it.

Source/OxyPlot/Rendering/TextArranger.cs
VisualMelon: I'm struggling to explain this clearly... maybe a picture will help for your benefit (obviously the documentation needs text).

This is `SquashTrimmedTextBounds = true`:
squash

This is `SquashTrimmedTextBounds = false`:
nosquash

When `true`, if text is trimmed horizontally, then the bounds are squashed to the bounds of trimmed value; otherwise they are 'clamped' to the max size. This only changes the way that the text is arranged, and looking at those two figures now, I think it should be `true`. We could remove it and just treat it as `true`... that would solve all my problems

VisualMelon: Just realised this only applies vertically... the horizontal is already bounded... I think I'll make it 'hard coded' `true` instead, to reduce confusion.

VisualMelon:
(Need to fix the tests after #1645)

VisualMelon:
Thinking about #391, do we need to be able to measure the size of trimmed text? E.g. if something asks for 3 lines of text, but only has space for 2.5 lines, we may not want it allocating the space for 2.5 lines when it could allocate space for just 2.

Jonarw:
> Thinking about #391, do we need to be able to measure the size of trimmed text? E.g. if something asks for 3 lines of text, but only has space for 2.5 lines, we may not want it allocating the space for 2.5 lines when it could allocate space for just 2.

This may make sense. But to be consistent, this should then also apply in the horizontal direction, i.e. it should return the actual width (including ellipsis character) and not the `maxWidth` of trimmed text.

JimBobSquarePants:
I saw ImageSharp tagged here so I thought I better comment. Just a heads up. We're currently doing a massive overhaul of our Font library regarding text layout. https://github.com/SixLabors/Fonts/pull/198

As part of that the `TextMeasurer` result will change but in a positive way. We'll be providing more typographically accurate results and additional `FontMetrics` data to help supplement any measurements. Both Horizontal and Vertical text will be supported as well as bidirectional text. Any line breaking matches the [Unicode LineBreaking](http://www.unicode.org/reports/tr14/tr14-22.html) algorithm

VisualMelon:
@JimBobSquarePants thanks for the heads up; I've not had time to look into this PR or the ImageSharp provision properly for a while. Are we OK referencing the newer versions of ImageSharp? I was unclear whether the change in licence would be a problem for us as we are MIT (though it looks like you are back on Apache, so maybe that's a moot question now).

Also, thanks for the link to unicode line-breaking algorithm; I should probably have investigated that properly before writing all this stuff, but thankfully that bit should be easy to change if I or anyone else ever gets around to working on this again.

JimBobSquarePants:
>Are we OK referencing the newer versions of ImageSharp? 

Yep, we're Apache 2.0. Made the switch back for the RC.

Regarding line breaking. I'm not sure what the capabilities are for different libraries regarding line breaking. I think WPF has older implementations nailed but Skia does not have an implementation.

VisualMelon:
> Yep, we're Apache 2.0. Made the switch back for the RC.

Grand, thanks!

VisualMelon:
Rebase against develop.
I think I probably broke the ImageSharp implementation in the rebase, so will try to test and fix that before going any further.
Changed #if NETSTANDARD2_0_OR_GREATER to #if NETSTANDARD2_1_OR_GREATER in SkiaRenderContext because I couldn't get it to build otherwise; not sure what is up with that.

VisualMelon:
Should idealy review https://www.unicode.org/reports/tr14/tr14-40.html and http://www.unicode.org/reports/tr14/tr14-22.html to try to conform to unicode standards, but I don't think this is essential for this PR given there was no support for anything prior and it will probably be very difficult to write something standard complaint (if anyone knows better, please do speak up), so we should take our time over it.

VisualMelon:
Not sure why the test is failing

VisualMelon:
Failing on the XKCD examples, something to do with font.

VisualMelon:
New code is crashing because it insists on measuring the text when rendering, which the old code didn't. Not immediately sure how to deal with this. Should we be throwing if the font isn't available, or should the drawing routines trap the exception and use a default? Certainly need to be able to detect when the font is missing so we can use a consistent default if we go that way.

Example stack trace:

```
  Failed ExportPngAndCompareWithBaseline(Test #1) [53 ms]
  Error Message:
   System.InvalidOperationException : No glyph typeface found
  Stack Trace:
     at OxyPlot.Wpf.CanvasRenderContext.GetFontMetrics(String fontFamily, Double fontSize, Double fontWeight) in /_/Source/OxyPlot.Wpf/CanvasRenderContext.cs:line 463
   at OxyPlot.Rendering.TextArranger.MeasureText(String text, String fontFamily, Double fontSize, Double fontWeight) in /_/Source/OxyPlot/Rendering/TextArranger.cs:line 197
   at OxyPlot.Rendering.TextArranger.ArrangeText(ScreenPoint p, String text, String fontFamily, Double fontSize, Double fontWeight, Double rotation, HorizontalAlignment horizontalAlignment, VerticalAlignment verticalAlignment, Nullable`1 maxSize, HorizontalAlignment targetHorizontalAlignment, TextVerticalAlignment targetVerticalAlignment, String[]& lines, ScreenPoint[]& linePositions) in /_/Source/OxyPlot/Rendering/TextArranger.cs:line 78
   at OxyPlot.Wpf.CanvasRenderContext.DrawText(ScreenPoint p, String text, OxyColor fill, String fontFamily, Double fontSize, Double fontWeight, Double rotation, HorizontalAlignment horizontalAlignment, VerticalAlignment verticalAlignment, Nullable`1 maxSize) in /_/Source/OxyPlot.Wpf/CanvasRenderContext.cs:line 344
   at OxyPlot.XkcdRenderingDecorator.DrawText(ScreenPoint p, String text, OxyColor fill, String fontFamily, Double fontSize, Double fontWeight, Double rotate, HorizontalAlignment halign, VerticalAlignment valign, Nullable`1 maxSize) in /_/Source/OxyPlot/Rendering/RenderContext/XkcdRenderingDecorator.cs:line 133
   at OxyPlot.PlotModel.RenderErrorMessage(IRenderContext rc, String title, String errorMessage, Double fontSize) in /_/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs:line 155
   at OxyPlot.PlotModel.RenderOverride(IRenderContext rc, OxyRect rect) in /_/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs:line 134
   at OxyPlot.PlotModel.OxyPlot.IPlotModel.Render(IRenderContext rc, OxyRect rect) in /_/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs:line 34
   at OxyPlot.Wpf.PngExporter.ExportToBitmap(IPlotModel model) in /_/Source/OxyPlot.Wpf/PngExporter.cs:line 99
   at OxyPlot.Wpf.PngExporter.Export(IPlotModel model, Stream stream) in /_/Source/OxyPlot.Wpf/PngExporter.cs:line 73
   at OxyPlot.Wpf.PngExporter.Export(IPlotModel model, String fileName, Int32 width, Int32 height, Double resolution) in /_/Source/OxyPlot.Wpf/PngExporter.cs:line 62
   at OxyPlot.Wpf.Tests.ExampleLibraryTests.<ExportPngAndCompareWithBaseline>g__ExportAndCompareToBaseline|4_0(PlotModel model, String fileName, <>c__DisplayClass4_0& ) in /_/Source/OxyPlot.Wpf.Tests/ExampleLibraryTests.cs:line 52
   at OxyPlot.Wpf.Tests.ExampleLibraryTests.ExportPngAndCompareWithBaseline(ExampleInfo example) in /_/Source/OxyPlot.Wpf.Tests/ExampleLibraryTests.cs:line 64
```

JimBobSquarePants:
Text layout always involves measuring so that line breaks can be handled.

VisualMelon:
Aye, I think it can't be avoided, but we need to decide how to handle a missing font when rendering: previously it would just be passed through to WPF and it would default to a different font. My preference at the moment would probably be to implement that behaviour explicitly, so we handle falling back to a different font, maybe allowing the user to configure if they want it to throw in this case or not.

VisualMelon:
Pushed a change so that `CanvasRenderContext` uses the default font if it can't get font metrics for the time being.

VisualMelon:
Rebased. @Jonarw do you have any opinion on the issue of default fonts? (see recent comments above and most recent commit)

Jonarw:
I think falling back to the default font seems sensible. How do other renderers handle this? As there are no crashes in the examples I assume they also just fall back to the default?

VisualMelon:
Rebased against master: @Jonarw given how slowly things are moving at the moment, I'd suggest we (squash) and merge this, and give ourselves time to find any remaining issues with the understanding that this isn't necessary a completely fixed API yet.

If this massive PR gets out of the way, I might get up the courage to start re-writing the Axis abstractions per #1640.

Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -98,6 +108,23 @@ public CanvasRenderContext(Canvas canvas)
...
+        /// Gets or sets the thickness limit for "balanced" line drawing.
+        /// </summary>
+        public double BalancedLineDrawingThicknessLimit { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to use stream geometry for lines and polygons rendering.
+        /// </summary>
+        /// <value><c>true</c> if stream geometry should be used; otherwise, <c>false</c> .</value>
+        /// <remarks>The XamlWriter does not serialize StreamGeometry, so set this to <c>false</c> if you want to export to XAML. Using stream geometry seems to be slightly faster than using path geometry.</remarks>
+        public bool UseStreamGeometry { get; set; }
```
Jonarw, line 121: I don't think this is used anywhere?

Jonarw (commented):
I am sorry for my prolonged absence, this came during a quite busy time for me, and then I just forgot about it.
I agree with you and would be happy to have this merged! I just found one property that I think is a probably a remnant of an earlier version of `CanvasRenderContext`?

PR #1558, opened 2020-05-11
plotModel title clipping feature

### LINKED ISSUE #1510: Title of plot exceeding plotarea
If the title of the plot is longer than the plot-width the title exceeds the plotarea and covers other content (WPF).

UndisputedBasement33:
Fixes #1510 .

### Checklist

- added title clipping feature (analogue to axis title feature)
-
-

This feature changes the default behavior of the title, because the clipping length defaults to 90% of the title area. If this is not desired, we can set the value to 100%.

(I'm really sorry for the inconvenience I caused with my first pull request, I'll try hard to avoid this for the future)

@oxyplot/admins


Jonarw:
Looks good to me in general!
Unfortunately, WPF seems to be pretty much the only platform that supports the `maxSize` parameter, which isn't a problem with this PR, but more a general one. I will open a separate issue regarding about that, but I think this PR can be merged.

PR #1560, opened 2020-05-11
added LabelSpacing, MultiLabel and LabelStep properties to contours series

### LINKED ISSUE #1511: Label Control for Contour Series not working
If I create a contour series like described in the Oxyplot documentation and I want to manipulate the labels with LabelSpacing and LabelSteps there is no effect.

Sample code (using Oxyplot v2.0.0 in WPF):
```
var cs = new ContourSeries
{
        Color = OxyColors.Black,
        LabelBackground = OxyColors.White,
        ColumnCoordinates = yy,
        RowCoordinates = xx,
        LabelSpacing = 10,
        LabelStep = 2,
        Data = peaksData
};
```

My target was to show the label only on every 2nd contour, and I wanted to move the label a bit, so that when you use transparent background for the label, that it doesn't get strikethrough by the contour. But LabelSpacing and LabelSteps seem to not have any effect.

UndisputedBasement33:
Fixes #1511 .

### Checklist

- implemented LabelSpacing and LabelStep to contour series
- in order to not change default behaviour added property MultiLabel, to enable several labels per contour to make use of LabelSpacing
-

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Series/ContourSeries.cs
+++ b/Source/OxyPlot/Series/ContourSeries.cs
@@ -11,6 +11,7 @@ namespace OxyPlot.Series
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
     using System.Linq;
 
     /// <summary>
@@ -45,9 +46,9 @@ namespace OxyPlot.Series
         public ContourSeries()
         {
             this.ContourLevelStep = double.NaN;
-
-            this.LabelSpacing = double.NaN;
             this.LabelStep = 1;
+            this.MultiLabel = false;
+            this.LabelSpacing = 150;
             this.LabelBackground = OxyColor.FromAColor(220, OxyColors.White);
 
             this.Color = OxyColors.Automatic;
@@ -80,7 +81,7 @@ namespace OxyPlot.Series
 
         /// <summary>
         /// Gets or sets the contour level step size.
-        /// This property is not used if the ContourLevels vector is set.
+        /// This property is not used if the <see cref="ContourLevels"/> vector is set.
         /// </summary>
         /// <value>The contour level step size.</value>
         public double ContourLevelStep { get; set; }
@@ -118,13 +119,18 @@ namespace OxyPlot.Series
         public string LabelFormatString { get; set; }
 
         /// <summary>
-        /// Gets or sets the label spacing.
+        /// Gets or sets the label spacing, which is the space between labels on the same contour. Not used if <see cref="MultiLabel"/>==<see langword="false"/>
         /// </summary>
         /// <value>The label spacing.</value>
         public double LabelSpacing { get; set; }
 
         /// <summary>
-        /// Gets or sets the label step (number of contours per label).
+        /// Gets or sets a value indicating whether multiple labels should be displayed per Contour. The default value is <c>false</c>
+        /// </summary>
+        public bool MultiLabel { get; set; }
+
+        /// <summary>
+        /// Gets or sets the interval between labeled contours. LabelStep = 1 is default and it means that all contours have a label
         /// </summary>
         /// <value>The label step.</value>
         public int LabelStep { get; set; }
@@ -183,6 +189,7 @@ namespace OxyPlot.Series
                     double step = range / 20;
                     double stepExp = Math.Round(Math.Log(Math.Abs(step), 10));
                     actualStep = Math.Pow(10, Math.Floor(stepExp));
+                    this.ContourLevelStep = actualStep;
                 }
 
                 max = Math.Round(actualStep * (int)Math.Ceiling(max / actualStep), 14);
@@ -234,7 +241,7 @@ namespace OxyPlot.Series
                     {
                         result = r;
                         result.Text = StringHelper.Format(
-                            this.ActualCulture, 
+                            this.ActualCulture,
                             this.TrackerFormatString,
                             null,
                             this.Title,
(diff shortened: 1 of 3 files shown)


VisualMelon:
I'm still getting my head around this, but could you run a Style-Cop pass and fix the issues? It's little things like spaces at the start of comments, and expanding `if () continue;` onto multiple lines for clarity.

Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -123,6 +125,12 @@ public OxyColor ActualColor
         /// <value>The label spacing.</value>
         public double LabelSpacing { get; set; }
 
+        /// <summary>
+        /// Gets or sets if multiple labels should get displayed per Contour.
```
VisualMelon, line 129: nit: "get" -> "be"

Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -46,8 +47,9 @@ public ContourSeries()
         {
             this.ContourLevelStep = double.NaN;
 
-            this.LabelSpacing = double.NaN;
-            this.LabelStep = 1;
+            this.MultiLabel = false;
+            this.LabelSpacing = 150;
+            this.LabelStep = 0;
```
VisualMelon, line 52: Having `LabelStep == 0` seems to have some dedicated behaviour: could this be explained in the inline documentation?

Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -11,6 +11,7 @@ namespace OxyPlot.Series
 {
     using System;
     using System.Collections.Generic;
+    using System.Diagnostics;
```
VisualMelon, line 14: Not needed

Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -277,26 +285,98 @@ public override void Render(IRenderContext rc)
...
+                    else
                     {
-                        this.AddContourLabels(contour, transformedPoints, clippingRect, contourLabels);
+                        contourPartLengthTarget = contourFirstPartLengthTarget + j * this.LabelSpacing;
                     }
+
+                    //find index of contour points where next label should be positioned
+                    for (int k = intervalIndex; k < transformedPoints.Length; k++)
+                    {
+                        contourPartLength += Math.Sqrt(Math.Pow(transformedPoints[k].x - transformedPoints[k - 1].x, 2) + Math.Pow(transformedPoints[k].y - transformedPoints[k - 1].y, 2));
```
VisualMelon, line 368: Could be 

                        contourPartLength += (transformedPoints[k] - transformedPoints[k - 1]).Length;

For clarity

UndisputedBasement33:
Hi,
thanks for reviewing. I run through style cop and took care about your other points.

Jonarw:
I am also not too familiar with the `ContourSeries`, but something does not seem completely right to me.

From the the way the `LabelStep` property is described, I would expect that it it set `LabelStep=2`, every second contour will have a label. This is true as long as the `ContourLevelStep==1` (which is the case for all examples in the ExampleLibrary).

But if I set e.g. `ContourLevelStep==0.5`, I will only have a label every fourth contour (and something weird happens for contours with level > 4):
image

From my understanding, in this case I should still have labels every second contour, so for contour values of 1, 2, 3 and so on. Would you agree @UndisputedBasement33 ? Or am I misunderstanding something about this series, having never worked with it?

UndisputedBasement33:
Thanks for the comment. I fixed the problem with the contours for levels > 4. I think the code is better now. I wanted too achieve a "loose" matching that's why I also added one more example with ContourLevelStep of 0.33. That means if you define LabelStep = 2 but the contour level is not exactly 2 but maybe 1.98 or 2.02 it should still show the label.

I changed the description for LabelStep, but I honestly struggle a bit to describe it precisly. Before I was relying on the Matlab describtion, which was - like you correctly mentioned - misleading. The feature should be analogue to the ContourLevelStep. With ContourLevelStep you define how many contours you want to have by defining the interval between the levels. And with LabelStep you define the interval between the levels for the labels. For example LabelStep 2 means only contours with a level which is an integral multiple of 2 will get a label. Do you understand now?

Should I again squash the commits or is it better for you to review, if I leave it like it is now?

Jonarw:
Thanks for the explanation, I think I understand better now how this works.
I do however still have some concerns. What if the user for some reason wants to use a non-integer `ContourLevelStep`? With `ConturLevelStep==Math.PI/2` and `LabelStep==1` I get no labels at all, except at 0:
image

I see two possibilities how this could be solved:
1. Let the `LabelStep` not define an interval in data space, rather let it just count contours. So e.g. for `LabelStep==2`, show a label on every second contour, regardless of the `ContourLevelStep`
2. Make the `LabelStep` a double and allow fractional label steps

Personally I think I would prefer possibility 1, as it seems more intuitive to me. But as I said I do not have too much experience with this kind of series, so maybe this would be different to others.

UndisputedBasement33:
I understand your concerns. I used the first option at the beginning, and it worked nicely if the contours are next to each other. But since the foreach-loop loops more or less randomly through the contours it wasn't necessarily the case that the next contour in the loop is also the contour that the user would define as next contour in the plot. Because of that out of every 2nd, it became a bit random.
But now I think I found a better way. I went back with the definition to what @Jonarw understood at first and what he suggested with his 1st option. Now `LabelStep==2` means every 2nd contour, `LabelStep==3` every 3rd contour etc..
Please let me know what you think about it

P.S: I have a failed check (AppVeyor was unable to build non-mergeab...). For me the solution builds fine, what can I do to fix that?

UndisputedBasement33:
I fixed the issues commented by @Jonarw . I squashed all commits and rebased it on the latest develop branch. I hope everything is done correctly. No all the checks have passed.

Jonarw:
Thanks for taking care of this! 
Looks mostly good to me now, I just noticed there seems to be a slight issue with the `MultiLabel` algorithm. When I zoom in on one of the contours, it looks like this:
image

UndisputedBasement33:
Yes this wasn't nice, glad you found it. Took me a bit to figure out, but it should be fixed now.

Jonarw:
Looks good to me now!

PR #1565, opened 2020-05-16
Implement native clipping for SvgRenderContext (#1564)

### LINKED ISSUE #1564: Native clipping for SvgExporter
It would be nice if `OxyPlot.SvgRenderContext` could use native SVG clipping using `clip-path`, similar to how the `OxyPlot.SkiaSharp` SVG exporter does it. The functionality seems to be already there in `SvgWriter`, so it seems almost trivial to support this.

Is there any downside why we would not want this?

`SvgRenderContext` is, as far as I know, the only remaining render context without native clipping. If we can support clipping here we could even think about completely removing the complex code path that is there to support renderers without native clipping - but this would be a separate issue.

Jonarw:
Fixes #1564.

### Checklist

- Enable native clipping for SvgRenderContext
- I slightly changed the implementation of clipping in `SvgWriter` (inspired by SkiaSharp SVG output), as the existing implementation did not work well with text
- Added a clipping example to the Rendering Capabilities

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
+++ b/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
@@ -16,7 +16,6 @@ namespace ExampleLibrary
     using OxyPlot;
     using OxyPlot.Annotations;
     using System.Linq;
-    using System.Collections.Generic;
 
     /// <summary>
     /// Provides rendering capability examples.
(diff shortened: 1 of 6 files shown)


Source/OxyPlot/Svg/SvgRenderContext.cs
```diff
@@ -262,11 +262,24 @@ public override OxySize MeasureText(string text, string fontFamily, double fontS
             this.w.WriteImage(srcX, srcY, srcWidth, srcHeight, destX, destY, destWidth, destHeight, source);
         }
 
+        /// <inheritdoc/>
+        public override bool SetClip(OxyRect rect)
+        {
+            this.w.BeginClip(rect.Left, rect.Top, rect.Width, rect.Height);
```
VisualMelon, line 268: Does this need to call `ResetClip` if we are currently clipping?

Source/OxyPlot/Svg/SvgRenderContext.cs
```diff
@@ -262,11 +262,24 @@ public override OxySize MeasureText(string text, string fontFamily, double fontS
...
+            return true;
+        }
+
+        /// <inheritdoc/>
+        public override void ResetClip()
+        {
+            this.w.EndClip();
```
VisualMelon, line 275: Should this (or `EndClip` itself) verify we are currently clipping?

Jonarw:
You are right of course, now I feel a bit silly for not thinking of this...

VisualMelon:
It only occurred to me that it was an issue when I was reading your original SkiaSharp implementation ;)

Jonarw:
I also added a basic test for this, and a `IRenderContext.AutoResetClip` extension because I thought that might come in handy.

VisualMelon:
Looks good to me.

Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
```diff
@@ -17,6 +17,7 @@ namespace ExampleLibrary
     using OxyPlot.Annotations;
     using System.Linq;
     using System.Collections.Generic;
+    using OxyPlot.Axes;
```
VisualMelon, line 20: Is this needed?

Jonarw: No, in fact the one above isn't either

Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
```diff
@@ -458,6 +457,59 @@ public static PlotModel DrawTextWithMetrics(string text, string font, double fon
             return model;
         }
 
+        [Example("Clipping")]
+        public static PlotModel Clipping()
+        {
+            var model = new PlotModel();
+            model.Annotations.Add(new DelegateAnnotation(rc =>
+            {
+                // reset a couple of times without any clipping being set to see that nothing explodes
```
VisualMelon, line 466: Ideal

Issue #729, opened 2016-01-12
ColumnSeries / BarSeries not working with more than one value-axis


benjaminrupp:
The properties currentBarOffset, currentPositiveBaseValues and currentNegativeBaseValues are not dependend on the value axis of the series. Hence they are not working if more than one value axis is used in a chart.

See the following example here:

![unbenannt](https://cloud.githubusercontent.com/assets/16608319/12258628/56f549d2-b910-11e5-9214-93436e6637ca.PNG)

tibel:
May you provide some sample code that shows the issue, please.

benjaminrupp:
I implemented a solution for the above mentioned bug. I was wrong in my first post. Only the currentXXX-properties have to be dependend on the value axis.
Please find the code in my pull request.

benjaminrupp:
I made a second pull request to replace my first one cause there was an error in my first implementation.

objorke:
Is this a workaround since we don't have subplots? #55
I think this should not be implemented in the CategoryAxis, but in the BarSeries/ColumnSeries.
The CategoryAxis should not be responsible for functionality of a series. I think the bar offsets and min/max values could be moved to the BarSeries/ColumnSeries. See #741

objorke:
I think this should not be implemented in the CategoryAxis, but in the BarSeries/ColumnSeries.
The CategoryAxis should not be responsible for functionality of a series. I think the bar offsets and min/max values could be moved to the BarSeries/ColumnSeries. See #741

Issue #388, opened 2015-02-21
PostRender or PostFirstRender event


tr8dr:
For an application, I need to get a view on rendered sizes post 1st render.  The proposal is to instrument the rendering function in each PlotView implementation with an event:

```csharp
/// <summary>
/// Event is called post-rendering, for each render cycle.  This event will be called at high-frequency,
// so take care to minimize work on callback.
/// </summary>
public event EventHandler PostRender;
```
Or alternatively, a one-shot event, to capture the 1st render:
```csharp
/// <summary>
/// Event is called after the 1st render and thereafter not called again.
/// </summary>
public event EventHandler PostFirstRender;
```

Here is a practical example of how am using.   This class aligns panes vertically (and synchronizes the X axis) by latching on to both the axis change events and the first render event:

```csharp
	/// <summary>
	/// Aligns OxyPlot PlotViews vertically:
	/// 
	/// - X axis is coordinated
	/// - X margin for Y axis is negotiated so that all panes have enough space for y-axis rendering
	/// - X dimension of panes must be the same size (can use a table or fixed size to ensure)
	/// </summary>
	public class AlignedVerticalPaneController
	{
		public AlignedVerticalPaneController ()
		{
			Panes = new List<PlotView> ();
			SpanSlack = 5;
		}

		// Properties

		/// <summary>
		/// List of vertically coordinated panes (the 1st of which is assumed to be the master)
		/// </summary>
		public List<PlotView> Panes { get; private set; }

		/// <summary>
		/// Amount of extra slack to give span
		/// </summary>
		public double SpanSlack { get; set; }

		// Functions

		/// <summary>
		/// Adds a pane to be managed.
		/// </summary>
		/// <param name="pane">Pane.</param>
		public void AddPane (PlotView pane)
		{
			var xaxis = OxyUtils.XAxisFor (pane);
			xaxis.AxisChanged += (sender, e) => XAxisChanged (pane, xaxis);

			foreach (var axis in pane.Model.Axes)
			{
				if (axis.IsVertical())
					axis.AxisChanged += (sender, e) => OnYAxisChange (pane, axis);
			}

			Panes.Add (pane);
			pane.PostFirstRender += (sender, epane) => InitialRender(epane);
		}


		/// <summary>
		/// Removes a pane from the controller
		/// </summary>
		/// <param name="pane">Pane.</param>
		public void DelPane (PlotView pane)
		{
			Panes.Remove (pane);
		}

 		/// <summary>
[...]

PR #1572, opened 2020-05-21
Refactor Bar Series (#741)

### LINKED ISSUE #741: CategoryAxis should not contain rendering information about BarSeries
the functionality related to stacked bar series should be moved out of CategoryAxis

Jonarw:
Fixes #741 and #729.

### Checklist

- Extract functionality related to `BarSeries` from `CategoryAxis` into new `BarSeriesManager` class
- Refactor bar series:
    - Change `CategorizedSeries` (a relict from the time when we had a separate `ColumnSeries`) to `BarSeriesBase<T>`
    - Move a lot of duplicated code from `BarSeries`, `TornadoBarSeries` and `IntervalBarSeries` into `BarSeriesBase<T>`
- My goal was to make the `BarSeriesManager` completely transparent to the `PlotModel`. This led to the somewhat strange `CheckIn...()` methods, which `BarSeriesBase` calls during updating and rendering. I could not think of a better solution for this, if someone else has one, please share :)
- I removed `Axis.UpdateFromSeries(...)` and `Series.UpdateValidData()`, as these seemed to be there just for the bar series and are not needed anymore

@oxyplot/admins


VisualMelon:
There seems to be a fairly serious regression: when there are disabled series (e.g. checking them off the legend in "All in one"), the plot goes completely insane when you pan/zoom/do anything.

Jonarw:
Yeah... thanks for noticing!
This in the end is a symptom of the weird interaction between series and manager... I guess I'll have to come up with a better way of doing this

Jonarw:
The latest commit fixes the above mentioned bug.

Outstanding concerns:

- If a bar series never has been updated, it will not render until the plot is updated with `updateData==true`. This occurs for example when enabling a bar series that was never visible by clicking on its legend entry (e.g. Series 3 in the All-in-One example). 
However this is consistent with the behavior before this PR and not necessarily something that is the fault of the series itself.
- I still don't really like the way the `BarSeriesBase` and the `BarSeriesManager` interact. However I can't think of a better alternative and I think that's the best we can do without involving the `PlotModel`, which was one of my goals with this.

VisualMelon:
It still seems to be possible to confuse stacked bar graphs by turning series visibility on and off, though panning/zooming seems to fix it.

image

It also seems that if a series 'starts' invisible (e.g. series 3 in `AllInOne`) that it doesn't display when you check it on or off on the legend. This isn't changed by zooming/panning.

Jonarw:
Yes, both of these issues in the end come down to the fact that the bar offsets are determined by `BarSeriesManager` during plot update. When you change the visibility of a graph without updating the plot (as is done when you click on the legend entry), it can become confused.

VisualMelon:
Right. I won't open that issue immediately, because it does seem to be specific to `BarSeries`, and I'd like to properly review your code before diving into the old stuff on my own.

Jonarw:
(rebased)
I took another look at the code in this PR and I am not really sure what to make of it.
Sure, the current design where the `CategoryAxis` is responsible for telling the `BarSeries` where to render is 'not ideal', and this PR definitely improves on that with the introduction of the `BarSeriesManager` class. But I still absolutely don't like how `BarSeriesManager`s are created by `BarSeriesBase` itself (who creates their own manager??) and how the `BarSeriesManager` kind of has to guess when all of its series have been updated / rendered.

This is kind of a consequence of my goal to make the `BarSeriesManager` completely transparent to the `PlotModel`. I think if we want to keep it that way, this is pretty much the best we can do. But maybe this goal is flawed to start with?

So I think we have four possibilities:
1. Merge this PR like it is (after some more testing and bug fixing), it works well enough for the moment
2. Change it such that the `BarSeriesManager` is managed by the `PlotModel`. `IPlotModel.Update()` would maybe call a dedicated method `UpdateBarSeries()`, where all managers are set up and updated. Similarly, `IPlotModel.Render()` would call a dedicated method `RenderBarSeries()`.
3. Solve this problem some other way I can't think of
4. Forget this PR ever existed, maybe someone else can take a shot at it in the future and figure out a better design

Input is appreciated :)

VisualMelon:
@Jonarw sorry, completely forgot to look at this.

The bit that looks most dodgy to me is the observation that the manager has to guess at when to do its own updates: this could be solved by adding it to a list of things that PlotModel 'updates' once it has done everything else. Another set that are updated before everything else would help with things like slaving axes.

This could just be done with events (see. https://github.com/oxyplot/oxyplot/issues/388#issuecomment-75500509).

Or have I missed the point?

Jonarw:
I have also thought about triggering updates of the manager by events.
The problem that remains in that case: When is the manager created in the first place? And when is it subscribed to the relevant PlotModel event? 
At the current status of this PR, the manager is created by the BarSeries itself (by the first one that gets updated), but I guess we would also want to avoid this... So we'd have to create it somewhere in the `PlotModel` and we are back to what I would have liked to avoid.

VisualMelon:
I don't think it's a problem for a `BarSeries` to create its manager. One could argue that's the only way that makes any sense, because nobody else necessarily knows when a `BarSeries` is added. A bigger issue in my eyes is that the `BoxSeries` have to keep track of its manager. It also feels like `BarSeriesManager` has too much work in not only managing the bar series but deciding which bar series to manage and when to perform these managerial tasks.

What is the issue with having `CategoryAxis` do this? (i.e. it's good to have `BarSeriesManager`, but could it be held and invoked by `CategoryAxis`?). If not the axes themselves, then I think there needs to be a single objects per `PlotModel` to manage the `BarSeriesManager` instances (which could be renamed to `BarSeriesArranger`, or w/e), because anything else will create a lot of additioanl complexity (complexity which current exists in `BarSeriesManger` but will be hard to test and I don't trust). [Original draft with some additional musings/insanity](https://gist.github.com/VisualMelon/6849a2ac53ec6c603274b193c576f42a).

Jonarw:
Thanks for your input! It really helps to hear another perspective!

> What is the issue with having `CategoryAxis` do this?

I would have two issues with this:

- `CategoryAxis` should not assume it is only ever used for `BarSeries`, there could also be other types of series that use a `CategoryAxis` (in fact at work I use one of these). So I don't think there should be any logic related to `BarSeries` in `CategoryAxis`.
- There can be multiple `BarSeriesManager`s with the same `CategoryAxis` (where multiple value axes are present)

As a concept, I changed the implementation such that the `PlotModel` now manages the `BarSeriesManager`s. This for sure is not a perfect solution, but more a pragmatic one. At the very least it's better than what I had before and has the nice side effect that it seems to handle invisible series fairly well (though it's not fully tested yet).

If we encounter other similar issues in the future we could still add something like the `IService` concept @VisualMelon proposed, but for just this one application this might be overkill?

VisualMelon:
>  There can be multiple BarSeriesManagers with the same `CategoryAxis`

Good point. I'd completely overlooked that possibility.

> but for just this one application this might be overkill

Agreed. We can work out a suitable abstraction later when we have more than one example from which to work (though I was wondering if we could abstract out the layout code: it might make `SubPlots` and e.g. axis slaving easier to bodge without modifying the library).

Jonarw:
Did some more testing, added an example with non-stacked `BarSeries` and multiple value axes and fixed an issue with the Transpose() function, which did not work properly for series that are invisible initially (such as in the All-in-one example).

Everything seems to work well enough to me now. Fancy taking another look @VisualMelon ?

Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
```diff
@@ -301,24 +301,38 @@ public static PlotModel Transpose(this PlotModel model)
 
             foreach (var series in model.Series.OfType<XYAxisSeries>())
             {
-                if (series.XAxis != null && series.XAxisKey == null)
+                if (series.XAxisKey == null)
                 {
-                    if (series.XAxis.Key == null)
+                    if (series.XAxis == null) // this can happen if the series in invisible initially
```
VisualMelon, line 306: Typo: `is`

VisualMelon:
Nothing jumps out at me, and I'll try to take a proper look in the next couple of days.

VisualMelon:
Could do with an example with more than one CategoryAxis, e.g. https://gist.github.com/VisualMelon/a31c6a4f8d4388548f357d51b0615bd0

I can't find any unexpected behaviour, but I've not looked at the code yet.

Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
```diff
@@ -301,24 +301,38 @@ public static PlotModel Transpose(this PlotModel model)
 
             foreach (var series in model.Series.OfType<XYAxisSeries>())
             {
-                if (series.XAxis != null && series.XAxisKey == null)
+                if (series.XAxisKey == null)
```
VisualMelon, line 304: Would it be possible to simplify this by calling `EnsureAxes` on the series before otherwise interrogating it?

Source/OxyPlot/Series/BarSeries/ErrorBarSeries.cs
```diff
@@ -53,7 +53,7 @@ protected internal override void UpdateMaxMin()
             base.UpdateMaxMin();
 
             //// Todo: refactor (lots of duplicate code here)
```
VisualMelon, line 55: Is this still needed?

Source/OxyPlot/Series/BarSeries/IBarSeries.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// Gets the <see cref="ValueAxis"/> the bar series uses.
+        /// </summary>
+        Axis ValueAxis { get; }
+
+        /// <summary>
+        /// Updates the valid data.
+        /// </summary>
+        void UpdateValidData();
+
+        // TODO: replace the following members by IReadOnlyList<BarItemBase> once we get rid of .NET 4.0.
```
VisualMelon, line 52: There was a specific request to keep Net Framework 4.0 support quite recently (or atleast it was readded in #1452), so I'm not sure if these comments are useful.

Would it be insane to provide our own type to expose these members so that we don't have the interface splatted into other types', making a future refit painless for everyone (e.g. just change `CustomReadOnlyListWrapperType` with `IReadOnlyList` and update the writes: minimal code change for consumers). This type could even implement `IReadOnlyList` for compilation targets that support it, keeping conditions in one place. Are there other places we could use such a type, or is it just a barmy idea?

Source/OxyPlot/Utilities/ArrayExtensions.cs
```diff
@@ -9,11 +9,36 @@
...
     /// Provides useful extension methods for arrays.
     /// </summary>
     public static class ArrayExtensions
     {
+        /// <summary>
+        /// Finds the maximum value in the sequence, or returns 0 if the sequence is empty.
+        /// </summary>
+        /// <param name="sequence">The sequence.</param>
+        /// <returns>The maximum value of the sequence, or 0 if the sequency is empty.</returns>
+        public static double MaxOrZero(this IEnumerable<double> sequence)
```
VisualMelon, line 25: Seems a tad specific. If it took a default value as a parameter that would be nice (the deafult value is a concern for the calling domain), but I'm not sure it's worth the effortof coming up with a different name.

Source/OxyPlot/Utilities/ArrayExtensions.cs
```diff
@@ -9,11 +9,36 @@
...
+        {
+            using var e = sequence.GetEnumerator();
```
VisualMelon, line 27: `null` check would be nice on `sequence`

VisualMelon:
There is a change of behaviour with regard to shared values axis: if 2 bar series share a category series but have separate value axes, then they would previously be grouped together. The new behaviour makes sense if the value axes are 'seperated' (as in the `MultipleValueAxes` example), but not if they 'overlap' (e.g. share a common zero position).

I don't think the old behaviour was really supportable or intentional given the the lack of support for axis slaving and the fact that the min/max go wrong if not manually specified - so I don't think this can be considered a regression -but I wonder if it would be a useful feature. The current system can't support it because it associates each manager with a single axis pair.

Source/OxyPlot/Series/BarSeries/ErrorBarSeries.cs
```diff
@@ -110,17 +110,8 @@ protected internal override void UpdateMaxMin()
                 }
             }
 
-            var valueAxis = this.GetValueAxis();
-            if (valueAxis.IsVertical())
```
VisualMelon, line 114: Was this a hold-over from when Column/Bar series were separate concerns?

Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
Jonarw: I like the idea in general, but this is not available here to use, as it is `protected internal`...

VisualMelon: Yep, sorry.

Source/OxyPlot/Series/BarSeries/ErrorBarSeries.cs
Jonarw: I would argue yes. The `UpdateMaxMin()` implementation of `ErrorBarSeries` is quite identical to the implementation of `BarSeries`, which I think this comment refers to.

Source/OxyPlot/Series/BarSeries/IBarSeries.cs
Jonarw: _(I do not quite agree on supporting .NET 4.0 for much longer and I hate that it forces us to do these kinds of considerations which should be a non-issue nowadays, but that's not the issue here.)_

I think there are quite some places where `IList<T>` is used that should use `IReadOnlyList<T>` instead. In this specific case this was not possible because of lack of covariance.

Possibly slightly insane idea: How about we define `IReadOnlyList<T>` (and maybe also `IReadOnlyCollection<T>`) ourselves (conditionally compiled only for NET40) and a custom list class implementing this (again only for NET40)? 
I'll do some experiments to see if this could even remotely work.

Source/OxyPlot/Utilities/ArrayExtensions.cs
Jonarw: Maybe we could just change it to `MaxOrDefault<T>(...)` and return `default(T)` if the sequence is empty?

Source/OxyPlot/Series/BarSeries/ErrorBarSeries.cs
Jonarw: Yes, exactly

Source/OxyPlot/Series/BarSeries/IBarSeries.cs
VisualMelon: I'd be hesitant to use the same names, because the 4.0 library could be used by 4.5 code and that might create confusion, but it would be nice to replace all the horridness with `IList<T>` with some well-typed horridness that can be quickly and easily removed if/when 4.0 support is dropped.

Source/OxyPlot/Utilities/ArrayExtensions.cs
VisualMelon: The issue is that in either case the default value for e.g. doubles is meaningful (i.e. it's not something like `null` which explodes if you use it). I'd suggest that the default should be provided by the calling code, because only it can know if there is a meaningful default (if there isn't, it should be using `Max` and throwing or something like `TryGetMax`). I don't think it's a huge issue, but providing a default default (heh) of zero makes the method limited and easy to misuse with `structs`.

I don't think this is a big issue, so feel free to ignore it, but it feels odd for a `public` method.

Source/OxyPlot/Series/BarSeries/BarSeries.cs
```diff
@@ -682,11 +357,90 @@ protected Axis GetValueAxis()
...
+                var baseValue = double.NaN;
+                if (this.IsStacked)
+                {
+                    baseValue = this.Manager.GetCurrentBaseValue(stackIndex, categoryIndex, value < 0);
+                }
+
+                if (double.IsNaN(baseValue))
+                {
+                    baseValue = this.BaseValue;
+                }
```
VisualMelon, line 391: This isn't new in this PR, but it's a bit weird.

VisualMelon:
I've not looked at all the removed code, but all the new code looks good (and surprisingly simple) to me (apart from the minor concerns discussed above).

Source/OxyPlot/Series/BarSeries/IBarSeries.cs
Jonarw: Of course we could also use this [package](https://www.nuget.org/packages/ReadOnlyCollectionInterfaces/) ([GitHub link](https://github.com/mausch/ReadOnlyCollections)) for NET4.0. 

You are right, it would not be possible to use the 4.0 library from 4.5 (it would complain when loading the package about ambiguous types I think), but is this a problem?

VisualMelon: Thanks for the link: I went looking for that project but couldn't find it. That said, I think it's probably too much for our purposes.

It shouldn't be ambiguous so long as the fully qualified name is distinct (which I think was the idea); I just think it could create confusion for users, and that being up-front that this is a thin wrapper over what is usually an array (I think) is a good idea. I suppose we need to decide whether it is important to improve the 4.0 interface, or just make life easier for everybody else. Something as simple as https://gist.github.com/VisualMelon/d8baadc9f28f8d5f1bf60bf9eaf8c318 might suffice.

Jonarw: Well this package just contains the three interfaces `IReadOnlyCollection<T>`, `IReadOnlyList<T>` and `IReadOnlyDictionary<T>`, so I wouldn't necessarily say it's too much ;).
However I realized we can't use this anyway, since the assembly is not signed, so doesn't matter anyway. If we would want to go this way, we would have to define the interfaces ourselves (which wouldn't be too difficult).

Your solutions seems neat, especially the implicit cast from `List<T>`. However there is one thing I am a bit concerned about: When we ever finally get rid of .NET 4.0, this will be a breaking change for all interfaces using `ReadOnlyList<T>` (as they will be changed to `IReadOnlyList<T>`). Ok granted, probably this change will not be noticed by most consumers of the API, so maybe this is the best option we have...

VisualMelon: Yeah, maybe I'm being silly, and we should just provide a conditionally defined `IReadOnlyList<T>` along with a wrapper type to re-expose `IList<T>`.

Jonarw: I think this topic warrants its own issue, so I'll open a new one.

VisualMelon: Good idea, and I don't think it should block this being merged.

Jonarw: Well I don't mind leaving this open for a couple more days until we have figured out what to do rather than going back later and adapting this to whatever solution we settle on.

Jonarw:
(rebased on #1601)
As discussed this now shows #1601 'in action' and makes the `IBarSeries` interface much nicer.

Source/OxyPlot/Utilities/ArrayExtensions.cs
Jonarw: Replaced this by `MaxOrDefault` with a user-defined default value.
Also added a corresponding `MinOrDefault` because it would feel weird otherwise.

Source/OxyPlot/Utilities/ArrayExtensions.cs
Jonarw: Replaced this by `MaxOrDefault` with a user-defined default value.
Also added a corresponding `MinOrDefault` because it would feel weird otherwise.

Source/OxyPlot/Utilities/ArrayExtensions.cs
Jonarw: Sorry, wanted to delete the accidentally misplaced comment, and for some reason it deleted your comment regarding the null check??

Anyway, I've added the null check.

VisualMelon:
Is there anything else you are planning to do with this or that you want another pair of eyes over?

Jonarw:
No, from my side I think this is pretty much done, except of course you (or anyone else) finds more issues.

(And by the way, I am not planning to squash the commits as I recently discovered GitHub's 'Squash and merge' feature. I think it is neat to preserve the full commit history on the original PR while still having a 'clean' history on the main branch.)

Jonarw:
Any objections to merging at this point?
@VisualMelon

VisualMelon:
None here.

PR #1576, opened 2020-05-24
Fix #1573 by upgrading SkiaSharp version, add test

### LINKED ISSUE #1573: Disposing a SkiaRenderContext can mess up fonts from another SkiaRenderContext instance
The following test fails at the last Assert:

        [Test]
        public void TestDisposal()
        {
            var plotModel = (IPlotModel)ShowCases.CreateNormalDistributionModel();
            ((PlotModel)plotModel).Background = OxyColors.White;
            ((PlotModel)plotModel).DefaultFont = "Arial";
            plotModel.Update(true);

            string ExportToBitmap(string fileName, SkiaRenderContext context)
            {
                using var bitmap = new SKBitmap(800, 600);
                using var canvas = new SKCanvas(bitmap);

                context.SkCanvas = canvas;
                canvas.Clear(plotModel.Background.ToSKColor());
                plotModel.Render(context, new OxyRect(0, 0, 800, 600));

                var path = Path.Combine(this.outputDirectory, fileName + ".png");
                using var fileStream = File.OpenWrite(path);
                using var skStream = new SKManagedWStream(fileStream);
                SKPixmap.Encode(skStream, bitmap, SKEncodedImageFormat.Png, 0);

                return path;
            }

            using var renderContext = new SkiaRenderContext() { RenderTarget = RenderTarget.PixelGraphic };
            var path1 = ExportToBitmap("export1", renderContext);

            string path2;
            using (var renderContext2 = new SkiaRenderContext() { RenderTarget = RenderTarget.PixelGraphic })
            {
                path2 = ExportToBitmap("export2", renderContext2);
            }

            var path3 = ExportToBitmap("export3", renderContext);

            Assert.IsTrue(File.ReadAllBytes(path1).SequenceEqual(File.ReadAllBytes(path2)));
            Assert.IsTrue(File.ReadAllBytes(path1).SequenceEqual(File.ReadAllBytes(path3)));
        }

export1.png (identical to export2.png, only relevant part):
image

export3.png (only relevant part) - font family, font weight and text shaping is broken:
image

- The issue persists regardless of `UseTextShaping==true` or not
- When `renderContext2` is disposed, it also disposes its cached fonts, see:
https://github.com/oxyplot/oxyplot/blob/7329013d6f48331e42b81c4315cd13fb5a82bb8f/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs#L473)
- Disposing the fonts seems to also destroy the native resources for the `SKFont` instances of `renderContext1`
- This may be a bug on SkiaSharp side, or just a misuse on our side. Maybe we are not supposed to cache `SKFont`s? (even though I read somewhere that you should do this for performance)
- I will investigate this some more and possibly create an issue over at SkiaSharp if I can't make sense of it

Jonarw:
Fixes #1573 .

### Checklist

- Upgrade SkiaSharp from 1.68.1.1 to 1.68.3. This fixes the issue.
- Add a test that reproduced the issue with the older SkiaSharp version

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.SkiaSharp.Tests/PngAssert.cs
+++ b/Source/OxyPlot.SkiaSharp.Tests/PngAssert.cs
@@ -1,0 +1,140 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="PngAssert.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Provides assertions on image files.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.SkiaSharp.Tests
+{
+    using global::SkiaSharp;
+    using NUnit.Framework;
+    using OxyPlot.SkiaSharp;
+    using System.IO;
+
+    /// <summary>
+    /// Provides assertions on image files.
+    /// </summary>
+    public static class PngAssert
+    {
+        /// <summary>
+        /// Determines if the images at the specified paths are equal.
+        /// </summary>
+        /// <param name="expected">Path to the expected image.</param>
+        /// <param name="actual">Path to the actual image.</param>
+        /// <param name="message">The message.</param>
+        /// <param name="output">The output difference file.</param>
+        public static void AreEqual(string expected, string actual, string message, string output)
+        {
+            var expectedImage = LoadImage(expected);
+            var actualImage = LoadImage(actual);
+
+            if (expectedImage == null)
+            {
+                EnsureFolder(expected);
+                File.Copy(actual, expected);
+                Assert.Fail("File not found: {0}", expected);
+            }
+
+            if (expectedImage.GetLength(0) != actualImage.GetLength(0))
+            {
+                Assert.Fail("Expected height: {0}\nActual height:{1}\n{2}", expectedImage.GetLength(0), actualImage.GetLength(0), message);
+            }
+
+            if (expectedImage.GetLength(1) != actualImage.GetLength(1))
+            {
+                Assert.Fail("Expected width: {0}\nActual width:{1}\n{2}", expectedImage.GetLength(1), actualImage.GetLength(1), message);
+            }
+
+            var w = expectedImage.GetLength(0);
+            var h = expectedImage.GetLength(1);
+            var differences = 0;
+            var differenceImage = new OxyColor[w, h];
+            for (var i = 0; i < h; i++)
+            {
+                for (var j = 0; j < w; j++)
+                {
+                    if (!expectedImage[j, i].Equals(actualImage[j, i]))
+                    {
+                        differences++;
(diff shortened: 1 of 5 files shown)


Source/OxyPlot.SkiaSharp.Tests/PngAssert.cs
```diff
@@ -0,0 +1,140 @@
...
+                    Path.GetFullPath(actual),
+                    Path.GetFullPath(output));
+            }
+        }
+
+        /// <summary>
+        /// Ensures that the folder for the specified path exists.
+        /// </summary>
+        /// <param name="path">The path to a file.</param>
+        private static void EnsureFolder(string path)
```
VisualMelon, line 104: I'd prefer a name which indicates we are creating the directory for the given file, rather than ensuring the given directory exists. Something like `EnsureDirectoryForFile` or something, I don't have any especially good suggestions.

Err... just realised that boat sailed in 2014. Nevermind ;)

Source/OxyPlot.SkiaSharp.Tests/PngAssert.cs
```diff
@@ -0,0 +1,140 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="PngAssert.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Provides assertions on image files.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Wpf.Tests
```
VisualMelon, line 10: Wrong namespace

Source/OxyPlot.SkiaSharp.Tests/SkiaRenderContextTests.cs
```diff
@@ -0,0 +1,66 @@
...
+                using var canvas = new SKCanvas(bitmap);
+
+                context.SkCanvas = canvas;
+                canvas.Clear(SKColors.White);
+                plotModel.Render(context, new OxyRect(0, 0, 800, 600));
+
+                var path = Path.Combine(this.outputDirectory, fileName + ".png");
+                using var fileStream = File.OpenWrite(path);
+                using var skStream = new SKManagedWStream(fileStream);
+                bitmap.Encode(skStream, SKEncodedImageFormat.Png, 0);
```
VisualMelon, line 46: Doesn't seem to exist in the old version of skia, but the test looks fine ;)

Source/OxyPlot.SkiaSharp.Tests/PngAssert.cs
Jonarw: Yeah... I pretty much just copy&pasted this file, so I'd suggest we leave this as is for now. Maybe we can find a way in the future to share this with `OxyPlot.Wpf.Tests`.

Jonarw:
(fixed namespace and squashed)

VisualMelon:
Looks good to merge to me.

PR #1578, opened 2020-05-25
Improve auto plot margins (#1577)

### LINKED ISSUE #1577: Auto plot margins sometimes reserves space that is not really needed
Since #1562, auto plot margins reserve space for axis labels. This can lead to fairly large margins when longer strings are used for axis labels, as in this example:
image

The empty space to the right of the plot is needed because the user could pan the category axis like this:
image

During the implementation of #1562, I experimented with making the auto margins adapt to the actual position of the axis labels as to not unnecessarily reserve space when it is not needed. However this lead to very counter-intuitive behavior during panning, so I decided to always reserve the space that could be needed by all the axis labels during panning. This is the behavior we have in the current `develop` version.

However, if the axis range is fixed and can't be changed by the user, we don't need to reserve this space. Therefore I would suggest:

- If `IsPanEnabled==false` and `IsZoomEnabled==false`, only reserve the space that is actually needed by the axis labels considering the axis range
- Otherwise, reserve enough space so all the axis labels could be accommodated even when zooming or panning (-> the current behavior)

Jonarw:
Fixes #1577.

### Checklist

- Improve auto plot margins when axis range is fixed
- Fix a small bug where the `ActualFont` etc. would be used for measuring the axis title instead of `AxisTitleFont` etc.
- Consolidate margin examples - instead of having them scattered over the PlotModel and Axis categories, I placed them all in one spot in the Axis category.

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
+++ b/Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
@@ -795,69 +795,6 @@ namespace ExampleLibrary
             return model;
         }
 
-        [Example("Auto adjusting plot margins")]
-        public static PlotModel AutoAdjustingMargins()
-        {
-            var model = new PlotModel
-            {
-                Title = "Auto adjusting plot margins",
-            };
-
-            Legend l = new Legend
-            {
-                LegendPosition = LegendPosition.RightBottom
-            };
-            model.Legends.Add(l);
-
-            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom, Title = "X", TickStyle = TickStyle.Outside });
-            model.Axes.Add(new LinearAxis { Position = AxisPosition.Left, Title = "Y", TickStyle = TickStyle.Outside });
-            model.Series.Add(new LineSeries { Title = "Butterfly curve", ItemsSource = ButterflyCurve(0, Math.PI * 4, 1000) });
-            return model;
-        }
-
-        [Example("Auto adjusting left plot margin")]
-        public static PlotModel AutoAdjustingLeftMargin()
-        {
-            var model = new PlotModel
-            {
-                Title = "Auto adjusting left plot margin",
-                PlotMargins = new OxyThickness(double.NaN, 40, 40, 40)
-            };
-
-            Legend l = new Legend
-            {
-                LegendPosition = LegendPosition.RightBottom
-            };
-            model.Legends.Add(l);
-
-            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom, Title = "X", TickStyle = TickStyle.Outside });
-            model.Axes.Add(new LinearAxis { Position = AxisPosition.Left, Title = "Y", TickStyle = TickStyle.Outside });
-            model.Series.Add(new LineSeries { Title = "Butterfly curve", ItemsSource = ButterflyCurve(0, Math.PI * 4, 1000) });
-            return model;
-        }
-
-
-        [Example("Manual plot margins")]
-        public static PlotModel ManualAdjustingMargins()
-        {
-            var model = new PlotModel
-            {
-                Title = "Manual plot margins",
-                PlotMargins = new OxyThickness(60, 4, 4, 40)
-            };
-
-            Legend l = new Legend
-            {
-                LegendPosition = LegendPosition.RightBottom
-            };
-            model.Legends.Add(l);
-
-            model.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom, Title = "X", TickStyle = TickStyle.Outside });
-            model.Axes.Add(new LinearAxis { Position = AxisPosition.Left, Title = "Y", TickStyle = TickStyle.Outside });
-            model.Series.Add(new LineSeries { Title = "Butterfly curve", ItemsSource = ButterflyCurve(0, Math.PI * 4, 1000) });
-            return model;
-        }
-
         [Example("Current culture")]
         public static PlotModel CurrentCulture()
         {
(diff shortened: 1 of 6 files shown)


Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
```diff
@@ -1409,5 +1346,81 @@ public static PlotModel MultiVerticalAxes()
...
+            var axis = GetLongLabelSeries();
+            axis.Angle = -90;
+            plotModel.Axes.Add(axis);
+            return plotModel;
+        }
+
+        [Example("Auto Margins - Wide Labels, fixed Range")]
+        public static PlotModel AutoMarginWideLabelsFixedRange()
+        {
+            var plotModel = new PlotModel() { Title = "Auto-adjusting plot margins - wide axis labels, fixed range", Subtitle = "When the axis range is fixed there should be no space reserved for axis labels" };
```
VisualMelon, line 1395: Nit: subtitle could be misleading (i.e. seems to contradict the next example).

Source/OxyPlot/Utilities/Helpers.cs
```diff
@@ -0,0 +1,27 @@
...
+    /// </summary>
+    public static class Helpers
+    {
+        /// <summary>
+        /// Switches the values of two specified variables.
+        /// </summary>
+        /// <typeparam name="T">The type of the variables.</typeparam>
+        /// <param name="value">The first value.</param>
+        /// <param name="other">The second value.</param>
+        public static void Switch<T>(this ref T value, ref T other) where T : struct
```
VisualMelon, line 20: I would oppose this being an extension mechanism on principle: there is no indication at the call-site that the `this` value is passed by reference, so the data-mutating semantics are unclear at a glance. In this case, I don't think there would be any loss of readability anyway, since neither parameter is 'more important'.

<hr />

I'd prefer it was called `Swap`, to distinguish it from the `switch` keyword and it's associated concepts.

<hr />

Is there a particular reason for the `T : struct` constraint?

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -868,39 +869,74 @@ public virtual void Measure(IRenderContext rc)
...
+            }
+            else if (majorLabelValues.Count > 0)
             {
-                case AxisPosition.Left:
-                case AxisPosition.Right:
-                    if (reachesMinPosition)
-                    {
-                        marginBottom = maximumTextSize.Height / 2;
-                    }
+                var minLabel = majorLabelValues.Where(d => d > this.ActualMinimum).Min();
```
VisualMelon, line 909: If the `Where` filter is necessary, that would suggest that the filtering should be done once and cached, so that it can be confirmed that this list isn't empty (otherwise `Min`/`Max` will thrown).

Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
```diff
@@ -1409,5 +1346,81 @@ public static PlotModel MultiVerticalAxes()
 
             return plotModel;
         }
+
```
VisualMelon, line 1349: Could we have an example with a vertical axis with rotated text? At the moment, it's hard to visually check that the transformations are doing the right thing.

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -825,7 +826,7 @@ public virtual void Measure(IRenderContext rc)
                 maximumTextSize = maximumTextSize.Include(size);
             }
 
-            var labelTextSize = rc.MeasureText(this.ActualTitle, this.ActualFont, this.ActualFontSize, this.ActualFontWeight);
+            var labelTextSize = rc.MeasureText(this.ActualTitle, this.ActualTitleFont, this.ActualTitleFontSize, this.ActualTitleFontWeight);
```
VisualMelon, line 829: Good catch.

Could this be renamed to `titleTextSize`?

Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
```diff
@@ -1409,5 +1346,81 @@ public static PlotModel MultiVerticalAxes()
...
+            var plotModel = new PlotModel() { Title = "Auto-adjusting plot margins - wide axis labels, fixed range", Subtitle = "The axis labels should exactly fit in the viewport" };
+            var axis = GetLongLabelSeries();
+            axis.IsPanEnabled = false;
+            axis.IsZoomEnabled = false;
+            axis.Minimum = -0.01;
+            axis.Maximum = 3.01;
+            plotModel.Axes.Add(axis);
+            return plotModel;
+        }
+
```
VisualMelon, line 1415: Needs an example with reversed axis (these don't work at the moment):

        [Example("Auto Margins - Wide Labels, fixed Range 3 (reversed axis)")]
        public static PlotModel AutoMarginWideLabelsFixedRange3()
        {
            var plotModel = new PlotModel() { Title = "Auto-adjusting plot margins - wide axis labels, fixed range", Subtitle = "The axis labels should exactly fit in the viewport" };
            var axis = GetLongLabelSeries();
            axis.StartPosition = 1;
            axis.EndPosition = 0;
            axis.IsPanEnabled = false;
            axis.IsZoomEnabled = false;
            axis.Minimum = -0.01;
            axis.Maximum = 3.01;
            plotModel.Axes.Add(axis);
            return plotModel;
        }

Source/OxyPlot/Utilities/Helpers.cs
Jonarw: Regarding your first comment: Personally I've used this extension for a while and therefore it seems natural to me. But I can absolutely see how this could be confusing to others, so I'll change that to a regular method.

Second comment: Fine with me.

Third comment: The constraint is just because C# only allows ref extension methods on value types. Can be removed once this is a regular method.

Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
Jonarw: You could always check the transposed version of the example with rotated text.
Or would you prefer to have it explicitly?

VisualMelon: … no, that's fine. I just forget they were there ;)

Source/OxyPlot/Utilities/Helpers.cs
VisualMelon: Oh yeah... I forgot about that detail completely. The new round of `ref` stuff makes me so sad.

Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
Jonarw: Oh wow you're absolutely right.
Misunderstanding from my side regarding what `ScreenMin` and `ScreenMax` actually mean...

VisualMelon: Maybe a 'reversed' checkbox to go with the 'transpose' checkbox would be a fun idea, so I can forget about it's existence as well.

VisualMelon: I've implemented a 'reversed' checkbox feature, and it didn't turn out as horribly as I thought it might, so I'll be putting a PR together soon.

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -817,6 +818,8 @@ public virtual void Measure(IRenderContext rc)
         {
             this.GetTickValues(out var majorLabelValues, out _, out _);
 
+            majorLabelValues = majorLabelValues.Where(d => d > this.ActualMinimum && d < this.ActualMaximum).ToList();
```
Jonarw, line 821: The test passes if we write >= and <= here... but that is not a complete solution, because the renderer also uses an epsilon when it checks if it should render a particular label: 
https://github.com/oxyplot/oxyplot/blob/9ffa2a89fdaac66f4cbe7f5cd1a527d02a6c5f87/Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs#L379

So basically we have no safe way of knowing here whether a label is actually rendered or not...

On the other hand, the worst that could happen is that we reserve slightly too much space for a label that isn't rendered. So maybe just use >= and <= and be done with it?

Jonarw: So I think there is no way to do this properly as long as we can't be sure which labels are actually rendered and which are not.
Why is it even the business of the renderer to decide which labels are rendered? Maybe we should move this responsibility out of the renderer altogether... I try something and see how it goes.

VisualMelon: I have many times lamented how inextensible axes are: I'd be very happy to help in any way I can with refitting them (testing, design, implementation) so that this sort of thing makes more sense. When I've tried in the past (and I've tried a few times), I have always given up and gone back to doing what I was meant to be doing before deciding to re-write the axes.

Jonarw:
Rebased, this now eliminates the need to to a range check in Axis.Measure.
I also snuck in a small fix for a bug introduced in #1581 where `CategoryAxis` could place major ticks outside the axis range when panning (how could I miss this??). I did not mention this in the changelog, as this is all unreleased stuff anyway.

VisualMelon:
There is an issue with angled axis labels:

image

Angled axis labels aren't evenly distributed on the left/right of the tick, so it's not giving enough to one margin, and too much to the other. Example was just `AutoMargin` with `Angle = 45`. The rendering behaviour with different angles is itself a bit weird.

Jonarw:
You are right of course.
Unfortunately however I don't see a good way to fix this with the current design. To figure out where angled labels end up, we would have to replicate the [logic](https://github.com/oxyplot/oxyplot/blob/bcb663fc72d68b9d42dcfef67ab95003f5df84c2/Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs#L674) used in the axis renderer, which is probably not what we want to do.

We should definitely open an issue for this, but I don't think we can feasibly support this currently, so I would say this is out of the scope of this PR.

Jonarw:
(rebased and squashed)

PR #1581, opened 2020-05-29
Fix #1580 Axis renderers do not render all tick values they are provided

### LINKED ISSUE #1580: Axis renderers do not render all tick values they are provided
This is kind of a blocker for #1577, which is why I had a look at this.

The basic problem is that that the Axis renderers [do](https://github.com/oxyplot/oxyplot/blob/9ffa2a89fdaac66f4cbe7f5cd1a527d02a6c5f87/Source/OxyPlot/Axes/Rendering/AngleAxisFullPlotAreaRenderer.cs#L59) [some](https://github.com/oxyplot/oxyplot/blob/9ffa2a89fdaac66f4cbe7f5cd1a527d02a6c5f87/Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs#L325) [checks](https://github.com/oxyplot/oxyplot/blob/9ffa2a89fdaac66f4cbe7f5cd1a527d02a6c5f87/Source/OxyPlot/Axes/Rendering/MagnitudeAxisRenderer.cs#L50) on the tick values provided to them and do not render all of them. This is problematic for a couple of reasons:

- The `Axis.Measure()` method can not do its job properly as it can't know for certain which labels are actually rendered and which might be discarded by the renderer
- The Axis renderers do some checks to remove minor ticks that are 'on top of' major ticks. They are [implemented in O(n²) complexity](https://github.com/oxyplot/oxyplot/blob/9ffa2a89fdaac66f4cbe7f5cd1a527d02a6c5f87/Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs#L550), though it should easily be possible to do this in O(n)
- Conceptually, I don't think it should be the responsibility of the renderer to decide which ticks to render. It should just take what it gets and draw it.

I propose to:

- Remove the above-mentioned checks from the axis renderers
- Change `AxisUtilities.CreateTickValues(...)` such that it only returns tick values inside the requested range
- Add a function to `AxisUtilities` which can be used to filter out minor ticks that are not needed because they are 'on top of' a major tick

I already had a look at this, I will do some more testing and will put together a PR later.

Jonarw:
Fixes #1580.

### Checklist

- Remove tick value checks from the axis renderers
- Change `AxisUtilities.CreateTickValues(...)` such that it only returns tick values inside the requested range
- Add `AxisUtilities.FilterRedundantMinorTicks(...)` which can be used to filter out minor ticks that are not needed because they are 'on top of' a major tick

Because these changes affect pretty much all kinds of plots, there is a lot of possibility of breaking things. Therefore I would be thankful if someone could second-guess this and tell me what I have forgotten...

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Axes/AxisUtilities.cs
+++ b/Source/OxyPlot/Axes/AxisUtilities.cs
@@ -62,35 +62,95 @@ namespace OxyPlot.Axes
         {
             if (step <= 0)
             {
-                throw new ArgumentException("Step cannot be zero or negative.", "step");
+                throw new ArgumentOutOfRangeException("Step cannot be zero or negative.", nameof(step));
             }
 
-            if (to <= from && step > 0)
+            if (to < from)
             {
-                step *= -1;
+                step = -step;
             }
 
-            var startValue = Math.Round(from / step) * step;
-            var numberOfValues = Math.Max((int)((to - from) / step), 1);
-            var epsilon = step * 1e-3 * Math.Sign(step);
-            var values = new List<double>(numberOfValues);
+            var epsilon = step * 1e-3;
+            from -= epsilon;
+            to += epsilon;
 
-            for (int k = 0; k < maxTicks; k++)
+            var startValue = Math.Ceiling(from / step) * step;
+            if (startValue == -0d)
             {
-                var lastValue = startValue + (step * k);
+                startValue = 0;
+            }
 
-                // If we hit the maximum value before reaching the max number of ticks, exit
-                if (lastValue > to + epsilon)
-                {
-                    break;
-                }
+            var values = new List<double>();
+            var sign = Math.Sign(step);
+            var i = 0;
 
-                // try to get rid of numerical noise
-                var v = Math.Round(lastValue / step, 14) * step;
-                values.Add(v);
+            var currentValue = startValue;
+            while ((to - currentValue) * sign >= 0 && i < maxTicks)
+            {
+                values.Add(currentValue);
+                currentValue = startValue + (++i * step);
             }
 
             return values;
+        }
+
+        /// <summary>
+        /// Analyses two lists of major and minor ticks and creates a new containing the subset of the minor ticks which are not too close to any of the major ticks.
+        /// </summary>
+        /// <param name="majorTicks">The major ticks. Must be monotonically ascending or descending.</param>
+        /// <param name="minorTicks">The minor ticks. Must be monotonically ascending or descending (same direction as major ticks).</param>
+        /// <returns>A new list containing a subset of the original minor ticks such that there are no minor ticks too close to a major tick.</returns>
+        public static IList<double> FilterRedundantMinorTicks(IList<double> majorTicks, IList<double> minorTicks)
+        {
+            if (majorTicks.Count == 0 || minorTicks.Count == 0)
+            {
+                return minorTicks;
+            }
+
+            var ret = new List<double>();
+            var previousMinorTick = 0d;
+            var j = 1;
+
+            var currentMajorTick = majorTicks.Count > 1 ? majorTicks[j] : majorTicks[0];
+
+            static double GetEpsilon(double tick1, double tick2)
+            {
+                return Math.Abs(tick1 - tick2) * 1e-3;
+            }
+
+            // If there is only one minor tick, we can't determine a meaningful epsilon. 
+            // But there also shouldn't be any precision loss, so we can require an exact match (epsilon = 0)
+            var epsilon = minorTicks.Count > 1 ? GetEpsilon(minorTicks[0], minorTicks[1]) : 0;
+
+            var sign = 1;
+            if (majorTicks.Count > 1 && majorTicks[0] > majorTicks[1])
+            {
+                sign = -1;
+            }
+
+            for (var i = 0; i < minorTicks.Count; i++)
+            {
+                var currentMinorTick = minorTicks[i];
+                if (i > 0)
+                {
+                    epsilon = GetEpsilon(currentMinorTick, previousMinorTick);
+                }
+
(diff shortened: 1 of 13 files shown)


VisualMelon:
I got an exception with Polar Plots > Semi-circle full plot area. I'm afraid I have no idea what caused it and can't seem to reproduce it, but I was resizing the example browser window at the time.

image

It happened while I was trying to get a screen shot of the magnitude axis rendering labels beyond the plot bounds:

image

This behaviour isn't new; I don't know whether the crash is. I wouldn't worry about it for the moment, but maybe message will help.

<hr />

I have tried to break the X/Y Axes, but they just seem to be improved over the old ones (minor tick filtering is more reliable).

Source/OxyPlot/Axes/Rendering/MagnitudeAxisFullPlotAreaRenderer.cs
```diff
@@ -115,7 +114,7 @@ public override void Render(Axis axis, int pass)
             {
                 OxyPen pen = this.MinorPen;
                 //var minorTicks = this.MinorTickValues.Where(x => x >= axis.ActualMinimum && x <= axis.ActualMaximum && !majorTicks.Contains(x)).ToArray();
```
VisualMelon, line 116: This comment could be removed.

VisualMelon:
There is a regression with `MagnitudeAxisFullPlotArea`: pre-filtering the minor tick values means that `ExtendTicks` can go wrong:

image

The outer-most minor ticks are not the right distance apart, so the 'extended' ticks are too spaced out. The example shown is `ArchimedeanSpiralFullPlotArea` with the series changed to:

    model.Series.Add(new FunctionSeries(t => t, t => t, 0, Math.PI * 6, 0.01));

No existing example presents this issue.

Jonarw:
Good find, thanks!
That `MagnitudeAxisFullPlotAreaRenderer` is a bit of a mess...

I've added a commit where I have moved the responsibility for determining the ticks to display out of the `MagnitudeAxisFullPlotAreaRenderer` and into `MagnitudeAxisFullPlotArea` to be consistent with the rest of this PR. This eliminates the need for the `ExtendTicks` method altogether, which is a good thing in my opinion.

Source/OxyPlot/Axes/AxisUtilities.cs
```diff
@@ -106,8 +106,18 @@ public static IList<double> FilterRedundantMinorTicks(IList<double> majorTicks,
             var ret = new List<double>();
             var previousMinorTick = 0d;
             var j = 1;
+
             var currentMajorTick = majorTicks.Count > 1 ? majorTicks[j] : majorTicks[0];
 
+            static double GetEpsilon(double tick1, double tick2)
+            {
+                return Math.Abs(tick1 - tick2) * 1e-3;
```
VisualMelon, line 114: I wonder if this could have issues with large values for `LogarithmicAxis.Base` (e.g. on the order of 10/1e-3).

Jonarw: Interesting thought!
I don't really know of an application where such large bases would be used, but of course that doesn't mean at all that they don't exist.

Just for fun I modified the LogarithmicAxis>Tick Calculation example by multiplying the bases by 1000. The result is kind of 'interesting':
image

So yes I do think `AxisUtilities.FilterRedundantMinorTicks(...)` could have a problem with these kinds of bases, but I don't think we really support them anyway.

VisualMelon: That first image looks like a bug in computing the minor tick interval rather than your stuff: did you happen to see what happens with the old code?

I agree this shouldn't be a blocker either way.

Jonarw: The first image is because the `LogarithmicAxis` falls back to linearly-spaced ticks when the range is too small. That's what I meant with we don't support these kinds of bases anyway.

Before this PR I reckon this would have looked pretty much the same.

Jonarw:
A bit of an oddity is that we see on many axis labels the tick label '-0' with this PR, for example:
image

I think I understand why this happens, this should just be because we display 5 significant digits from something like -0.0000000001. I am unsure however how to best fix this. I don't think the tick calculation does anything wrong here, this is just what you get when you work with floating-point numbers. 

I think this should be addressed in the label formatter. Could we just hard-code the case '-0' and return '0' instead? Is there a better way? Are there other edge cases like this?

VisualMelon:
I'd guess it is because `Math.Ceiling(-0.0000000001)` gives you `-0` instead of `0`. This could be the place to 'fix' this, I think.

Jonarw:
Oh yes you are right of course!

Jonarw:
(squashed)
I think I am pretty happy with this now, unless you can find some more issues I've overlooked?

Jonarw:
Apparently on AppVeyor a test of some Histogram example just randomly failed...

VisualMelon:
@Jonarw that's my fault: looks like a _random_ example failed. That can be ignored; I'll make the example deterministic some time in the future.

Jonarw:
(rebased)

Jonarw:
From my side this would be ready to merge.
Any more concerns?

VisualMelon:
@Jonarw if it's OK with you, I'd like to take another look (can probably do it tonight if I don't get distracted; otherwise, pester me tomorrow). I have also not looked at the latest changes to `MagnitudeAxisFullPlotAreaRenderer` at all yet.

VisualMelon:
Everything here looks line to me. `MagnitudeAxisFullPlotArea` looks much tidier now.

Jonarw:
Thanks for taking the time!

Issue #1401, opened 2019-10-20
Merge polar axis classes


objorke:
To simplify the axes used for polar plots.

The `AngleAxisFullPlotArea` and `MagnitudeAxisFullPlotArea` could be merged into the `AngleAxis` and `MagnitudeAxis`. The additional properties on the magnitude axis could also apply to the "not full plot" axes. 

Add a `CoverFullPlotArea` bool property on `MagnitudeAxis` and `AngleAxis` instead of creating the subclasses.

PR #1589, opened 2020-06-01
Fix full plot-area polar plot rendering with non-zero minimum values

### LINKED ISSUE #1586: Full Plotarea Magnitude Axis with negative minimum rendering is dodgy
When a `MagnitudeAxisFullPlotArea` has a negative minimum, the magnitude grid lines escape the plot area. The more of the axis is 'below zero', the more noticeable the aberrations.

image

Note that by default the magnitude axis has a non-zero `MinimumPadding`, which means that most examples have a negative minimum, but it is not large enough relative to the maximum to be noticeable if it is occurring.

This is similar to #1364.

Obviously this behaviour is not ideal, but does it make sense to allow the minimum to go negative? #44 is somewhat bare, but I wonder if this needs discussion.

VisualMelon:
Fixes #1586

### Checklist

- [ ] I have included examples or tests (see #1588 ) 

- Removes the erroneous transformation from screen space in `MagnitudeAxisFullPlotAreaRenderer.Render` performed before computing sweep angles

As far as I can tell, the transformation didn't make any sense, and uses a method that is explicitly marked as only for use by non-Polar plots. This was passing negative and otherwise incorrect values to `atan2` messing up the sweep angles. The additional checks added in #1370 remain necessary to deal with zero-values passed to `atan2`.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ All notable changes to this project will be documented in this file.
 - WPF PlotView still focusable when Focusable is false (#1440)
 - Disposing a SkiaRenderContext can mess up fonts from another SkiaRenderContext instance (#1573)
 - Display of ampersands in OxyPlot.WindowsForms Tracker (#1585)
+- Full Plotarea Polar plot rendering with non-zero minimum values (#1586)
 
 ## [2.0.0] - 2019-10-19
 ### Added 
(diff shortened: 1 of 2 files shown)


Jonarw:
There is some more weird stuff going on with the full-plot-area axes (try setting an angle of 15° on an `AngleAxisFullPlotArea` for example. Also the render performance on WPF is fairly bad)

This PR definitely improves the situation, so I would say it is worth merging it.

However I think a more substantial redesign of these classes, possibly in conjunction with #1401, might be in order at some point in the future.

PR #1597, opened 2020-06-05
Add supports HiDPI for winforms examples

rstm-sf:
### Checklist

- Add supports HiDPI for winforms examples

@oxyplot/admins


Jonarw:
Thanks for your contribution!
Could you please:

- Add a small note about this change to the [changelog](https://github.com/oxyplot/oxyplot/blob/develop/CHANGELOG.md)
- Add your name to the [list of contributors](https://github.com/oxyplot/oxyplot/blob/develop/CONTRIBUTORS)

Thanks!

Jonarw:
Ok so having now had a closer look at this, I realized that our WinForms-Renderer does not quite support any DPI other than 96 (it uses a fixed 1:1 ratio between 1/96th inch device-independent units and pixels). This is not a problem with this PR per se, but a limitation of the `GraphicsRenderContext` implementation.

I think we could still merge this just to have a showcase of what is _not_ supported at the moment and maybe at some point we'll want to update the `GraphicsRenderContext` accordingly.

What do you think @VisualMelon?

VisualMelon:
High DPI on WinForms is not something I've ever needed, so my input is limit to what I can find on SO and the [msdn](https://docs.microsoft.com/en-us/dotnet/framework/winforms/high-dpi-support-in-windows-forms).

I think you are right that this would be worth including: the example browser is in some senses a separate concern, and it will be useful for testing the `PlotView` if it is updated to support variable Dpi.

I'd be happy to work on adding DPI support to `GraphicsRenderContext`. I'm not convinced the font sizes are correct in WinFroms (e.g. the 0.8 font factor should be 0.75?), but that's a question for another day.

rstm-sf:
Fix TreeView for HiDPI when full framework

rstm-sf:
> High DPI on WinForms is not something I've ever needed, so my input is limit to what I can find on SO and the [msdn](https://docs.microsoft.com/en-us/dotnet/framework/winforms/high-dpi-support-in-windows-forms).

Support in core looks better. And it looks like it will continue to improve https://github.com/dotnet/winforms/issues/3202#issuecomment-626072014

rstm-sf:
I [tried to set](https://docs.microsoft.com/en-us/windows/win32/hidpi/setting-the-default-dpi-awareness-for-a-process#setting-default-awareness-with-the-application-manifest) system aware, maybe this will help?

rstm-sf:
Last change, so as not to fix with dpi == 96

VisualMelon:
I'll take a proper look at this sometime in the next few days, and I'll put to together a fix for the checkboxes which we can either merge into this or I'll just push through as a separate PR.

VisualMelon:
Taking a quick look just now (because I'm contemplating adding DPI support to GTK if it needs it), the tree view is still too big when using .NET Framework.

If I disable `FixTreeViewDpi`, with 175% scaling and NET Framework I get this (which looks fine):

image

With it enabled, I get this:

image

Similar behaviour with NET Core when I remove the `NETFRAMEWORK` check in `InitTree`

rstm-sf:
Thanks for the clarification! This problem also reproduces for me, but at 125 everything is fine

VisualMelon:
@rstm-sf sorry I didn't get back to you before. I'm not sure what you previous comment means, or what that last commit does: do you suspect that the winforms behaviour changes between 100 and 125 and 150?

rstm-sf:
I myself wonder why this is happening, but not enough to try to understand it at the moment... maybe later? Sorry :(

rstm-sf:
> the winforms behaviour changes between 100 and 125 and 150?

This is true, for some reason I do not understand, I have to adjust at 125%. This is not observed in .net core

rstm-sf:
I read that there will be a release soon. Will this PR wait a couple of weeks?

VisualMelon:
@rstm-sf yeah, I'd hope we will put the preview out in a couple of weeks. It would be nice to slip this in, but there are no API changes, so if it's not perfect on release we will get over it. Ping me if you want anything tested again.

rstm-sf:
Ok, thanks

VisualMelon:
Oh, and it's an example, so it won't actually _be_ released. Not sure how I missed that... It would non-the-less be nice to get this done, then we can sort out support in GDI itself.

rstm-sf:
Hello! Sorry for the lack

After a little research, I came to the conclusion that for some reason Legacy TreeView does not react to non-standard dpi up to a value equal to 144, i.e. 150%

Source/Examples/WindowsForms/ExampleBrowser/MainForm.cs
```diff
@@ -51,10 +51,10 @@ private void FixTreeViewDpi()
         {
             using (var g = this.CreateGraphics())
             {
-                var scaleFactor = g.DpiY / 96f;
-                if (Math.Abs(scaleFactor - 1.25f) > 1e-5)
+                if (g.DpiY <= 96.0 && 144.0 <= g.DpiY)
```
VisualMelon, line 54: Are these numbers from observations/testing, or is there some sort of reference?

rstm-sf: Yes, this is from tests

VisualMelon:
Thanks for picking this back up. I'll have another proper look tomorrow.

Source/Examples/WindowsForms/ExampleBrowser/MainForm.cs
VisualMelon: Could you add a comment to that effect? It's sure to confuse some more people in the future, and it's always a pain tracking down PR discussion.

rstm-sf: Good point ;)

VisualMelon:
Just took another proper look at this. I can't see any obviously broken behaviour now. There are some small differences between netcore and netframework, but I think it was just spacing and not a big issue.

I can't recall if there is any DPI stuff in net 5.0 that might warrant our adding _another_ target.

rstm-sf:
> Just took another proper look at this. I can't see any obviously broken behaviour now. There are some small differences between netcore and netframework, but I think it was just spacing and not a big issue.

Yes, it can be better configured there. I decided that it would be enough to make it look more or less

> I can't recall if there is any DPI stuff in net 5.0 that might warrant our adding another target.

I also :)

Jonarw:
Hi @rstm-sf! We are currently cleaning up old PRs (see discussion in #2012).
I am not quite sure what is missing in this PR. Would there be interest from your side to continue and getting this ready to be merged?

rstm-sf:
@Jonarw hello! I sync pr with develop. Is there anything else that needs to be done?

VisualMelon:
Again, can't test properly, but all seems OK to me

Source/Examples/WindowsForms/WindowsFormsDemo/app.manifest
```diff
@@ -0,0 +1,60 @@
...
+            <supportedOS Id="{35138b9a-5d96-4fbd-8e2d-a2440225f93a}" />
+
+            <!-- Windows 8 -->
+            <supportedOS Id="{4a2f28e3-53b9-4441-ba9c-d69d4a4a6e38}" />
+
+            <!-- Windows 8.1 -->
+            <supportedOS Id="{1f676c76-80e1-4239-95bb-83d0f6d0da78}" />
+
+            <!-- Windows 10 -->
+            <supportedOS Id="{8e0f7a12-bfb3-4fe8-b9a5-48fd50a15a9a}" />
```
VisualMelon, line 43: Do we need an entry for Win11? What is this list?

rstm-sf: Oops, missed it, but it's equals win10 https://learn.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1

VisualMelon: @rstm-sf I don't think you missed anything; just that this code is 3 years old ;)

Jonarw:
This is what it looks like on my screen that is running at 175% DPI scaling with the changes in this PR:
image

And this is on the current develop branch:
image

So with this PR everything is definitely sharper and not blurred by pixel scaling anymore. However, somehow the text in the TreeView becomes huge, do you know what is up with that @rstm-sf?

rstm-sf:
> This is what it looks like on my screen that is running at 175% DPI scaling with the changes in this PR

Mistake with conditions

Jonarw:
At 175%:
image

At 125%:
image

At 100%:
image

It seems to me that this PR is now doing what it's intended to do. The Winforms UI (and also the Plot) seem to be crisp at any DPI scaling.
It also makes it obvious that `GraphicsRenderContext` does not make any effort to support different DPI settings other than scaling the font, but that's not a problem with this PR.
Other than that, I can't comment much, as I haven't worked with Winforms for a long time. I would be happy to have this merged if you are @VisualMelon

VisualMelon:
> It also makes it obvious that GraphicsRenderContext does not make any effort to support different DPI settings other than scaling the font

Wow, I thought I'd implemented that at some point, but evidently not.

I'll try to look this over again later today.

VisualMelon:
No other comments from my side. I'm happy for this to be merged. I'll implement DPI support in WinForms at some point.

Jonarw (approved):

VisualMelon (approved):

PR #1601, opened 2020-06-07
Add Read-only Collection Interfaces for .NET 4.0 (#1600)

### LINKED ISSUE #1600: .NET 4.0 support - how to deal with lack of read-only collection interfaces?
As it became apparent in #1572, we may want to develop some kind of strategy how we deal with the support for .NET 4.0. As this was re-added only recently in #1452, we'll probably want to support this for a while (at least until the next major release I guess).

The specific problem that came up in the above PR was the lack of read-only collection interfaces in .NET 4.0 (specifically `IReadOnlyList<T>` and `IReadOnlyCollection<T>`).
There already are [some](https://github.com/oxyplot/oxyplot/blob/113cee0696bd37d4b6be207536d069992de67d72/Source/OxyPlot/Axes/Axis.cs#L718) [instances](https://github.com/oxyplot/oxyplot/blob/113cee0696bd37d4b6be207536d069992de67d72/Source/OxyPlot/Rendering/RenderContext/IRenderContext.cs#L58) where we use `IList<T>` where `IReadOnlyList<T>` would arguably be a better choice. In #1572 however this workaround is not possible due to the lack of covariance when using `IList<T>`.

There are several possibilities of dealing with this:

### 1. Add separate members for Count and Indexer to interfaces, like I've done in this [draft](https://github.com/oxyplot/oxyplot/pull/1572/files/3a08870f13ca39b4482a089e3aafed61196d773b#diff-191cb1ecc2002af11ce735d27a8195c8).

    public interface IExample<T>
    {
        IEnumerable<T> Items { get; }
        int ItemsCount { get; }
        T this[int index];
    }

Pro: 
- Very simple
- Works without restriction on all frameworks

Con:
- Clunky interface
- Breaking change for implementation whenever we drop .NET 4.0 support
- Breaking change for consumers whenever we drop .NET 4.0 support

### 2. Add a wrapper class for `IList<T>` that can be used like IReadOnlyList<T>, as proposed by VisualMelon in this [gist](https://gist.github.com/VisualMelon/d8baadc9f28f8d5f1bf60bf9eaf8c318).

    public interface IExample<T>
    {
        ReadOnlyList<T> Items { get; }
    }

Pro: 
- Works without restriction on all frameworks

Con:
- Breaking change for consumers whenever we drop .NET 4.0 support

### 3. a) Reference this [package](https://github.com/mausch/ReadOnlyCollections)
_(unfortunately the assemblies of this package are not signed, so we can't use it)_
### 3. b) Define `IReadOnlyList<T>` and `IReadOnlyCollection<T>` ourselves

    public interface IExample<T>
    {
        IReadOnlyList<T> Items { get; }
    }

This would be conditionally compiled for .NET 4.0 and would 'live' in `System.Collections.Generic`, like [this](https://github.com/mausch/ReadOnlyCollections/blob/master/ReadOnlyCollectionsInterfaces/Interfaces.cs).
We would also have to add a conditionally-compiled wrapper class on .NET 4.0.

Pro:
- We could use `IReadOnlyList<T>` and `IReadOnlyCollection<T>` exactly as we are used to
- No breaking changes whenever we drop .NET 4.0 support

Con:
- Potential confusion
- .NET 4.0 users who already use the same workaround may run into ambiguity issues

Jonarw:
Fixes #1600.

### Checklist

- Add read-only collection interfaces (conditionally compiled only for .NET 4.0)
- Add wrapper class implementing `IReadOnlyList<T>` and `IList<T>`

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Net40Compatibility/CompatibilityExtensions.cs
+++ b/Source/OxyPlot/Net40Compatibility/CompatibilityExtensions.cs
@@ -1,0 +1,55 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="CompatibilityExtensions.cs" company="OxyPlot">
+//   Copyright (c) 2020 OxyPlot contributors
+// </copyright>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot
+{
+#if NET40
+    using OxyPlot.Net40Compatibility;
+#endif
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// Provides extension methods facilitating compatibility with .NET 4.0.
+    /// </summary>
+    public static class CompatibilityExtensions
+    {
+        /// <summary>
+        /// Returns a <see cref="IReadOnlyList{T}"/> corresponding to the specified <see cref="List{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of the list.</typeparam>
+        /// <param name="list">The list.</param>
+        /// <returns>
+        /// On .NET4.0, this returns a wrapper around the <paramref name="list"/> which implements <see cref="IReadOnlyList{T}"/> and <see cref="IList{T}"/>.
+        /// On .NET4.5 and later, this directly returns the <paramref name="list"/>.
+        /// </returns>
+        public static IReadOnlyList<T> AsReadOnlyList<T>(this List<T> list)
+        {
+#if NET40
+            return new ReadOnlyListWrapper<T>(list);
+#else
+            return list;
+#endif
+        }
+        
+        /// <summary>
+        /// Returns a <see cref="IReadOnlyList{T}"/> corresponding to the specified array.
+        /// </summary>
+        /// <typeparam name="T">The type of the array.</typeparam>
+        /// <param name="array">The array.</param>
+        /// <returns>
+        /// On .NET4.0, this returns a wrapper around the <paramref name="array"/> which implements <see cref="IReadOnlyList{T}"/> and <see cref="IList{T}"/>.
+        /// On .NET4.5 and later, this directly returns the <paramref name="array"/>.
+        /// </returns>
+        public static IReadOnlyList<T> AsReadOnlyList<T>(this T[] array)
+        {
+#if NET40
+            return new ReadOnlyListWrapper<T>(array);
+#else
+            return array;
+#endif
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


Source/OxyPlot/Net40Compatibility/CompatibilityExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// <summary>
+        /// Returns a <see cref="IReadOnlyList{T}"/> corresponding to the specified <see cref="List{T}"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of the list.</typeparam>
+        /// <param name="list">The list.</param>
+        /// <returns>
+        /// On .NET4.0, this returns a wrapper around the <paramref name="list"/> which implements <see cref="IReadOnlyList{T}"/> and <see cref="IList{T}"/>.
+        /// On .NET4.5 and later, this directly returns the <paramref name="list"/>.
+        /// </returns>
+        public static IReadOnlyList<T> AsReadOnlyList<T>(this List<T> list)
```
VisualMelon, line 28: Is there an issue with just having one overload for `IList<T>`?

Jonarw: Yes, as `IList<T>` does not implement `IReadOnlyList<T>`, we can't directly assign a `IList<T>` to a `IReadOnlyList<T>` (hopefully this madness will stop with .NET 5...)

Jonarw:
This is really not as bad as I feared in terms of conditionals, and I also like that the compatibility stuff is all in one place and transparent to the rest of the library.

Any objections?

VisualMelon:
It looks fine to me, though I think we need to see it 'in context' before we can judge it: would be intending to refit the various `IList<T>`s we have dotted around before merging anything?

Jonarw:
Well I don't know if it makes sense to go through all the usages of `IList<T>` right now. This would involve for example some interface changes on `IRenderContext` and those are always fun...

My idea would have been more along the lines of:
We add the possibility of using `IReadOnlyList<T>` now, so future code can take advantage. But existing code using `IList<T>` remains unchanged for now and could be incrementally updated when we work on the related API anyway.

Maybe we can find an isolated piece of code to update right now where this functionality could be shown 'in action'. Thinking about it, #1572 wouldn't be a bad candidate for this.

VisualMelon:
That does sound like a better strategy. I wouldn't complain about shipping it with/ahead of 1572.

PR #1608, opened 2020-06-12
Advance System.Drawing.Common to version 4.7.0

### LINKED ISSUE #1605: Font family problem
There is a problem with texts, see the images.

Linux
Linux

Windows
Windows

Make a simple chart and compile for linux-x64 and run it on linux.

Platform:
.NET version: .net core 3.1

VisualMelon:
Fixes #1605 .

### Checklist

- Advance the dependency on System.Drawing.Common to 4.7.0. This shouldn't lose backwards compatability, and seems to fix text rendering under Linux. with .NET Core 3.1

@oxyplot/admins


Jonarw:
Should we mention this in the changelog? Or is this change so transparent that no one cares?

VisualMelon:
I don't really know. I suppose there can be no harm in it.

PR #1609, opened 2020-06-14
Wpf Speed Up

imalip:
### Checklist

I have kept the history as separate commits to try and make the review easier.

- Refactor the implementation of DrawLineSegmentsByStreamGeometry() by making a single figure, and applying the stroke to only every other segment. This removes many expensive calls to BeginFigure(), and speeds up the rendering of axes ticks, grids, and some of the markers.
- Add implementations of DrawRectangle(), DrawRectangles(), and DrawEllipses() using a StreamGeometry. This speeds up the rendering significantly when using Square or Circle markers.
- Add a property to allow changing the TextMeasurementMethod at runtime. This can help reduce the rendering time.
- Implement a DrawLineUsingStreamGeometry. The code itself is not really faster than using a Polyline, but again the rendering time is shorter. There is also a new GetActualPointsForStreamGeometry() to create an IList<Point> ready to be used by StreamGeometryContext.PloyLineTo(). The alternatives are to either call LineTo multiple times, or 2 calls to the Linq ToList() extension.
- I am still trying to assess what DrawLineBalanced actually does differently; but it has a really significant impact on speed as well (5.6 times for the C#, plus the rendering time). Incidentally I am wondering whether the MaxFiguresPerGeometry is still relevant. Removing these could allow further speed improvements.

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -789,46 +789,25 @@ namespace OxyPlot.Wpf
             double[] dashArray,
             LineJoin lineJoin)
         {
-            StreamGeometry streamGeometry = null;
-            StreamGeometryContext streamGeometryContext = null;
+            if (points.Count < 2)
+                return;
 
-            int count = 0;
+            var path = this.CreateAndAdd<Path>();
+            this.SetStroke(path, stroke, thickness, edgeRenderingMode, lineJoin, dashArray, 0);
+
             var actualStrokeThickness = this.GetActualStrokeThickness(thickness, edgeRenderingMode);
+            var actualPoints = this.GetActualPoints(points, actualStrokeThickness, edgeRenderingMode).ToList();
 
-            for (int i = 0; i + 1 < points.Count; i += 2)
+            var streamGeometry = new StreamGeometry();
+            using (var streamGeometryContext = streamGeometry.Open())
             {
-                if (streamGeometry == null)
-                {
-                    streamGeometry = new StreamGeometry();
-                    streamGeometryContext = streamGeometry.Open();
-                }
-
-                var actualPoints = this.GetActualPoints(new[] { points[i], points[i + 1] }, actualStrokeThickness, edgeRenderingMode).ToList();
-
                 streamGeometryContext.BeginFigure(actualPoints[0], false, false);
-                streamGeometryContext.LineTo(actualPoints[1], true, false);
-
-                count++;
-
-                // Must limit the number of figures, otherwise drawing errors...
-                if (count > MaxFiguresPerGeometry || dashArray != null)
+                for (int i = 1; i < actualPoints.Count; i++)
                 {
-                    streamGeometryContext.Close();
-                    var path = this.CreateAndAdd<Path>();
-                    this.SetStroke(path, stroke, thickness, edgeRenderingMode, lineJoin, dashArray, 0);
-                    path.Data = streamGeometry;
-                    streamGeometry = null;
-                    count = 0;
+                    streamGeometryContext.LineTo(actualPoints[i], (i & 0x1) != 0, false);
                 }
             }
-
-            if (streamGeometry != null)
-            {
-                streamGeometryContext.Close();
-                var path = this.CreateAndAdd<Path>();
-                this.SetStroke(path, stroke, thickness, edgeRenderingMode, lineJoin, null, 0);
-                path.Data = streamGeometry;
-            }
+            path.Data = streamGeometry;
         }
 
         /// <summary>
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -142,6 +142,12 @@ namespace OxyPlot.Wpf
         ///<inheritdoc/>
         public void DrawEllipses(IList<OxyRect> rectangles, OxyColor fill, OxyColor stroke, double thickness, EdgeRenderingMode edgeRenderingMode)
         {
+            if (this.UseStreamGeometry)
+            {
+                this.DrawEllipsesByStreamGeometry(rectangles, fill, stroke, thickness, edgeRenderingMode);
+                return;
+            }
+
             var path = this.CreateAndAdd<Path>();
             this.SetStroke(path, stroke, thickness, edgeRenderingMode);
             if (!fill.IsUndefined())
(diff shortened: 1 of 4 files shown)


Jonarw:
Thanks for your contribution! I'll take a closer look over the weekend.

Source/OxyPlot.Wpf/PlotView.cs
```diff
@@ -82,6 +82,7 @@ protected override void OnRender(DrawingContext drawingContext)
         /// <inheritdoc/>
         protected override void RenderOverride()
         {
+            RenderContext.TextMeasurementMethod = TextMeasurementMethod;
```
Jonarw, line 85: Missing `this`

Source/OxyPlot.Wpf/PlotView.cs
```diff
@@ -125,5 +126,22 @@ private void DoCopy(object sender, ExecutedRoutedEventArgs e)
...
+
+        /// <summary>
+        /// Gets or sets the vertical zoom cursor.
+        /// </summary>
+        /// <value>The zoom vertical cursor.</value>
+        public TextMeasurementMethod TextMeasurementMethod
+        {
+            get => (TextMeasurementMethod)this.GetValue(TextMeasurementMethodProperty);
+            set => this.SetValue(TextMeasurementMethodProperty, value);
+        }
```
Jonarw, line 145: I think those should be at the top of the class to be consistent with our general code style.

Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -1064,6 +1210,39 @@ private IEnumerable<Point> GetActualPoints(IList<ScreenPoint> points, double str
...
         }
 
+        /// <summary>
+        /// Snaps points to pixels if required by the edge rendering mode.
+        /// </summary>
+        /// <param name="points">The points.</param>
+        /// <param name="strokeThickness">The stroke thickness.</param>
+        /// <param name="edgeRenderingMode">The edge rendering mode.</param>
+        /// <returns>The processed points.</returns>
+        private Tuple<Point, IList<Point>> GetActualPointsForStreamGeometry(IList<ScreenPoint> points, double strokeThickness, EdgeRenderingMode edgeRenderingMode)
```
Jonarw, line 1220: I can't quite figure out why this function is needed - as far as I can see it does the exact same thing as `GetActualPoints(...)`, except it returns the first point separately. But can't the caller access this via something like `actualPoints[0]` anyway?

Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -789,46 +956,25 @@ private void ApplyToolTip(FrameworkElement element)
...
-                }
-
-                var actualPoints = this.GetActualPoints(new[] { points[i], points[i + 1] }, actualStrokeThickness, edgeRenderingMode).ToList();
-
-                streamGeometryContext.BeginFigure(actualPoints[0], false, false);
-                streamGeometryContext.LineTo(actualPoints[1], true, false);
-
-                count++;
-
-                // Must limit the number of figures, otherwise drawing errors...
```
Jonarw, line 813: Are we sure we don't create any regressions by removing this?
To be honest, I have no idea which issue this code originally solved, but I guess it was here for some reason.
For what it's worth, on my machine I don't get any drawing errors with or without this. If we can pin this to one specific issue in WPF rendering that is solved by now, I guess it is safe to remove this, otherwise I'm not sure...

Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -395,6 +498,70 @@ public void DrawRectangles(IList<OxyRect> rectangles, OxyColor fill, OxyColor st
             path.Data = gg;
         }
 
+        /// <summary>
+        /// Draws a collection of rectangles, where all have the same stroke and fill.
```
Jonarw, line 502: copy&paste error - only one rectangle here ;)

Source/OxyPlot.Wpf/PlotView.cs
imalip: Sorry, that one was a late addition, I'll fix it

Source/OxyPlot.Wpf/CanvasRenderContext.cs
imalip: That's exactly what it does. The issue is not about using actualPoints[0], but that StreamGeometryContext.PolyLineTo takes a an IList<Point> from the second point. Basically there are 2 ways to do it using GetActualPoints() :
- multiple individual StreamGeometryContext.LineTo() calls and either use GetActualPoints().ToList() or count manually to treat the firspoint differently
- Call GetActualPoints().ToList() to get the first point, then use actualPoints.Skip(1).ToList(), which means alllocating an additional array

This implementation allowed me to use the faster PolyLineTo(), while not spending time in ToList(). Not super happy about it, but it's the best solution I could find

Source/OxyPlot.Wpf/PlotView.cs
imalip: sure, that one is easy to fix :)
I wasn't sure where to put it as all the other properties are under OxyPlot.Wpf.Shared, but TextMeasurementMethod isn't

Source/OxyPlot.Wpf/CanvasRenderContext.cs
imalip: I'm pretty sure. This value is here to limit the number of figures per geometry, but this changes it to having a single geometry with a single figure in it, so this limitation doesn't apply anymore.

Source/OxyPlot.Wpf/CanvasRenderContext.cs
imalip: Sorry, I have been lazy :)

imalip:
Hi,

- I will open an issue, no problem
- The rectangles corners issue has been identified and fixed (at least locally)
- I'll run some benchmarks, though some will be a bit tricky to measure as they happen outside of what tools like dotTrace measure.

imalip:
Ok, I just ran some performance tests using dotTrace. I created 30 scatter series with 1000 random points each, and called PlotModel.InvalidatePlot(false) so it would run continuously for 3 minutes. All series get the same marker.
So essentially, I've been testing the performance through the number of calls to ScatterSeries.Render()
The results for different marker types :
- Square: From 1,980 calls to 23,130
- Circle: From 991 calls to 14,457
- Star: From 2,820 calls to 7,452

Interestingly, the average time per call doesn't change much according to dotTrace, so the majority of the time gained appears to be in the underlying WPF layers

Source/OxyPlot.Wpf/CanvasRenderContext.cs
Jonarw: Hmm, I don't really like the fact that the implementation of `GetActualPoints(...)` is duplicated. Could we use something like this instead:

        static void GetFirstAndRest<T>(IEnumerable<T> sequence, out T first, out IList<T> rest)
        {
            using var e = sequence.GetEnumerator();
            rest = new List<T>();
            if (!e.MoveNext())
            {
                first = default;
                return;
            }

            first = e.Current;
            while (e.MoveNext())
            {
                rest.Add(e.Current);
            }
        }

(untested, just as a concept)

Source/OxyPlot.Wpf/CanvasRenderContext.cs
Jonarw: Ah I see, thanks for the clarification!

Jonarw:
Thanks for the benchmark results, looks promising!

Jonarw:
> 
> 
> Hi,
> 
>     * I will open an issue, no problem
> 
>     * The rectangles corners issue has been identified and fixed (at least locally)
> 
>     * I'll run some benchmarks, though some will be a bit tricky to measure as they happen outside of what tools like dotTrace measure.

Hi @imalip! Where you planning to follow up on these points? There are definitely people who could be interested in your improvements! (#1639)

imalip:
> 
> 
> > Hi,
> > ```
> > * I will open an issue, no problem
> > 
> > * The rectangles corners issue has been identified and fixed (at least locally)
> > 
> > * I'll run some benchmarks, though some will be a bit tricky to measure as they happen outside of what tools like dotTrace measure.
> > ```
> 
> Hi @imalip! Where you planning to follow up on these points? There are definitely people who could be interested in your improvements! (#1639)

Hi,

Apologies, work caught up at the beginning of July and I've barely had a day available. I'll get back to it this week and should be able to offer updated patches soon.

Jonarw:
Hi @imalip!
Would it be alright with you if I picked this up from here? Of course I'd keep you as co-author.

PR #1613, opened 2020-06-24
fix not correct char

lwqwag:
Fixes the char not correct in vs for chinese

@oxyplot/admins


VisualMelon:
Could you elaborate on the issue? I'm afraid I don't know anything about non-western script, but the degree symbol should be there.

lwqwag:
> Could you elaborate on the issue? I'm afraid I don't know anything about non-western script, but the degree symbol should be there.

At first I don't know it's a  symbol because In Chinese  visual studio ,the symbol can't display correct like the image. 

For now ,I agree with you and close the pr. Thank you!

VisualMelon:
Out of interest, which font do you use for the editor?

lwqwag:
> Out of interest, which font do you use for the editor?

The text editor font is Luicida Sans Typewriter. I found the problem was that Chinese (language) version  visualstudio would select GB-2312 charset.

Issue #1621, opened 2020-07-03
Issues with interaction without rendering


VisualMelon:
When performing interactions quickly in e.g. OxyPlot.WinfowsForms, it is possible for an interaction (e.g. mouse wheel) to take place, and then another to take place before rendering occurs. Because rendering has no occurred, the Axes are not updated, and consequently the position in data-space where the screen point is projected is incorrect.

This is extremely noticeable when scrolling in the [immediate mode renderer for OxyPlot-Avalonia](https://github.com/VisualMelon/oxyplot-avalonia/tree/ImmediateMode) I threw together last night.

Much of the issue could be resolved by calling `UpdateAxisTransforms` after any interaction, as currently the axes are left in a sort of half-updated state that returns incorrect values.

However, this still leaves problems when the axis margins change, producing behaviour that is inconsistent with the same set of interactions performed more slowly. The only solution to that would be to layout the whole plot after each interaction, which would be expensive and is presently impossible without rendering e.g. to a `null` context.

Separation between render and layout may facilitate this. I'll try to dig out my old core/musings and put them in #391

PR #1623, opened 2020-07-12
Axis margin

### LINKED ISSUE #1569: Screen space axis padding
### Screen space axis padding (Feature request)

Essentially, the addition of 2 properties to `Axis`, which determine some amount of screen padding that is included beyond the minimum and maximum values, either computed or user-defined. The purpose is to ensure some amount of screen space between the plot area borders and the series geometry, which is necessary for the clear presentation of information.

image

Existing approaches to adding this padding are not ideal:
 - The `MinimumPadding` and `MaximumPadding` properties are relative to the size of the plot: as such, no single value is suitable for both small and large figures: consistency between the two is infeasible
 - The axis `StartPosition` and `EndPosition` have the same problem or being relative to the plot size, and clip the axes themselves, which may not be desirable
 - Setting `Minimum` and `Maximum` manually has the effect of perfectly bounding the defined region in data-space, which means the user has to adjust the chosen values to compensate (i.e. they don't make sense if you actually want to see the minimum and maximum)

The idea is that the computed minimum and maximum could be left as the actual series extents, or the user could set the minimum and maximum explicitly. Having tried a couple of times in the past, I don't think that implementing such a feature will be easy. Furthermore, it is debateable what the behaviour should be upon resizing a plot: should the data-space bounds be preserved with or without the padding? What would `Minimum` and `Maximum` really mean? Would it be better to avoid that issue, and keep `Minimum` and `Maximum` as they are, and try to bolt on another two properties that are the specified minimum and maximum _before_ user interaction (e.g. they are ignored once the minimum and maximum are computed for the first time)?

Sorry if this is a bit wooly: it's a feature I've wanted for years, and never managed to work out how to implement it before fudging a solution and getting on with what I was meant to be doing. If there is interest from other parties, perhaps we can come up with a design that won't too greatly increase the complexity already involved in determining the default axis extents in data-space.

An analogous set of screen-space properties that crop the axis itself (rather than changing the framing of the data) could be useful. These would be significantly easier to implement, and it should just involve deflating the plot area given to `UpdateIntervals`.

VisualMelon:
Fixes #1569 .

### Checklist

- Adds properties for `MinimumMargin`, `MaximumMargin`, which apply to all axes except angle axes, adding an absolute screen-space margin within the axis (as described in #1569) (where `MinimumPadding` and `MaximumPadding` pad in relative data-space).
- Adds `DisplayMinimum` and `DisplayMaximum` properties to `Axis`, and replaces most usages of `ActualMinimum` and `ActualMaximum` with these: the `Actual` bounds are those determined from data min/max or user zooming/panning; the `Display` bounds are the `Actual` bounds with the margin added.
- Adds properties `MinimumOuterMargin`, and `MaximumOuterMargin` which only apply to horizontal/vertical axes, which limit the axis bounds in absolute screen space (where `StartPosition` and `EndPosition` limit them in relative screen space).
- Adds various examples (happy to add more on request).
- New properties have default `0`, and should have no impact on any existing behaviour by design when they are `<= 0`

Here is an example with extra annotations to show what is going on:

image

_(Axis Examples > All margins and padding)_

### Outstanding Concerns

 - The names are not great.
 - The behaviour is dodgy/not considered for when the margins are greater than the plot width/height.
 - Existing code depending on `ActualMinimum` and `ActualMaximum` is probably now broken: most things want `DisplayMinium` and `DisplayMaximum` instead.
 - The outer margin and position for the middle of a magnitude axis do not work

@oxyplot/admins


VisualMelon:
@Jonarw just wondering if I could get your opinion on some naming. I'm wondering if OuterMargin should be Margin, and Margin should be Inset (I would have called it padding, but that's taken).

VisualMelon:
I've finally addressed all the little concerns I was putting off, so I'm opening this for review.

Jonarw:
Regarding naming, just as an idea: How about we call the `OuterMargin` `TickMargin` and the `Margin` `SeriesMargin`?

Other than that I'll take a look at the code soon.

VisualMelon:
@Jonarw that's a great idea! I'd be tempted to go for `DisplayMargin` (to correspond with `DisplayMinimum` and because it also applies to clipping and gridlines and the axis line itself), and I'm not sold on `SeriesMargin`, but they make the semantic distinction that I was completely ignoring before.

Source/OxyPlot/Series/RectangleSeries.cs
```diff
@@ -216,6 +216,11 @@ protected void RenderRectangles(IRenderContext rc, OxyRect clippingRect, ICollec
 
                 var rectrect = new OxyRect(p1, p2);
 
+                if (!clippingRect.Intersects(rectrect))
+                {
+                    continue;
+                }
+
```
Jonarw, line 223: This isn't really related to the axis margin, or is it? Should this be a separate issue/PR?

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -1098,8 +1143,8 @@ public virtual double Transform(double x)
         /// <param name="newScale">The new scale.</param>
         public virtual void Zoom(double newScale)
         {
-            var oldMinimum = this.ActualMinimum;
-            var oldMaximum = this.ActualMaximum;
+            var oldMinimum = this.ActualMaximum;
+            var oldMaximum = this.ActualMinimum;
```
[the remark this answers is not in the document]
VisualMelon: No, that looks like a (weird) mistake.

Source/OxyPlot/Series/RectangleSeries.cs
VisualMelon: Yes, it should be a separate PR.

VisualMelon:
> On the other hand, this only becomes an issue if MinimumMargin != 0, correct?

Yeah, that's absolutely the idea.

> I still wonder if we could make the difference between DisplayMinimum and ActualMinimum more obvious from the naming

Yeah, I'm by no means attached to any names, and an option I've still not fully would be to 'rename' `DisplayMinimum` as `ActualMinimum`, and rename `ActualMinimum` some other way. I'll take another look at that over the weekend, because I _think_ `ActualMinimum` is only really used internally, so using it as the display bounds would limit the another things that break.

VisualMelon:
Sorry, squashed everything because I thought I'd make a mistake with the revert, but it just conflicts in the changelog...

VisualMelon:
Over the weekend I concluded it would be perfectly reasonable to change `DisplayMinimum` back to `ActualMinimum` (which corresponds to most existing usage, since it has always been informed by display concerns). `ActualMinimum` has been renamed to `LogicalMinimum`. Since they are what one would traditionally consider margins, I've renamed the `MinimumOuterMargin` to just `MinimumMargin` (i.e. I don't think a prefix like `Display` is necessary; `Actual` would just be weird), and renamed the `MinimumMargin` (the 'inner' one) to `MinimumLogicalMargin`.

image

I think names for the 'outer' stuff (i.e. what is now `ActualMinimum` and `MinimumMargin`) are better, but I'm not happy with `Logical` for the inside stuff. Possibilities:
 - Logical
 - Data
 - Series

Again, I'm not attached to any of the names, so any suggestions are welcome.

VisualMelon:
Looks like Snyk is unhappy with me again.

Source/OxyPlot/Axes/LogarithmicAxis.cs
```diff
@@ -52,6 +52,16 @@ public LogarithmicAxis()
...
+        /// <summary>
+        /// Gets or sets the display logarithmic maximum value of the axis.
+        /// </summary>
+        protected double LogDisplayMaximum { get; set; }
+
+        /// <summary>
+        /// Gets or sets the display logarithmic minimum value of the axis.
+        /// </summary>
+        protected double LogDisplayMinimum { get; set; }
+
```
Jonarw, line 64: Should these be renamed to be consistent?

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -298,8 +298,8 @@ protected virtual void RenderMajorItems(Axis axis, double axisPosition, double t
         {
             double eps = axis.ActualMinorStep * 1e-3;
 
-            double actualMinimum = axis.ActualMinimum;
-            double actualMaximum = axis.ActualMaximum;
+            double displayMinimum = axis.ActualMinimum;
+            double displayMaximum = axis.ActualMaximum;
```
Jonarw, line 302: Does this rename still make sense?

Source/Examples/ExampleLibrary/Axes/AxisExamples.cs
```diff
@@ -1413,6 +1413,341 @@ public static PlotModel AutoMarginWideLabelsFixedRange2()
...
+            model.Series.Add(new FunctionSeries(t => t, t => t, 0, Math.PI * 6, 0.01));
+            return model;
+        }
+
+        [Example("Axis Margins")]
+        public static PlotModel AxisOutMargins()
+        {
+            var plot = new PlotModel
+            {
+                Title = "YAxis are evenly distributed with a constant gap",
```
Jonarw, line 1571: Should say 'YAx**e**s'

VisualMelon:
@Jonarw feedback addressed. For the record, my only reservation with `Data` and `Series` is they don't make sense when the user starts interacting. I don't like `Logical` because it doesn't convey anything at all.

@objorke I wondered if you might have time to provide an opinion on the naming? It's probably easiest just to look at the grey screenshot above to see the current state of naming. `Actual` I think will stick (because it provides backwards compatibility), but `Logical` (the bounds determined by `data min/max + padding` _or_ user interaction) doesn't fill us with joy.

<hr />

 [ ] Note to self: update the changelog when we settle on the naming

VisualMelon:
A big issue has just occurred to me while looing through the axis tests: the `ActualMinimum`/`ActualMaximum` are no longer known until the plot renders (they can't be known until the plot measures itself, and currently these happen at once).

I've not noticed any weirdness anywhere, but someone could be calling `IPlotModel.Update` and expecting that to update `ActualMinimum` (as it would before), only to find it isn't updated, and that `LogicalMinimum` isn't what they want. What they want is to be able to measure/layout the plot (see #391). This provides a case for changing `Logical` back to `Actual` so that it is updated when expected.

Jonarw:
Maybe a slightly crazy idea:
What if we avoid this whole naming conundrum by completely removing `ActualMinimum` and `ActualMaximum` (or at least marking it as obsolete, redirecting it and mentioning the new properties in the obsolete message)? Then people will be aware that something changed here and can choose the appropriate property themselves.

A possible naming scheme could be:

`ActualMinimum` -> `TickMinimum`
`LogicalMargin` -> `DataMargin`
`LogicalMinimum` -> ??? maybe `SuggestedDataMinimum` or `DataMarginMinimum` or `DataMinimumWithMargin` or ???

Just brainstorming, feel free to completely dismiss all of this ;)

VisualMelon:
I think I'm slowly coming coming to like `Data` as the prefix for the inner bounds and margins... but I'm also rapidly coming to the conclusion that I'm not going to be happy with anything until someone can call `Measure` on the `PlotModel` and find out the outer margins. I'll have another look at that today.

Obsoleting and redirecting `ActualMinimum` etc. sounds like a plan to me... that or it should be the inner bounds so that it is always up-to-date.

VisualMelon:
I've gone with `ActualMinium` this time, determined by `DataMinimumMargin`; and `ClipMinimum` determined by `MinimumMargin`. The rationale is that `ActualMargin` is the exactly the same as ever when the `Data` margins are zero. While the match is worse when the `Data` margins are non-zero, the `Clip` margin is only known at render-time, and we don't want that to be true for `ActualMargin` because that would be nightmare inducing silent change, and the odds of old-code being updated in response to it being marked obsolete are probably quite low.

image

I'm not happy with `Clip`, but `Tick` is too specific I think. I'm gravitating back to `Display`, but I don't have very objective reasons.

VisualMelon:
If we can expedite #1556, then we can hopefully do a nice job of #391 (because we can split `IRenderContext` and `ITextMeasurer`), and then the outer bounds can be computed before rendering, which will make me feel much better (and deal with #1621), but I think keeping `Actual` as it is now is the right thing to do, whether we make it obsolete or not.

Jonarw:
Sorry, I was away for a couple of days.

`DataMinimum` sounds nice in theory, but is actually already taken and represents the `minimum value of the data displayed on this axis`.

I don't think there is a perfect option, but I guess `Clip` will have to do.

So I'd say, if there is nothing more from your side, this should be good to go.

VisualMelon:
@Jonarw thanks. I'll hold on for the moment, but when I have time to start working on things properly again I'll get this merged (assuming no other problems comes up).

VisualMelon:
Plan for the weekend:
 - Review what's already here (in particular, the documentation and changelog)
 - Review `ScreenMin` and `ScreenMax`: half of the values are nonsense... but I need to need to check how they are used. A fix will probably go into another Issue/PR, but I want to check it won't throw up any problems with this
 - Get this merged

VisualMelon:
@Jonarw I've not checked if there are conflicts, but if #1661 goes through today, I'll try to get this though as well (nothing much else to do today). Then I think it's worth thinking about a 2.1 release to get Skia out there, but I'd like this feature too ;). If we think a release is a good idea, we should quickly decide whether #1556 should be included or not. My main concern is that I don't know who to test it properly with non-latin script. I think by bringing the test trimming 'in-house', issues like #1659 should stop occurring.

Jonarw:
I'm guessing there are no major changes since I last looked through the code, right?

One more thing that occured to me - we have this nice example:
image

Could the subtitle state the exact names of the related properties in code, e.g. `ClipMinimum` etc.?
(Or if we want to be fancy this could be even done with some `TextAnnotation`s in the plot, but I guess that's not really necessary)

VisualMelon:
@Jonarw How about this?

image

I don't think anything has changed; I think I was just working on the examples. I still need to go through all the documentation again for my sanity, which I'll get on with now.

Jonarw:
Yes, I think that's good. 
I think it's good to have a place to look when I forget which property is which (and to have somewhere to point others if the question comes up).

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -1402,6 +1470,33 @@ internal virtual void UpdateTransform(OxyRect bounds)
...
+                this.ClipMaximum = this.ActualMaximum;
+            }
+
+            this.ActualMaximumAndMinimumChangedOverride();
+        }
+
+        /// <summary>
+        /// Invoked when <see cref="ClipMinimum"/> and <see cref="ClipMaximum"/> are changed.
+        /// </summary>
+        protected virtual void ActualMaximumAndMinimumChangedOverride()
```
VisualMelon, line 1498: ... this is technically true; the `Actual` bounds are changed a lot without anyone knowing, but the method name is preserved. I'll change it to "Invoked when <see cref="ActualMinimum"/>, <see cref="ActualMaximum"/>, <see cref="ClipMinimum"/>, and <see cref="ClipMaximum"/> are changed."

VisualMelon:
Rebased. I've nothing else to add.

PR #1624, opened 2020-07-12
Assign colors to invisible series (#1599)

### LINKED ISSUE #1599: Series made invisible via the legend lose their automatic colors
### Steps to reproduce

1. Open any of the examples with a legend
2. Change the visibility of the top-most series

Platform: Any
.NET version: Any

This isn't a bug in the sense that this behaviour is as expected and is a natural extension of the original behaviour for making axes invisible, but it can be jolly confusing to say the least, especially when testing changes like #741 where the complex behaviour is all revealed by changing axis visibility and the series keep changing color.

Do we want to do anything about this? Some off-hand notes:
 - We don't want `IsVisible` to change to lie
 - The current automatic color assignment mechanism involves the series requesting colors as and when they need them, and the plot model cycles them for each rendering
 - It would be nice if the color of the axis could be displayed in the legend when it isn't being shown as an additional visual identifier
 - This would obviously be a breaking change, so it might be worth adding a per-series property to control the behaviour of whether a series 'claims' its own colors when it isn't being draw, rather than, say, reinterpreting a combination of existing properties when `IsVisible` is `false`.
 - The `SetDefaultValues` call is already handled by `PlotModel` rather than `Series`, so any changes should be small

VisualMelon:
Fixes #1599 .

### Checklist

- Adds a `bool AssignColorsToInvisibleSeries` property to `PlotModel` to determine whether invisible series should be assigned automatic colors
- Default for `AssignColorsToInvisibleSeries` is true, which is a breaking change, but I believe provides a significantly better experience when checking series on and off via the legend

Good existing examples are Bar Series > All in one, and Function Series > Lamé Curves.

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/PlotModel/PlotModel.cs
+++ b/Source/OxyPlot/PlotModel/PlotModel.cs
@@ -128,6 +128,7 @@ namespace OxyPlot
             this.PlotAreaBorderThickness = new OxyThickness(1);
             this.EdgeRenderingMode = EdgeRenderingMode.Automatic;
 
+            this.AssignColorsToInvisibleSeries = true;
             this.IsLegendVisible = true;
 
             this.DefaultColors = new List<OxyColor>
@@ -251,6 +252,11 @@ namespace OxyPlot
         /// </summary>
         /// <value>The edge rendering mode. The default is <see cref="EdgeRenderingMode.Automatic"/>.</value>
         public EdgeRenderingMode EdgeRenderingMode { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether invisible series should be assigned automatic colors.
+        /// </summary>
+        public bool AssignColorsToInvisibleSeries { get; set; }
 
         /// <summary>
         /// Gets or sets a value indicating whether the legend is visible. The titles of the series must be set to use the legend.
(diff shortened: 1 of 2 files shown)


Source/OxyPlot/PlotModel/PlotModel.cs
```diff
@@ -128,6 +128,7 @@ public PlotModel()
             this.PlotAreaBorderThickness = new OxyThickness(1);
             this.EdgeRenderingMode = EdgeRenderingMode.Automatic;
 
+            this.AssignColorsToInvisibleSeries = true;
```
Jonarw, line 131: This will be a breaking change, as in it will potentially change the automatic colors given to series when invisible series are present.
I'm not saying it's necessarily a bad thing, just wanted to consider it. In most cases it is probably preferable to assign colors to invisible series, but one could also imagine situations where this might lead to 'odd' results.

Overall I think it's fine to enable this by default. Maybe the fact that the default behaviour changed could be stressed a bit more in the changelog?

VisualMelon: If you think it's OK to change the default, then I say let's go for it. Equally, I'd be happy making it not the default for compatibility: my only real interest is in the default for the example browser.

What might you suggest for the changelog? Some explicit indication that this is a silent breaking change?

Jonarw: Maybe mention the new `PlotModel.AssignColorsToInvisibleSeries` in the `Added` section and add a separate point in the `Changed` section along the lines of `Enable PlotModel.AssignColorsToInvisibleSeries by default`? I think this would make it more obvious.

But if you have a better idea I'm also happy :)

VisualMelon: Yeah, might as well add an entry to `Added`. I'll get on with that now.

PR #1627, opened 2020-07-14
Overlapping bar series (#1265)

### LINKED ISSUE #1265: How can I make Columns/Bar Series overlap?
*IF BUG, INCLUDE THIS PART:*

### Steps to reproduce
http://docs.oxyplot.org/en/latest/models/series/BarSeries.html
1. Create a Model with multiple Column Series 
2. See that they are created beside each other
Tried playing with negative GapWidth in the CategoryAxis which is a step in the right direction but is not the desired outcome and not controllable enough

Platform:
.NET version: 4.5.1
Oxyplot.Core, OxyPlot.WindowsForm v1

### Expected behaviour
http://www.stephanieevergreen.com/wp-content/uploads/2016/04/Overlapping4.png
image

Options in Column/Bar Series for Column width, and Offset property perhaps. IsOverlap? similarly to isStacked...    

Tell us what happens instead
Can you also include a screen shot?
image

*IF IT IS A NEW FEATURE REQUEST, INCLUDE THIS PART:*

### Feature description
Allow overlapping bars in BarSeries and ColumnSeries.
Options in Column/Bar Series for Column/Bar width, and Offset properties perhaps or IsOverlap property? similarly to isStacked... 
http://www.stephanieevergreen.com/wp-content/uploads/2016/04/Overlapping4.png
image

VisualMelon:
Fixes #1265 .

### Checklist

- Adds an `OverlapsStack` property to `IStackableSeries` which determines whether a stacked series should overlaps its stack, with the effect that it is not integrating in the stack base position
- Adds a dedicated example, and an additional series to the 'All in one' example

I'm not convinced by the name `OverlapsStack`, but this is a simple way of implementing this.

### Outstanding concerns

 - Do we want this feature at all? It seemed like a good idea when I was implementing it, but I'm not sure it is now.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ All notable changes to this project will be documented in this file.
 - Native Clipping for OxyPlot.SvgRenderContext (#1564)
 - Examples of full plot area polar plots with non-zero minimums (#1586)
 - Read-Only collection interfaces for .NET 4.0 (#1600)
+- Overlapping bar series (#1265)
 
 ### Changed
 - Legends model (#644)
--- a/Source/Examples/ExampleLibrary/Series/BarSeriesExamples.cs
+++ b/Source/Examples/ExampleLibrary/Series/BarSeriesExamples.cs
@@ -909,6 +909,29 @@ namespace ExampleLibrary
             return model;
         }
 
+        [Example("BaseValue (overlaping)")]
+        public static PlotModel BaseValueOverlaping()
+        {
+            var model = new PlotModel { Title = "BaseValue (overlaping)", Subtitle = "BaseValue = -1" };
+            var series1 = new BarSeries { Title = "Series 1", IsStacked = true, OverlapsStack = true, BaseValue = -1 };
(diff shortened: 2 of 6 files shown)


Jonarw:
Looks good to me if there is nothing more from your side :)

VisualMelon:
@Jonarw do you think the naming and redundancy is OK (e.g. this does nothing if the series isn't stacked)? I'm not sold on the feature, but it shouldn't be high maintenance going forward.

Jonarw:
I don't think I have an issue with the redundancy, as long as it is made explicit in the documentation that the `OverlapsStack` property only has an effect for stacked series - which is the case.

The feature itself I think would be nice to have.

Regarding the name I agree with you, but I also can't really think of a better alternative.

VisualMelon:
@Jonarw I'm happy to merge then, if you are. Is the conflict something we can results in the new-fangled merging system, or do I need to rebase at my end?

Jonarw:
Simple conflicts like this can be easily enough resolved in the web editor, so no need to rebase :) 
Will merge once CI is happy.

PR #1643, opened 2020-08-14
ColorAxes request no margin with AxisPosition.None (#1574)

### LINKED ISSUE #1574: Plotarea is calculated wrong in HeatMap
I'm using the latest commit (699e9ddb) of Oxyplot from Github. There is an undesired behavior with the heatmap series, which can be explained with the examples.
This is the 2x3 interpolated example:
with color axis

If I change it now to not display the ColorAxis it looks like this:
no color axis

It still leaves the space on the right side for the ColorAxis and additionally it creates an extra margin between the title and the chart. It also looks like there is more margin on the left side.

VisualMelon:
Fixes #1574.

### Checklist

- Overrides the `Measure` method in the 3 Color Axes (`Linear-`, `Range-`, `Category`) so that they have a desired margin of zero when they are set to `Position.None`
    - Before, they would request margins on all sides, because they were behaving like a polar axis
- Adds an example for LinearColorAxis (though not the other 2 color axes) where `Position == AxisPosition.None`

@oxyplot/admins


Jonarw:
Hmm, seeing it like this somehow makes it feel not quite right to me.
Just semantically speaking, why would an axis with `AxisPosition.None` ever need a margin? How can it need a margin when it has no position?

This makes me think that the problem is more that we assume that axes with `AxisPosition.None` are always polar axes in `Axis.Measure()` - but does this make sense? Maybe we should return a 0-margin by default in `Axis.Measure()` for `AxisPosition.None` and the polar axes should be the ones overriding this behavior?

VisualMelon:
@Jonarw I had the same thought about pushing it down to `Axis`. The color axes have explicit checks in their `Render` overrides as well, which could also be pushed down to `Axis`.

For lack of anything intelligent to say, I'll make that change, and see if anything breaks.

VisualMelon:
OK... that immediately broke Polar plots, of course, but I'm wondering now if the margins on Polar plots make _any_ sense at the moment, and I'm back to your original suggestion of adding an `AxisPosition.All`, which I've pushed here for the time being because that's probably easiest.

We could make `AngleAxisFullPlotArea` use `Position.None`, but there are some rendering bugs at the moment which should probably be fixed first.

Honestly, this all feels too easy, and I'm sure I must be forgetting something, but I can't find anything broken.

Jonarw:
To me this looks quite sensible now. I also can't find any obvious defects.

VisualMelon:
@Jonarw I'll close this an open another issue.

PR #1645, opened 2020-08-14
Remove support for IRenderContext implementations without native clipping

### LINKED ISSUE #1593: Remove support for IRenderContext implementations without native clipping?
Since #1565, all 'official' `IRenderContext` implementations support native clipping (if I didn't miss any). We could think about removing the support for `IRenderContext` implementations without native clipping. 

This would implicate the following changes:

- Change return type of `bool IRenderContext.SetClip(...)` to `void` (or maybe even go one step further and do something like `PushClip(...)` and `PopClip()`)
- Remove the corresponding code paths in `RenderingExtensions`
- Possibly `CohenSutherlandClipping` and `SutherlandHodgmanClipping` could be completely removed?

This would have some advantages:

- We could remove some fairly complex code which is currently unused
- The rendering of some plot elements could be significantly simplified if there was a `IRenderContext.SetClip()` which is guaranteed to succeed (e.g. in the `MagnitudeAxisFullPlotArea`)
- This would open up some opportunities for future enhancements regarding clipping. It would be nice for example (in terms of performance and file size of SVG and PDF files) if clipping was not set and reset for every single series that is rendered, but only once for every group of series with the same axes.

There are two disadvantages I can think of:

- The barrier of entry is slightly higher for new `IRenderContext` implementations
- This would be a breaking change of `IRenderContext.SetClip(...)`

What are your thoughts? Is it worth it?

Jonarw:
Fixes #1593.

### Checklist

- Change the `SetClip(...)` and `ResetClip()` methods in `IRenderContext` to `PushClip(...)` and `PopClip()` and implement a clipping stack in `RenderContextBase`
- Remove the logic in `RenderingExtensions` that provided compatibility to `IRenderContext` implementations that didn't support clipping, also remove `CohenSutherlandClipping` and `SutherlandHodgmanClipping`
- Add an example in '9 Rendering Capabilities' which shows how the clipping is supposed to work
- Make all `IRenderContext` implementations render the '9 Rendering Capabilities' section during unit testing to verify they clip correctly (this also makes sense for this whole section in general I think)

### Remarks

- `SkiaRenderContext` is the only remaining render context not derived from `RenderContextBase`. Instead it uses Skia's clipping stack.

### Issues

- Clipping does not work for `OxyPlot.PdfExporter`. But this is how it was before, so nothing new here. Considering this is deprecated I left it as is.

@oxyplot/admins


Jonarw:
I'll take a look at the failed tests tomorrow, no idea why they didn't fail on my machine... Or maybe I just didn't look carefully enough.

VisualMelon:
I get three failures in `RenderingExtensionsTests`, all of which can just be removed, because they are testing the removed code.

Source/OxyPlot.ImageSharp/ImageRenderContext.cs
```diff
@@ -434,34 +434,22 @@ public override void DrawPolygon(IList<ScreenPoint> points, OxyColor fill, OxyCo
...
+        protected override void ResetClip()
         {
             this.EnsureClippedRegion();
 
             this.clipping = false;
         }
 
         /// <inheritdoc/>
-        public override bool SetClip(OxyRect clippingRectangle)
+        protected override void SetClip(OxyRect clippingRectangle)
```
VisualMelon, line 445: This could be optimised by overriding `Push` and `Pop`, as a `Push` would require ensuring the clipping region, but not blitting back from the image. I don't think this should be a priority for this PR.

Source/OxyPlot/Series/BarSeries/LinearBarSeries.cs
```diff
@@ -158,11 +158,8 @@ public override void Render(IRenderContext rc)
             this.VerifyAxes();
 
             var clippingRect = this.GetClippingRect();
-            rc.SetClip(clippingRect);
```
VisualMelon, line 161: Why is this one not like the others? (i.e. it defers to the `Clipped` method, rather than setting the clipping rectangle here or in `RenderBars` just once). Is this just missing the `using var _ = rc.AutoResetClip(clippingRect);` statement?

(Note: if the push/pop methods check for duplicates, then even with the clipped methods repeating the work, the performance should be better already).

Source/OxyPlot/Series/HistogramSeries.cs
```diff
@@ -137,11 +137,7 @@ public override void Render(IRenderContext rc)
             this.VerifyAxes();
 
             var clippingRect = this.GetClippingRect();
-            rc.SetClip(clippingRect);
```
VisualMelon, line 140: Again here, what am I missing?

Source/OxyPlot/Rendering/RenderContext/RenderContextBase.cs
```diff
@@ -228,21 +230,42 @@ public virtual void CleanUp()
...
+            if (this.clipStack.Count == 0)
+            {
+                throw new InvalidOperationException($"Unbalanced call to {nameof(PopClip)}.");
+            }
+
+            this.ResetClip();
+            this.clipStack.Pop();
+            if (this.clipStack.Count > 0)
+            {
+                this.SetClip(this.clipStack.Peek());
```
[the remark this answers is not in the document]
Jonarw: Yes I think you are right.

Also I think it the implementation in the XKCD decorator should override `PushClip` and `PopClip` directly, rather than `SetClip` and `ResetClip`.

Source/OxyPlot/Series/BarSeries/LinearBarSeries.cs
Jonarw: I didn't want to go into too much detail optimising the clipping calls in this PR. There are probably many places in the code where we could do so. In fact, I think ideally series shouldn't have to be concerned with clipping at all, clipping could be set once for a given combination of X and Y axis and then all series can just do their stuff without having to consider clipping. But I think I'd like to leave this for a future PR.

Anyway I'll change the two places you mentioned as they are really obvious and quickly optimised.

Source/OxyPlot/Rendering/RenderContext/IRenderContext.cs
```diff
@@ -213,17 +213,18 @@ public interface IRenderContext
...
+        /// Pushes the clipping rectangle onto the clipping stack.
         /// </summary>
         /// <param name="clippingRectangle">The clipping rectangle.</param>
-        /// <returns>
-        ///   <c>true</c> if the clipping rectangle was set.
-        /// </returns>
-        bool SetClip(OxyRect clippingRectangle);
+        /// <remarks>
+        /// If there is already a clipping rectangle on the clipping stack, the new clipping rectangle will be intersected with the existing clipping rectangle.
+        /// Calls to this function must be balanced by a call to <see cref="PopClip"/>.
```
VisualMelon, line 221: Nit: 'method' again ;)

Jonarw:
> What happens if an exception is thrown during rendering? We probably need a way to (fully) reset the clipping stack so we can reset it in the event that rendering goes wrong. (This would also resolve an old bug I've never cared to address, where you can't sometimes read the plot exception because it gets clipped).

I think that is a very valid concern. Right now, one unhandled exception can break the `IRenderContext` 'forever'.

One way of dealing with this is to use `AutoResetClip(...)` with `using` everywhere, that way we make sure that clipping is popped even in case of an exception.

But maybe we could just clear the clipping stack in `IRenderContext.CleanUp()` - I think this should avoid any problems in this regard.

VisualMelon:
Oh yeah, I'd forgotten about that.

> One way of dealing with this is to use AutoResetClip(...) with using everywhere, that way we make sure that clipping is popped even in case of an exception.

If we could enforce this, that would be ideal, but that's not really possible. Internally it shouldn't be a problem.

It's possible that the render context could have clipping set up before it is passed to the renderer (maybe it's part of a larger canvas, or it's a sub-plot), which complicates the concept of a CleanUp somewhat.

Jonarw:
The only way we could really enforce the `using` pattern would be to replace the `PushClip(...)` / `PopClip()` API by `IDisposable PushClip(...)`, so the `IDisposable` is actually the _only_ way to pop a clipping. I'm not quite sure how I'd feel about that though.

The alternative (remember clipping state before rendering, restore it in case of exception) seems fairly complicated.

I'll have to consider this some more, input is welcome :)

VisualMelon:
Even if you return the token as part of the API, there is still no way to enforce disposal. The only way to enforce this would be with a `Clip` method that take a callback with a `ref` type parameter, and that would be hideous.

I think any solution is going to add complexity, but we should avoid passing that onto the render contexts if possible. One solution would be wrap the render context with another that manages its own stack, but that's pretty heavy handed.

Jonarw:
> Even if you return the token as part of the API, there is still no way to enforce disposal.

Well if you call a function which returns a `IDisposable` I think it's a safe assumption that people will get the hint that they are supposed to wrap this in a `using`. I think there are even code analysers that give a warning if you fail to do so.
Anyway, I also think that there are legitimate use cases for `PushClip` and `PopClip` instead of using `AutoResetClip`, so I think the `IRenderContext` API should stay as it is.

Maybe the most pragmatic solution would be:

- Replace internal usages of `PushClip` and `PopClip` by `AutoResetClip` as far as possible
- Discourage the direct use of `PushClip` and `PopClip` in favor of `AutoResetClip` in the documentation and recommend to wrap it in a `try...finally` if it is used

VisualMelon:
We probably do need a way to check the depth of the stack though, so that we can detect if the user doesn't `Pop`  the right number of clipping rectangles (and throw an exception), otherwise bugs could go unnoticed and the stack could fill up under normal use. The same mechanism could be used in the exceptional case to reset the stack without complicating the `IRenderContext` implementation.

Jonarw:
(rebased)
I guess you are right. Unfortunately I think there is no way of doing this without changing the `IRenderContext` in one way or the other. I've now added the `ClipCount` property and used it to address the exception-during-rendering issue. I've also added some examples.

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
```diff
@@ -42,8 +42,11 @@ protected virtual void RenderOverride(IRenderContext rc, OxyRect rect)
         {
             lock (this.SyncRoot)
             {
+                var initialClipCount = 0;
+
                 try
                 {
+                    initialClipCount = rc.ClipCount;
```
VisualMelon, line 49: Does this need to be inside the try?

CHANGELOG.md
```diff
@@ -64,6 +64,7 @@ All notable changes to this project will be documented in this file.
 - StemSeries, AreaSeries, TwoColorAreaSeries, and StairStepSeries use `ActualMarkerColor` (#1630)
 - Axes with `AxisPosition.None` make no contribution to margins (#1574)
 - `AngleAxis` has position `AxisPosition.All` by default (#1574)
+- Clipping API changed from SetClip(...) and ResetClip() to PushClip(...) and PopClip() (#1593)
```
VisualMelon, line 67: Should we mention adding `ClipCount`?

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
Jonarw: Well we don't know how other `IRenderContext`s implement this property and if they might throw an exception (because of a faulty implementation).

I guess it is then a question whether we want to catch and render this exception, or throw it. I'd plead for catch and render, as we do with other exceptions during rendering.

VisualMelon: In that case, it might make sense to put a `try...catch` around the poping in the `catch`.

Jonarw: Oh I see what you mean - maybe then it is better to move the assignment outside of the try... If `rc.ClipCount` doesn't work reliably, the exception message might not get properly rendered and it might be better to just throw.

VisualMelon: The problem is that throwing in `Render` has a habit of not showing up for the user to see (i.e. the control dies but the application doesn't crash and burn). That said, `LastException` will be set, so it's not totally undebuggable if something goes wrong, and you'd hope we can write reliable enough code that `ClipCount` isn't throwing regularly ;)

VisualMelon:
@Jonarw yep. I was trying very hard to justify it, but as you say, it makes it less clear that you are meant to dispose it, and the grand plan is to reduce the number of these over time. Those that we do mass-produce will be short lived, so shouldn't be too troublesome.

Jonarw:
Thanks for all your feedback! @VisualMelon

PR #1662, opened 2020-09-20
Optimize clipping calls #1661

### LINKED ISSUE #1661: Optimize Clipping Calls
Currently, every plot element is responsible for its own clipping. Many series use the `DrawClippedXxx(...)` extensions from the `RenderingExtensions` and that way sometimes set and reset clipping multiple times per render ([example](https://github.com/oxyplot/oxyplot/blob/1a6e9222ec5566f03295c97e4e2ecd4ef725182e/Source/OxyPlot/Series/BoxPlotSeries.cs#L329)).
Also, when you render many `Series` sharing the same axes, it wouldn't really be necessary to set and reset clipping every time.

Since #1645, we have some potential to reduce the amount of calls to `IRenderContext.PushClip(...)` and `IRenderContext.PopClip()` by moving the responsibility for the clipping out of the `Series` and `Annotation`s. 

**Specifically I propose to:**
- Remove the `DrawClippedXxx(...)` extensions
- Make `Series` and `Annotation`s completely oblivious to clipping
- Set and reset clipping in [`PlotModel.RenderAxes(...)`](https://github.com/oxyplot/oxyplot/blob/1a6e9222ec5566f03295c97e4e2ecd4ef725182e/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs#L295) and [`PlotModel.RenderAnnotations(...)`](https://github.com/oxyplot/oxyplot/blob/1a6e9222ec5566f03295c97e4e2ecd4ef725182e/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs#L279) respectively, avoiding setting and resetting where possible

**Why would this be desirable?**
- Depending on the platform there could be a more or less significant performance penalty involved with every clipping
- The file size of vector-based exports (especially SVG) can get bloated by many redundant clipping calls
- Rendering code in `Series` and `Annotations` could be simplified in some places, as it wouldn't have to worry about clipping any more

**Potential Problems**
- Removal of the `DrawClippedXxx(...)` extensions will involve some breaking changes for custom implementations. Sure, we could also just keep these, but I'd like to make people aware that they shouldn't be concerned with clipping anymore when implementing custom `Series` / `Annotations`.
- ` CandleStickAndVolumeSeries` - this is a bit of a weird one, because really it's two series in one and therefore needs to set different clipping areas during rendering. I would suggest to do a (hacky) fix for now and mark this as obsolete... I think people should be better off using individual `CandleStickSeries` and `VolumeSeries` anyway.
- (more might emerge once I refine the proof-of-concept implementation I currently have)

I will put together a PR soon so we can talk more specifically about the proposed changes.

Jonarw:
Fixes #1661.

### Checklist

- [ ] I have included examples or tests (there are enough examples demonstration clipping I think)

- Move responsibility for clipping out of the `Render` methods of `Series` and `Annotation`s
- Remove `DrawClippedXxx()` extensions
- Avoid unnecessary clipping calls

### Remarks

- The `PathAnnotation.ClipText` property for obvious reasons was broken by these changes. As it didn't quite seem to fit the overall design anyway and appears not to be very widely used, I decided to remove it. The only place that actually used this property was the Train Schedule example. As I quite liked this example I fixed it using an `Axis` for the station labels, which seems more sensible to me anyway.
- The `CandleStickAndVolumeSeries` could only be fixed by a rather ugly hack. If anything, this indicates to me that this series violates some of the design principles the library is built around, so I applied the hack and marked it as obsolete with the recommendation to use separate `CandleStickSeries` and `VolumeSeries` instead.
- I noticed that `Annotation`s could return different clipping rectangles depending on whether you use `Annotation.GetClippingRect()` or the `ITransposablePlotElement.GetClippingRect()` extension. As this is not an ideal situation, I slightly changed the design here to make it more consistent.

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
+++ b/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
@@ -807,7 +807,7 @@ namespace ExampleLibrary
 
             public override OxyRect GetClippingRect()
             {
-                return new OxyRect(0, 0, double.PositiveInfinity, double.PositiveInfinity);
+                return OxyRect.Everything;
             }
         }
     }
--- a/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
+++ b/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
@@ -804,7 +804,6 @@ namespace ExampleLibrary
             }
 
             /// <inheritdoc/>
-
             public override OxyRect GetClippingRect()
             {
                 return OxyRect.Everything;
--- a/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
+++ b/Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
@@ -802,6 +802,13 @@ namespace ExampleLibrary
                 base.Render(rc);
                 this.Rendering(rc);
             }
+
+            /// <inheritdoc/>
+
+            public override OxyRect GetClippingRect()
+            {
+                return new OxyRect(0, 0, double.PositiveInfinity, double.PositiveInfinity);
+            }
         }
     }
 }
--- a/Source/Examples/ExampleLibrary/Misc/MiscExamples.cs
+++ b/Source/Examples/ExampleLibrary/Misc/MiscExamples.cs
@@ -243,7 +243,6 @@ namespace ExampleLibrary
                 PlotAreaBorderColor = OxyColors.LightGray,
             };
 
-
             var distanceAxis = new LinearAxis
             {
                 Position = AxisPosition.Left,
--- a/Source/Examples/ExampleLibrary/Misc/MiscExamples.cs
+++ b/Source/Examples/ExampleLibrary/Misc/MiscExamples.cs
@@ -23,6 +23,7 @@ namespace ExampleLibrary
     using OxyPlot.Axes;
     using OxyPlot.Series;
     using OxyPlot.Legends;
+    using System.Linq;
 
     [Examples("Misc")]
     public static class MiscExamples
@@ -239,19 +240,42 @@ namespace ExampleLibrary
                 Title = "Train schedule",
                 Subtitle = "Bergensbanen (Oslo-Bergen, Norway)",
                 IsLegendVisible = false,
-                PlotAreaBorderThickness = new OxyThickness(0),
-                PlotMargins = new OxyThickness(60, 4, 60, 40)
+                PlotAreaBorderColor = OxyColors.LightGray,
             };
-            model.Axes.Add(
-                new LinearAxis
-                {
-                    Position = AxisPosition.Left,
-                    Minimum = -20,
-                    Maximum = 540,
-                    Title = "Distance from Oslo S",
-                    IsAxisVisible = true,
-                    StringFormat = "0"
-                });
+
+
+            var distanceAxis = new LinearAxis
+            {
+                Position = AxisPosition.Left,
+                Minimum = -20,
+                Maximum = 540,
+                Title = "Distance from Oslo S",
+                IsAxisVisible = true,
+                StringFormat = "0",
+            };
+
+            model.Axes.Add(distanceAxis);
+
+            var stationAxis = new CustomAxis
+            {
+                MajorGridlineStyle = LineStyle.Solid,
+                MajorGridlineColor = OxyColors.LightGray,
+                Minimum = distanceAxis.Minimum,
+                Maximum = distanceAxis.Maximum,
+                Position = AxisPosition.Right,
+                IsPanEnabled = false,
+                IsZoomEnabled = false,
+                MajorTickSize = 0,
+            };
+
+            distanceAxis.AxisChanged += (sender, e) =>
+            {
+                stationAxis.Minimum = distanceAxis.ActualMinimum;
+                stationAxis.Maximum = distanceAxis.ActualMaximum;
+            };
+
+            model.Axes.Add(stationAxis);
+
             model.Axes.Add(
                 new TimeSpanAxis
                 {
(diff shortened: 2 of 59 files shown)


VisualMelon (commented):
Basically looks good to me. I'm glad that everything still has control over it's clipping rectangle, even if it doesn't set it.

Source/OxyPlot/Annotations/ImageAnnotation.cs
```diff
@@ -310,5 +298,16 @@ protected ScreenVector GetVector(PlotLength x, PlotLength y, PlotModel model)
...
+
+        /// <inheritdoc/>
+        public override OxyRect GetClippingRect()
+        {
+            if (this.X.Unit == PlotLengthUnit.Data || this.Y.Unit == PlotLengthUnit.Data)
+            {
+                return base.GetClippingRect();
+            }
+
+            return new OxyRect(0, 0, double.PositiveInfinity, double.PositiveInfinity);
```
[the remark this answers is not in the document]
Jonarw: > Should we introduce a standard Everything OxyRect (like DataPoint.Invalidate) to which we can compare in the code that sets the clipping rectangle to check it isn't an 'Everything' OxyRect?

Good idea!

> Also, the ImageSharp render context won't cope with out-of-bounds clipping rects

Actually I guess it doesn't have to when we implement your first suggestion.

> Thinking about it further, that would be useful if GetClippingRect where moved to PlotElement, but might also create confusion if it can't be used to splat a rectangle over the whole screen when passed as an argument to a drawing method (which I happen to do in some code I have, though I can't remember why). Perhaps a better option would be to use PlotModel.PlotBounds.

I am not quite sure I get your point here? We do clip to plot bounds anyway [here](https://github.com/Jonarw/oxyplot/blob/41b6c24ebff1bb646bdca62ad6fb3629461b72be/Source/OxyPlot/PlotModel/PlotModel.Rendering.cs#L49), so out-of-bounds clipping rects will be trimmed down [here](https://github.com/Jonarw/oxyplot/blob/41b6c24ebff1bb646bdca62ad6fb3629461b72be/Source/OxyPlot/Rendering/RenderContext/ClippingRenderContext.cs#L49).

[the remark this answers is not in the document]
VisualMelon: That last comment was a bit jumbled, I shouldn't worry about; trimming should cover everything smartly.

Source/OxyPlot/PlotModel/PlotModel.Rendering.cs
```diff
@@ -365,10 +360,38 @@ private void RenderSeries(IRenderContext rc)
...
+            this.RenderPlotElements(this.Series.Where(s => s.IsVisible), rc, series => series.Render(rc));
+        }
+
+        private void RenderPlotElements<T>(IEnumerable<T> plotElements, IRenderContext rc, Action<T> renderAction) where T: PlotElement
+        {
+            OxyRect? previousClippingRect = null;
+
+            foreach (var plotElement in plotElements)
+            {
+                var currentClippingRect = (plotElement as ITransposablePlotElement)?.GetClippingRect();
```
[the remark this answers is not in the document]
Jonarw: > Is there any reason we shouldn't move GetClippingRect to PlotElement? It doesn't make sense for this to be restricted to transposable elements.

I can remember considering this briefly, but for some reason I can't remember I decided against it. But if I think about it now it actually does make a lot of sense, so I'll give it a try.

> I'm a bit unsure that Annotation should implement ITransposablePlotElement directly (Series doesn't). TileMapAnnotation, for example, isn't transpoable, and we have to work around that in ExampleLibrary.Utilities.PlotModelUtilities.

Well `Series` doesn't even have `XAxis` and `YAxis` properties, which `Annotation` does have. I guess we could introduce a `IXyAxisPlotElement` and make `ITransposablePlotElement` inherit from this. `Annotation` would then only implement `IXyAxisPlotElement` and we'd have also have `TransposableAnnotation`. I'll give it a try and see how it looks.

Jonarw:
So this went a bit further than planned, but I do think the design makes more sense now.

Basically I realized that it can be problematic that the `Transform()` and `InverseTransform()` can not be overridden by `Series` and `Annotation`s, as they are defined as static extension methods. This made it kind of difficult to implement your feedback.
If I remember correctly the original goal of this design was to enable code sharing between `Series` and `Annotation`s. But I guess we can also have this with instance methods as long as we keep the actual functionality in a static utility method, which I've done in my latest commit.

Jonarw:
Should we disable/comment out the `TileMapAnnotation` and related examples for now as it is broken until someone puts in some effort to fix it?

VisualMelon:
@Jonarw feel free to disable them for the moment, but I take it there is no reason your changes make it unviable?

I'll and get some stuff done tomorrow (e.g. review this, tidy up the margin PR, fix maps): I'm a bit busy today.

Source/Examples/ExampleLibrary/Examples/RenderingCapabilities.cs
```diff
@@ -802,6 +802,13 @@ public override void Render(IRenderContext rc)
                 base.Render(rc);
                 this.Rendering(rc);
             }
+
+            /// <inheritdoc/>
+
```
VisualMelon, line 807: Weird space

Source/Examples/ExampleLibrary/Misc/MiscExamples.cs
```diff
@@ -239,19 +240,42 @@ public static PlotModel TrainSchedule()
...
-                new LinearAxis
-                {
-                    Position = AxisPosition.Left,
-                    Minimum = -20,
-                    Maximum = 540,
-                    Title = "Distance from Oslo S",
-                    IsAxisVisible = true,
-                    StringFormat = "0"
-                });
+
```
VisualMelon, line 245: More weird spacing

Source/OxyPlot/PlotModel/ITransposablePlotElement.cs
```diff
@@ -9,21 +9,10 @@
 
 namespace OxyPlot
 {
-    using OxyPlot.Axes;
-
     /// <summary>
     /// The TransposablePlotElement interface.
     /// </summary>
-    public interface ITransposablePlotElement : IPlotElement
+    public interface ITransposablePlotElement : IXyAxisPlotElement
```
VisualMelon, line 15: Given we are keeping this, we can probably re-fit the example library `PlotModelUtilites` without the hashsets.

Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
```diff
@@ -55,7 +55,6 @@ public static class PlotModelUtilities
         /// </summary>
         private static readonly HashSet<Type> NonTransposableDataSpaceAnnotationTypes = new HashSet<Type>
         {
-            typeof(TileMapAnnotation),
```
VisualMelon, line 58: They aren't transposable, because we can't draw rotated images.

VisualMelon (commented):
Looks great now! Everything is so much tidier.

Source/Examples/ExampleLibrary/Utilities/PlotModelUtilities.cs
Jonarw: Yes, but `TileMapAnnotation` doesn't implement `ITransposablePlotElement`, therefore it should not even be considered to be transposable. But now that you mention this I do realize this check is never actually made in `PlotModelUtilities`. I will get on that.

Source/OxyPlot/PlotModel/ITransposablePlotElement.cs
Jonarw: For the annotations yes, but for `Series` I think for now we still need to hardcode `CandleStickAndVolumeSeries` and `OldCandleStickSeries`.

VisualMelon (approved):

Jonarw:
(rebased)

Issue #1400, opened 2019-10-20
Move code to oxyplot-contrib


objorke:
## Annotations
- any?

## Axes
- any?

## Series
- PieSeries
- OldCandleStickSeries
- all FinancialSeries?

VisualMelon:
`DateTimeAxis` would seem too important to evict; are there particular issues with it that need addressing?

objorke:
I agree that `DateTimeAxis` is important to have, striking it out now. But I have a feeling that there is a lot of issues related to it and there should probably be a re-design on how it works.

objorke:
`FunctionAxis` is very useful in simple examples and has been very stable. I suggest to keep it.

VisualMelon:
Regarding redesigning how DateTimeAxis works, I have been meaning for ages to implement a mechanism so that you can swap out the tick/label placing logic (i.e. put it behind an interface). This should (hopefully) allow for relatively easy implementation of things like #1241.

It would also be nice to be able to assign the label directly (i.e. rather than indicating where the tick is and letting the formatters deal with it), but that would be a big change. I'll put a proposal together to solicit comments at some point, or maybe I'll just finally get on and implement it.

PR #1667, opened 2020-09-27
Fix OpenStreetMap example (#1642)

### LINKED ISSUE #1642: TileMapAnnotation requires custom user agent on openstreet map.
### Steps to reproduce

1. Open Oxyplot.WindowsForms Example browser
2. Exceptions are thrown when loading the data 429

### Expected behaviour

Should show the map

### Actual behaviour
no Map

### Solution
Add a UserAgent to the request and limit the threads to max 2.

They updated their usage policy
https://operations.osmfoundation.org/policies/tiles/

VisualMelon:
Fixes #1642

### Checklist

- As per the suggestions in #1642 and the OSM policy documentation (https://operations.osmfoundation.org/policies/tiles/)
  - Limits the number of download threads for OSM to 2
  - Adds a configurable UserAgent defaulting to "OxyPlotExampleLibrary"

Setting the user agent is a bit unpleasant, but has been tested with the NET Framework 4.6.1 and NET Core 3.1 WinForms example browser.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ All notable changes to this project will be documented in this file.
 - Axes with `AxisPosition.None` make no contribution to margins (#1574)
 - `AngleAxis` has position `AxisPosition.All` by default (#1574)
 - Clipping API changed from SetClip(...) and ResetClip() to PushClip(...) and PopClip() (#1593)
+- Remove TileMapAnnotation examples from automated testing (#1667)
 
 ### Removed
 - Remove PlotModel.Legends (#644)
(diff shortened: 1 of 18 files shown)


VisualMelon:
I'm not sure what the rules on User-Agents for non-browsers are; someone with some awareness of web technologies might want to take a look. I'm also tempted to suggest we remove the tile map examples for the automated tests regardless of whether they work or not, because they depend on an external resource (and the downloading is the only bit isn't covered by other components).

Jonarw:
Yes I think it is a good idea to remove them from unit tests - if for some reason the tile server is not reachable we don't want our tests to fail because of that.

Unfortunately, I am not very knowledgeable about web technologies, but as long as it works I guess it's fine.

VisualMelon:
@Jonarw had a go at that. It's a bit ugly, but I don't know how much that matters.

Jonarw:
Looks fine to me, good to go from your side?

VisualMelon:
@Jonarw If you don't think the `ForAutomatedTest` postfix is too hideous, I have nothing more to add.

VisualMelon:
Actually, let me fix a couple of formatting things...

objorke:
See also #1400

PR #1669, opened 2020-09-27
Add netstandard2.0 target

### LINKED ISSUE #1668: Add netstandard2.0 target?
I recently studied this page for an unrelated project:
https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/cross-platform-targeting

Especially this point I think is relevant for us:

> DO include a netstandard2.0 target if you require a netstandard1.x target.
> All platforms supporting .NET Standard 2.0 will use the netstandard2.0 target and benefit from having a smaller package graph while older platforms will still work and fall back to using the netstandard1.x target.

That is something that I actually did notice in the past, but never thought about it too much: when referencing OxyPlot from .NET Core a lot of Microsoft packages are added in the process. 
I guess it would be nice to avoid this, so I would propose to add the `netstandard2.0` target to the core library (and to all other libraries targeting any `netstandard1.x`, e.g. `OxyPlot.SkiaSharp`).

---

By the way, I think this other point should already be covered by our `net45` target:
> CONSIDER adding a target for net461 when you're offering a netstandard2.0 target.
> Using .NET Standard 2.0 from .NET Framework has some issues that were addressed in .NET Framework 4.7.2. You can improve the experience for developers that are still on .NET Framework 4.6.1 - 4.7.1 by offering them a binary that is built for .NET Framework 4.6.1.

Jonarw:
Fixes #1668 .

### Checklist

- Add `netstandard2.0` target to `OxyPlot`, `OxyPlot.SkiaSharp`, `OxyPlot.ImageSharp` and `ExampleLibrary`

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ All notable changes to this project will be documented in this file.
 - IRenderContext.ClipCount property (#1593)
 - Additional parameters for HistogramSeries LabelFormatString
 - Absolute screen-space axis margins (#1569)
+- netstandard2.0 TargetFramework (#1668)
 
 ### Changed
 - Legends model (#644)
--- a/Source/Examples/ExampleLibrary/ExampleLibrary.csproj
+++ b/Source/Examples/ExampleLibrary/ExampleLibrary.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;net45</TargetFrameworks>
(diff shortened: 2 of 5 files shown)


VisualMelon:
We can ignore snyk. I think so long as this builds and runs then it's good to ship.

objorke:
I have tried to deactivate the snyk scan, let us see on the next PR...

VisualMelon:
@objorke it's not a problem; I only commented to documented that it was expected and it wasn't highlighting an actual concern (i.e. I'd personally checked the manifest changes weren't dodgy).

Issue #107, opened 2014-08-14
Vector field series


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/10127](http://oxyplot.codeplex.com/workitem/10127)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2014-02-04 19:15
add a series that shows a vector field 

<a href="http://en.wikipedia.org/wiki/Vector_field" rel="nofollow">http://en.wikipedia.org/wiki/Vector_field</a>

VisualMelon:
It wouldn't be hard to build one of these working from HeatMap as a basis: swap out `double[,] Data` for `DataVector[,] Data` (where `DataVector` is a new struct based on `ScreenVector` and `DataPoint`).

I have written such things in the past when I've needed them, but never tidied them properly. What requirements would there be for a shippable `VectorFieldSeries`? Obvious questions:

 - Colors: Single color or ColorAxis (from Vector length?)? (or both?)

 - Vector rendering: `DataVector` would imply X and Y components are in Data Space; but this means a (1, 1) vector won't always be rendered at an angle 45degrees (e.g. if one axis is stretched): if not Data Space vectors then what? (How do they scale, etc.?)

 - Vector Positioning: individual placement, or HeatMap style grid? (Two different series?)

It may be worth having a more bulky `VectorFieldItem` type which can allow a custom `Value` and `Color`; norms may be inadequate, color might be a function of angle. Helper methods can provided to bypass complexities for regular use-cases.

As I said, I use Vector Fields from time to time for very specific things, so I never have to make any design decisions; I'd be glad to write a 'good' implementation from a rough spec and arrange a Pull Request when I've worked out how to use git properly.

VisualMelon:
Assigning myself, because I should probably get around to doing this at some point.

Jonarw:
A couple of thoughts (some or all of this may be non-sense):

- Could this be implemented as a `ScatterSeries<VectorPoint>`?
- I think the angle should be specified explicitly, while X and Y data components specify the location
- `VectorPoint` could have the following properties (in addition to those inherited by `ScatterPoint`):
    - `Angle`
    - `Color`
- The `Value` would specify the length and could be used with a `IColorAxis` for color (overridden by `Color` property)
- There would have to be some property / conversion factor on the Series which would determine how `Value`s are mapped to an actual length (in render units)
- By default this conversion factor could automatically be determined by normalizing

PR #1672, opened 2020-09-29
VectorSeries

VisualMelon:
Working on #107, putting this here so that anyone who wants to explain all the things wrong with it can do so, because I've not touched it recently and am making no progress, and would appreciate any feedback. There are also a couple of things in here I find myself wanting to point at on occasion.

### Checklist

- Add a `VectorSeries`, which plots a series of `VectorItems` as arrows on an X/Y axis pair
   - `VectorItems` have an `Origin`, `Direction`, and `Value` (i.e. it's a vector with some arbitrary value positioned at a particular point)
   - The value is mapped to a color on a `ColorAxis`
- Adds a `DataVector` struct
  - I would like to go through the existing code base and replace usage of `DataPoint` as a vector with `DataVector`, though I'm not sure there are any instances (I need to check)

I'm not entirely happy with rendering, but that might be because I'm never entirely happy.

It doesn't make plotting a vector field especially easy, because it just plots arrows. We could either provider helpers to map a vector field (i.e. 2D array of vectors) to a list of `VectorItems`, or we could make `VectorItem` optionally read from a multidimensional array, or we could provide a separate `VectorFieldSeries` to do that.

I opted for X/Y vector rather than Angle/Mag because the former would be confusing wrt. the plot aspect ratio: if a user wants angle/mag, then they should be fixing the aspect ratio. (Note: that isn't easy...)

Have I totally missed the point?

@oxyplot/admins


Jonarw:
I'm not sure I can completely get behind the idea of defining the vectors using X/Y coordinates rather than Magnitude/Angle. This might have to do with the only kind of vector fields I ever really was in contact with - which is optical representations of magnetic/electric fields, something like [here](https://pages.vassar.edu/magnes/advanced-electromagnetism-phys-341/rachel-wagner-kaiser/).

Here it doesn't make sense to specify the vector in terms of X/Y coordinates, because it represents a completely different thing: while the X/Y coordinates represent a point in 2-dimensional space, the vector length and direction represents the field strength and direction. Consequently it also doesn't make sense to change the direction or length of the vector when the X or Y axes are scaled. Also in this context it makes no sense for the vector to become curved on logarithmic axes.

Of course there might be other applications I am unaware of where the vector is actually directly related to the X and Y coordinates - did you have such an application in mind?

VisualMelon:
@Jonarw thanks for taking a look. I don't have any particular application in mind (I did 3  or 4 years ago, but I can't remember what; I threw something together at the time and moved on).

I completely understand your point about the magnitudes being meaningful. The vector field with angle/mag only makes sense on a set of well aligned and scaled axes, in which case the representation doesn't matter: a vector is just a vector, and the user could use a `FromAngleAndMagnitude` method to create them in if suits their use case and enforce a pair of well aligned and scaled axes.
 - With an angle/mag representation, the magnitude is always meaningful, but the angle will be misleading on a pair of mis-aligned/scaled axes.
 - With a dx/dy representation, the direction is always meaningful, but the magnitude will be misleading on a pair of mis-scaled axis

On the other hand, a plot of some arbitrary differential function doesn't necessarily warrant an angle/magnitude interpretation at all. If interpretability is a concern then there should be a constraint on the axes to enforce interpretability, regardless of the representation.

Regarding X/Y being coordinates, this is why I introduce the `DataVector` type, which is explicitly _not_ a point, but rather a vector. The same has existed for ages with `ScreenPoint`/`ScreenVector`.

Regarding curves on log/other axis, I mostly implemented this for fun and to explore alternative to the current algorithm in `LineAnnotation`. It doesn't make sense with an angle/mag interpretation, but I don't know that an angle/mag interpretation could ever make sense on a pair of non-linear 2D axes.

To sum up, I think that either representation is problematic on mis-aligned/scaled axes, which leads to the conclusion that we should have a more powerful constraints mechanism to enforce consistency between axes. It's very hard to enforce constraints at the moment beyond the `Cartesian`: we should this as an extensibility point. In my opinion, x/y is the more natural representation of a vector, but I'm willing to be convinced otherwise.

Could it make sense to have 2 series?
 - one with a data-space vector representation (would it be better to just have a more general purpose stream plot?)
 - one with a angle/mag vector representation

Jonarw:
Hmm yes I see your point. I'll have to think about it some more.
In the meantime a (possibly stupid) idea: How about the the vector was specified by magnitude and angle, where the magnitude would be specified in screen space (and therefore unaffected by axes) and the angle in data space, so it would still be meaningful when the axes not cartesian. This would in theory solve the problem, but at the moment I am not sure if it wouldn't be too confusing or if makes any sense at all.

VisualMelon:
@Jonarw I might change it so that it is angle/mag, but the mag is data/screen space configurable to see how that feels. I really must take another look at matlab/matplotlib to see what people expect as well.

cliff-lane:
Has there been any more thought put into this project?   I am in need of a plotting library that includes Contour Plots, Histograms, and Vector Fields.  It appears that OxyPlot has everything I need with the exception of the finalized Vector Fields.  The data I need to plot is basically misalignment data where I have a (x,y) of where the point should be and a (x,y) of how far off the point is from where it should be.

VisualMelon:
Sorry, no: not worked on this at all.

For your use-case, sounds like it would make sense to have a dx/dy vector description, which is what the draft currently implements. You should be able to mostly C&P the implementation if you want to mess around with it: there's a lot of junk in the PR that shouldn't be there and can be ignored.

VisualMelon:
Sorry, no: not worked on this at all.

For your use-case, sounds like it would make sense to have a dx/dy vector description, which is what the draft currently implements. You should be able to mostly C&P the implementation if you want to mess around with it: there's a lot of junk in the PR that shouldn't be there and can be ignored.

tylerdoboy:
Will this become available before next year? I have been implementing my own vector field but it would be nice to have this series available.

VisualMelon:
@tylerdoboy seems there is a fair bit of interest in a vector series, so that'll push it higher in my list of things TODO if I have time.

Please do share any details you can about your use-case: would the design here work for you? Are there any features you need that are missing? etc. this will help with the design and let us know when we have something worth including in the library.

VisualMelon:
Rebased, tidied a bit, and added the ability to position vector arrows/text across along their length to support a vector-field scenario.

If anyone has any fun examples they'd like to see implemented to ensure they have good support, let me know.

VisualMelon:
Do we need a nice means for plotting a 2D vector field? (current API demands you provide it with per-vector `VectorItem`s.

This could be a helper that generates a list of points from the 2D array, or it could be an alternative to `Items` in the series itself, or it could be implemented as a special ItemSource.

VisualMelon:
Definitely need to do some tidying here, and add a more sensible example, but reasonably happy with the API as is. Need to review the data-bounds logic as well.

VisualMelon:
Current data-range is defined by the 'Origin' of each `VectorItem`; this means that the arrows may be clipped. We could change this so that it's the min/max point where a rendered arrow starts or stops, but I'm not completely sold on that idea.

Source/OxyPlot/Foundation/DataPoint.cs
```diff
@@ -93,6 +93,98 @@ public bool Equals(DataPoint other)
...
+            if (!notParallel)
+            {
+                intersection = default;
+                return false;
+            }
+
+            intersection = line1Start + line1Dir * lambda1;
+            return lambda0 >= 0 && lambda0 <= 1 && lambda1 >= 0 && lambda1 <= 1;
+        }
+
```
Jonarw: I'm not sure if these function really belong in the `DataPoint` class - maybe they would be better suited in a separate static helper class? I am not sure if we already have one where this could fit, it not maybe we could add one?

Source/OxyPlot/Foundation/DataVector.cs
```diff
@@ -0,0 +1,199 @@
...
+        [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1307:AccessibleFieldsMustBeginWithUpperCaseLetter", Justification = "Reviewed. Suppression is OK here.")]
+        // ReSharper disable once InconsistentNaming
+        internal double x;
+
+        /// <summary>
+        /// The y-coordinate.
+        /// </summary>
+        [SuppressMessage("StyleCop.CSharp.NamingRules", "SA1307:AccessibleFieldsMustBeginWithUpperCaseLetter", Justification = "Reviewed. Suppression is OK here.")]
+        // ReSharper disable once InconsistentNaming
+        internal double y;
```
Jonarw: Do we need these to be accessible internally? Couldn't they just be private?

Source/OxyPlot/Foundation/DataVector.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// </remarks>
+        public void Normalize()
+        {
+            double l = Math.Sqrt((this.x * this.x) + (this.y * this.y));
+            if (l > 0)
+            {
+                this.x /= l;
+                this.y /= l;
+            }
+        }
```
Jonarw: I always have a bad feeling about mutable `struct`s. I think I would like it more if this returned a new struct.
Also this should use the `Length` property I reckon.

Source/OxyPlot/Foundation/DataVector.cs
```diff
@@ -0,0 +1,199 @@
...
+        /// Determines whether this point is defined.
+        /// </summary>
+        /// <returns><c>true</c> if this point is defined; otherwise, <c>false</c>.</returns>
+        public bool IsDefined()
+        {
+            // check that x and y is not NaN (the code below is faster than double.IsNaN)
+#pragma warning disable 1718
+            // ReSharper disable EqualExpressionComparison
+            // ReSharper disable CompareOfFloatsByEqualityOperator
+            return this.x == this.x && this.y == this.y;
```
Jonarw, line 177: I am not quite sure if that comparison is actually faster than `double.NaN` (anymore?), however I am fairly sure that it wouldn't really matter in terms of magnitude even if there was a difference... so I think I would prefer `double.NaN` here to avoid all the surrounding clutter.

Source/OxyPlot/Rendering/ScreenPoint.cs
```diff
@@ -117,6 +117,63 @@ public static bool IsUndefined(ScreenPoint point)
...
         }
 
+        /// <summary>
+        /// Determines whether the given values is a real fine value.
+        /// </summary>
+        /// <param name="f">The number to evaluate.</param>
+        /// <returns><c>true</c> if <paramref name="f"/> is non-NaN and non-Infinite, otherwise <c>false</c>.</returns>
+        private static bool IsFinite(double f)
+        {
+            return !(double.IsNaN(f) || double.IsInfinity(f));
```
Jonarw: Does this do something different than `double.IsFinite`?

Source/OxyPlot/Rendering/ScreenPoint.cs
```diff
@@ -117,6 +117,63 @@ public static bool IsUndefined(ScreenPoint point)
...
+
+            if (!notParallel)
+            {
+                intersection = default;
+                return false;
+            }
+
+            intersection = line1Start + line1Dir * lambda1;
+            return lambda0 >= 0 && lambda0 <= 1 && lambda1 >= 0 && lambda1 <= 1;
+        }
```
Jonarw: I don't really like the fact that these are duplicated between `ScreenPoint` and `DataPoint`, but I suppose the same could also be said about other functionality that is common to `ScreenPoint` and `DataPoint`. Maybe we should consider adding something like an `IPoint` interface in the future and implement stuff like this as generic static functions. But I guess this might be out of scope of this PR, so I'm happy if it stays like this for now

Source/OxyPlot/Rendering/ScreenVector.cs
```diff
@@ -134,6 +134,9 @@ public double Y
         /// <summary>
         /// Normalizes this vector.
         /// </summary>
+        /// <remarks>
+        /// If the vector has zero length, then it is left unchanged.
+        /// </remarks>
         public void Normalize()
```
Jonarw, line 140: I see that there is some precedent for the mutable `struct` stuff, and also for pretty much all the other stuff I didn't like in `DataVector`... And unfortunately, changing `ScreenVector.Normalize` to be non-mutable would be a breaking change... But maybe one we would be willing to take?

Source/OxyPlot/Series/VectorItem.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// Determines whether this Vector item is defined.
+        /// </summary>
+        /// <returns><c>true</c> if this point is defined; otherwise, <c>false</c>.</returns>
+        public bool IsDefined()
+        {
+            // check that x and y is not NaN (the code below is faster than double.IsNaN)
+#pragma warning disable 1718
+            // ReSharper disable EqualExpressionComparison
+            // ReSharper disable CompareOfFloatsByEqualityOperator
+            return this.Origin.IsDefined() && this.Direction.IsDefined() && !double.IsNaN(this.Value);
```
Jonarw, line 89: I think the surrounding comments are not really applicable - Copy&Paste?

Jonarw (commented):
Some smaller comments just from looking at the code changes, I will play around with the example later.

> Current data-range is defined by the 'Origin' of each VectorItem; this means that the arrows may be clipped. We could change this so that it's the min/max point where a rendered arrow starts or stops, but I'm not completely sold on that idea.

I guess arguments could be made for both, I am not quite sure yet on which side I am on this ;)

Source/OxyPlot/Foundation/DataVector.cs
VisualMelon: I completely agree: this is a hideous pattern and I can't believe it will be faster these days than `double.IsNaN`

VisualMelon:
@Jonarw lots of comments here about `DataVector` that apply equally to `ScreenVector`: I'd suggest we handle those in a separate PR where we update both to be less hideous.

Jonarw:
> @Jonarw lots of comments here about `DataVector` that apply equally to `ScreenVector`: I'd suggest we handle those in a separate PR where we update both to be less hideous.

Yes, fine with me!

Source/OxyPlot/Rendering/ScreenVector.cs
VisualMelon: Yep, unfortunate history here. I think we should make `DataVector` immutable whether `ScreenVector` is or not (as you say, mutable struct semantics are ludicrously complicated), and probably mark this method as obsolete.

Source/OxyPlot/Rendering/ScreenPoint.cs
VisualMelon: I don't think so.

Source/OxyPlot/Series/VectorItem.cs
VisualMelon: Fixed this in VectorItem and RectangleItem (another C&P job). Also changed the definition of `Equals` in both: this is a breaking change for `RectangleItem` but the old versions looks wrong to me.

Source/OxyPlot/Rendering/ScreenVector.cs
VisualMelon: Removed the method in `DataVector` and made `x` and `y` readonly for the time being so there's no risk of not getting around to that once this goes in.

VisualMelon:
Regarding examples, they're pretty terrible at the moment, but I'm happy to implement any more exciting/useful ideas you might have.

Source/OxyPlot/Rendering/ScreenPoint.cs
VisualMelon: These statics aren't actually being used; I'm not sure why I added them. I will remove them.

Source/OxyPlot/Foundation/DataVector.cs
VisualMelon: Will leave as is for now; another thing we could change throughout the code-base. Certainly no point in having them if we want the structs to be immutable.

VisualMelon:
I've marked the bits that are not vector-series specific as resolved for now to keep things clear for this PR; observations so far:
 - Would prefer everything is immutable
 - If everything is immutable, not point having `interval` fields for x/y
 - Normalise should use `Length` rather than computing it independently
 - Use `IsNaN` instead of `x == x` everywhere

Source/OxyPlot/Series/VectorSeries.cs
```diff
@@ -0,0 +1,569 @@
...
+        /// </summary>
+        /// <value>The mapping.</value>
+        /// <remarks>Example: series1.Mapping = item => new VectorItem(new DataPoint((MyType)item).Time1, ((MyType)item).Value1), new DataPoint((MyType)item).Time2, ((MyType)item).Value2));</remarks>
+        public Func<object, VectorItem> Mapping { get; set; }
+
+        /// <summary>
+        /// Gets the list of Vectors.
+        /// </summary>
+        /// <value>A list of <see cref="VectorItem" />. This list is used if <see cref="ItemsSeries.ItemsSource" /> is not set.</value>
+        public List<VectorItem> Items { get; } = new List<VectorItem>();
```
Jonarw: I would prefer to expose an `IList<VectorItem>`. I know we expose `List`s elsewhere, so it would be somewhat inconsistent, but as far as I am concerned it is still preferable.

Source/OxyPlot/Series/VectorSeries.cs
```diff
@@ -0,0 +1,569 @@
...
+                this.actualItems.AddRange(sourceAsEnumerableDataRects);
+            }
+        }
+
+        /// <summary>
+        /// Renders the points as line, broken line and markers.
+        /// </summary>
+        /// <param name="rc">The rendering context.</param>
+        /// <param name="items">The Items to render.</param>
+        protected void RenderVectors(IRenderContext rc, ICollection<VectorItem> items)
```
Jonarw: `items` could be `IEnumerable<>`

Jonarw (approved):
I have two small, somwhat opinion-based, comments about the API, otherwise I think this is good to merge!

The Examples are sufficient to see that everything works as it should, so they fulfill their purpose I think.

VisualMelon:
I agree with both of those, and relaxed the protected `ActualItems` also (should it be `IReadOnlyList`?): no reason we can't use more sensible types on new series, and if we later decide to change to the concrete type then it's not a code breaking change.

_**Edit**: can't be IReadOnlyList because IList doesn't implement it_

VisualMelon:
Forgot the changelog...

VisualMelon:
Hopefully this is in a mergeable state, but I'd rather we didn't merge it just yet, as I'd like to give it another review myself in a couple of days when my head is clearer.

VisualMelon:
Rebased. I don't have any more changes I want to make now.

Jonarw (approved):

VisualMelon:
Rebased

Jonarw (approved):

Issue #1652, opened 2020-08-29
Axis.ScreenMin and Axis.ScreenMax are weird


VisualMelon:
Horizontal and Vertical axes do not provide useful values for `ScreenMin` and `ScreenMax`. They suffice for finding the start and end of the axis in the dimension in which they exist, but do not provide any information about the other dimensions. The odd piece of code is: https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/Axes/Axis.cs#L1372-L1373

            this.ScreenMin = new ScreenPoint(a0, a1);
            this.ScreenMax = new ScreenPoint(a1, a0);

Existing code ignores this weirdness because it all explicitly checks whether the axis is horizontal or vertical before querying the min/max values.

Note that the bounds the in the other dimensions are not known when these bounds are set: they are only known after rendering, when the margins from other tiers are known. As such, there is no obvious solution, but it would make sense to me to update the ScreenMin and ScreenMax after rendering. This will help with simplifying axis hit-testing (which is performed inside `PlotModel` at the moment). The information that is 'missing' from the screen min and max is ultimately recorded in `PositionTierMinShift` and `PositionTierMaxShift`, relative to the plot bounds. Even these are `internal`, so stuff like #1651 is impossible to do reliably.

The issue of when these things should be updated is related to #1621.

Jonarw:
Fully agree that this is 'weird' to say the least.

Maybe this could be rectified as part of a potential larger refactoring of the Axes code as in #1640?

PR #1682, opened 2020-10-06
Potential fix for ScreenMin and ScreenMax on Horizontal and Vertical Axes

VisualMelon:
Addresses #1652 .

### Checklist

- Make the ScreenMin and ScreenMax of horizontal and vertical axes depend on their own clipping region and the plot bounds.

This makes much more sense than what they currently do, and the rendering code is already only using the 'correct' components of the two values. It may be possible to simplify some code (e.g. gridline cropping) as now the cropping region for the axis pair is just the intersection of the rectangles determined by ScreenMin/ScreenMax.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -112,6 +112,7 @@ All notable changes to this project will be documented in this file.
 - Incorrect margins when using Color Axes with AxisPosition.None (#1574)
 - OpenStreetMap example (#1642)
 - Incorrect clipping in TwoColorAreaSeries (#1678)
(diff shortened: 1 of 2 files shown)


Jonarw:
Yes I think this makes much more sense now.
Also it doesn't appear to have any side effects - probably because, as you said, the rendering code already only uses the values that made sense before.

I think this should be good to go if you are still happy.

VisualMelon:
@Jonarw yep, I'm still happy.

Issue #1673, opened 2020-09-30
Performance WPF


kyooi93:
### Steps to reproduce

1. Create ScatterSeries Graph
2. Generate around 5000 to 10000 point

Platform:
.NET version:

### Expected behaviour

1. Zoom in and out with smooth behavior
2. Auto size the graph 
### Actual behaviour

1 .when zoom in and out its hang and laggy. make whole program slow and hang.
2. Not resize the graph when plotting

![image](https://user-images.githubusercontent.com/71051053/94643255-f6399780-0318-11eb-8177-a7ea5326d571.png)

VisualMelon:
#1609 should help with the performance side. Marker performance is known to be less than fantastic in WPF (and Avalonia).

What do you mean about it not resizing the plot?

kyooi93:
> #1609 should help with the performance side. Marker performance is known to be less than fantastic in WPF (and Avalonia).
> 
> What do you mean about it not resizing the plot?

kyooi93:
> #1609 should help with the performance side. Marker performance is known to be less than fantastic in WPF (and Avalonia).
> 
> What do you mean about it not resizing the plot?

kyooi93:
> > #1609 should help with the performance side. Marker performance is known to be less than fantastic in WPF (and Avalonia).
> > What do you mean about it not resizing the plot?

The resize mean when the graph was plotted with data , the graph cant be auto zoom out to view the graph as overview.
Let me show you picture, will be more clearer.

**After Plot with Data**
![image](https://user-images.githubusercontent.com/71051053/94651429-05294580-032b-11eb-9721-c490fba0571a.png)

**Have to zoom out manually** 
![image](https://user-images.githubusercontent.com/71051053/94651482-18d4ac00-032b-11eb-91e3-6ac03b97a393.png)

kyooi93:
If want to plot the graph like this what method can use to improve the performance ?

VisualMelon:
Ah, right. You need to explicitly ask the plot to reset the view with a call to `PlotModel.ResetAllAxes();`

With regards to improving performance, I'm afraid I can't suggest anything. I'm not overly familiar with the WPF renderer. There is a new immediate-mode renderer based on Skia Sharp (which is generally much snappier), but that's only available in pre-release CI packages at the moment.

kyooi93:
> Ah, right. You may need to explicitly ask the plot to reset the view with a call to `PlotModel.ResetAllAxes();`
> 
[...]

PR #1687, opened 2020-10-10
Improve performance of CanvasRenderContext

Jonarw:
Improves #1673.

This PR is based on #1609 where @imalip did the heavy lifting. I fixed some minor issues and moved the non-`StreamGeometry` code into its own class (`XamlRenderContext`).

### Checklist

- Improve performance of `CanvasRenderContext` by using `StreamGeometry` as much as possible instead of drawing individual paths
- Add `PlotView.TextMeasurementMethod` property to `OxyPlot.WPF.PlotView` to allow using the much faster `GlyphTypeface`-based measurement
- As stuff that is drawn using `StreamGeometry` is not serializable, we need to keep the 'regular' rendering functions for the `XamlExporter` to work properly. Previously this was integrated into `CanvasRenderContext` and controlled by the `UseStreamGeometry` property, but that was a bit messy and made the class needlessly complicated. So I decided to actually move this code into its own class, `XamlRenderContext`, which is only used by the `XamlExporter`.

### Notes

- To make the new `XamlRenderContext` testable, I added the renderer 'Canvas_XAML' to the WPF ExampleBrowser
- As line joins were problematic initially with `StreamGeometry`, I added a `LineJoin` example to the rendering capabilities section.
_sidenote: This revealed that the `SkiaSharp` renderer actually ignores `LineJoin=Miter` for small angles. Fix to follow._
- I would have loved to resurrect the Benchmark functionality of the WPF ExampleBrowser to actually see some numbers regarding performance. But I couldn't find a straight-forward way to make this work with both Canvas and Skia. @imalip did some Benchmarks in #1609. Also the difference is quite noticeable with certain examples (`Performance>LineSeries, 100k points, markers` and `Polar Plots>Spiral full plot area` for example).
- I did not make any effort optimizing the `XamlRenderContext`. I have no idea what the `DrawLineBalanced` stuff is all about, so I didn't touch it. But I guess that's ok, because it's for XAML export only.

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.Wpf/CanvasRenderContext.cs
+++ b/Source/OxyPlot.Wpf/CanvasRenderContext.cs
@@ -219,9 +219,48 @@ namespace OxyPlot.Wpf
                 return;
             }
 
+            if (this.UseStreamGeometry)
+            {
+                this.DrawLineByStreamGeometry(points, stroke, thickness, edgeRenderingMode, dashArray, lineJoin);
+                return;
+            }
+
             var e = this.CreateAndAdd<Polyline>();
             this.SetStroke(e, stroke, thickness, edgeRenderingMode, lineJoin, dashArray, 0);
             e.Points = this.ToPointCollection(points, e.StrokeThickness, edgeRenderingMode);
+        }
+
+        /// <summary>
+        /// Draws a polyline using a StreamGeometry.
+        /// </summary>
+        /// <param name="points">The points defining the polyline. The polyline is drawn from point 0, to point 1, to point 2 and so on.</param>
+        /// <param name="stroke">The stroke color.</param>
+        /// <param name="thickness">The stroke thickness (in device independent units, 1/96 inch).</param>
+        /// <param name="edgeRenderingMode">The edge rendering mode.</param>
+        /// <param name="dashArray">The dash array (in device independent units, 1/96 inch). Use <c>null</c> to get a solid line.</param>
+        /// <param name="lineJoin">The line join type.</param>
+        public void DrawLineByStreamGeometry(
+            IList<ScreenPoint> points,
+            OxyColor stroke,
+            double thickness,
+            EdgeRenderingMode edgeRenderingMode,
+            double[] dashArray,
+            LineJoin lineJoin)
+        {
+            var path = this.CreateAndAdd<Path>();
+            this.SetStroke(path, stroke, thickness, edgeRenderingMode, lineJoin, dashArray);
+
+            var actualStrokeThickness = this.GetActualStrokeThickness(thickness, edgeRenderingMode);
+            var actualPoints = this.GetActualPointsForStreamGeometry(points, actualStrokeThickness, edgeRenderingMode);
+
+            var streamGeometry = new StreamGeometry();
+            using (var streamGeometryContext = streamGeometry.Open())
+            {
+                streamGeometryContext.BeginFigure(actualPoints.Item1, false, false);
+                streamGeometryContext.PolyLineTo(actualPoints.Item2, !stroke.IsUndefined(), true);
+            }
+
+            path.Data = streamGeometry;
         }
 
         ///<inheritdoc/>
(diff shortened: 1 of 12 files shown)


Source/OxyPlot.Wpf/CanvasRenderContext.cs
```diff
@@ -136,20 +107,43 @@ public override void DrawEllipse(OxyRect rect, OxyColor fill, OxyColor stroke, d
...
+                    const double arcAsBezier = 0.551784;
+
+                    var centerX = rect.Center.X;
+                    var centerY = rect.Center.Y;
+
+                    var midX = Math.Abs(rect.Width / 2.0) * arcAsBezier;
+                    var midY = Math.Abs(rect.Height / 2.0) * arcAsBezier;
+
+                    sgc.BeginFigure(new Point(rect.Right, centerY), isFilled, true);
+                    sgc.BezierTo(new Point(rect.Right, centerY + midY), new Point(centerX + midX, rect.Bottom), new Point(centerX, rect.Bottom), isStroke, true);
```
VisualMelon, line 138: I can see how this may be preferable for markers and such as a potential performance consideration, but I'd prefer there was an option like `UseFastEllipseApproximation` or something, and that this used arcs by default for precision. Is this measurably faster anyway? I'd guess that the savings from using streaming geometry are all that really matters here.

imalip:
Hi Jonathan,

Sorry I didn't reply, been feeling a bit under the weather after a full month of solid work, thanks for integrating the patches.

Source/OxyPlot.Wpf/CanvasRenderContext.cs
VisualMelon: Ah, you can't draw oblong ellipses with StreamingGeomtry can you, only Circles...

Jonarw: I didn't really question this to be honest. You can actually draw arcs using `StreamGeometry`, so you should be able to draw a full ellipse exactly by drawing two arcs. I will give that a try and check if there is a noticeable performance difference.

Jonarw:
No problem at all! Thank you for this contribution!

Source/OxyPlot.Wpf/CanvasRenderContext.cs
VisualMelon: Ah, great: I didn't realise you could specify different radii for arcs.

    sgc.ArcTo(new Point(rect.Left, centerY), new Size(midX, midY), 180, false, SweepDirection.Clockwise, isStroke, true);
    sgc.ArcTo(new Point(rect.Right, centerY), new Size(midX, midY), 180, false, SweepDirection.Clockwise, isStroke, true);

I can't perceive much difference between the two; the Bezier's perhaps feel a little snappier; I've not tried to properly measure anything.

imalip: Hi,
For the record, I didn't try using arcs and went straight for Bezier curves, because that's what is used internally in WPF to draw an ellipse geometry. Did you notice any noticeable inaccuracy using an ellipse geometry ?

VisualMelon: > that's what is used internally in WPF to draw an ellipse geometry

Could you provide a reference for that? I was looking for some implementation earlier, but couldn't find it (though I didn't look too hard). The `Arc` method seems to be produce the same results in the `CanvasRenderContext` as the 'old' `EllipseGeometry` based method does in the new `XamlRenderContext`.

> Did you notice any noticeable inaccuracy using an ellipse geometry

No, and for most ellipses it will be imperceptible, but for large ones it will mean differences between renderers and potentially hit-testing, and if precision is needed for some reason then it may not be acceptable. If it makes markers 2* faster, however, then there is an obvious benefit to be had ;)

imalip: > 
> 
> > that's what is used internally in WPF to draw an ellipse geometry
> 
> Could you provide a reference for that? I was looking for some implementation earlier, but couldn't find it (though I didn't look too hard). The `Arc` method seems to be produce the same results in the `CanvasRenderContext` as the 'old' `EllipseGeometry` based method does in the new `XamlRenderContext`.

https://github.com/dotnet/wpf/blob/master/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/EllipseGeometry.cs
All the path calculations start at line 272.
Although I chose a different factor ; but I have no strong feelings about it. IIRC the value used in WPF ensures all points are within the circle, the one I picked minimises the absolute error.

VisualMelon: Thanks for the link. It looks like [oblong arcs are approximated with Beziers anyway](Https://github.com/dotnet/wpf/blob/b697746306a695af80a756586cad3f84a62e366c/src/Microsoft.DotNet.Wpf/src/PresentationCore/System/Windows/Media/ArcSegment.cs#L76) - presumably as you describe - explaining why the `EllipseGeometry` and arc based streaming geometry look the same. This is somewhat upsetting, but oh well.

Jonarw: Added an example for this in RenderingCapabilities. I don't think we need to worry too much about this.
image

VisualMelon: For the record, I would prefer we use Arcs if there is no significant performance hit because it's simper, expresses the intention better, and is consistent with the `XAML` renderer; however, I agree that it won't be an issue for the typical use-cases.

Jonarw: I guess you are right. It's also nice that the Arc-method doesn't need the 'magic' constant. I could not detect a significant performance difference in my (admittedly somewhat crude) benchmark.

VisualMelon (approved):
All looks good to me.

PR #1700, opened 2020-10-18
Overhaul ContourSeries.JoinContourSegments()

### LINKED ISSUE #1685: ContourSeries produce fake connections
# Steps to reproduce
For some data, if one used some improper parameters, ContourSeries will produce some fake connections/contours.
The data containing two isolated peaks [attached](https://drive.google.com/file/d/17GJbHztDo0KeYu6HLMI-iC9gBuo_V8SM/view?usp=sharing)  using 

```
            double[] contour_levels = { 0.2, 0.4, 0.6, 0.8, 0.9, 1.0 };
            double maxi = max value in data[, ];
           
            double[] tmp = new double[contourlevels.Length];
            for (int i = 0; i < contourlevels.Length; i++)
            {
                tmp[i] = maxi * contourlevels[i];
            }

           ContourSeries cs = new ContourSeries
            {
                TextColor = OxyColors.Transparent,   
                LabelBackground = OxyColors.Transparent,   
                ContourLevels = tmp,
                Data = data,
            };
```
would produce some artificial connections between the two peaks.

Platform: Windows 10 x64
.NET version: 4.7.2

### Expected behaviour

There should be no fake contours.

### Actual behaviour

Two isolated peaks connected with each other when drawing with ContourSeries.

Jonarw:
Fixes #1685.

### Checklist

- [ ] I have included examples or tests (example already added in #1698)

- Use a new algorithm for `ContourSeries.JoinContourSegments()`. This fixes the 'fake connections' problem and also should improve performance: While the previous algorithm was O(n²), the new one should usually be O(n log(n)).
- Small concern: the new algorithm also results in a different 'start' point of the contours, causing labels to be placed in different locations than before. I guess there is not really a 'correct' position for these labels to be in, so it should be fine?! If there are suggestions how to make this more consistent I'd be happy to take another look.

 @oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Series/ContourSeries.cs
+++ b/Source/OxyPlot/Series/ContourSeries.cs
@@ -462,8 +462,8 @@ namespace OxyPlot.Series
         /// <summary>
         /// Joins the contour segments.
         /// </summary>
-        /// <param name="eps">The tolerance for segment ends to connect (squared distance).</param>
-        private void JoinContourSegments(double eps = 1e-10)
+        /// <param name="epsFactor">The tolerance for segment ends to connect (maximum allowed [length of distance vector] / [length of position vector]).</param>
+        private void JoinContourSegments(double epsFactor = 1e-10)
         {
             this.contours = new List<Contour>();
 
@@ -491,13 +491,16 @@ namespace OxyPlot.Series
                         continue;
                     }
 
+                    var positionVectorLength = Math.Sqrt(Math.Pow(currentPoint.Point.X, 2) + Math.Pow(currentPoint.Point.Y, 2));
+                    var eps = positionVectorLength * epsFactor;
+
                     var maxX = currentPoint.Point.X + eps;
                     var i2 = i + 1;
                     SegmentPoint joinPoint;
 
                     // search for a point with the same coordinates (within eps) as the current point
                     // as points are sorted by X, we typically only need to check the point immediately following the current point
-                    do
+                    while (true)
                     {
                         if (i2 >= points.Count)
                         {
@@ -517,8 +520,13 @@ namespace OxyPlot.Series
                             joinPoint = null;
                             break;
                         }
+
+                        var distance = Math.Sqrt(Math.Pow(joinPoint.Point.X - currentPoint.Point.X, 2) + Math.Pow(joinPoint.Point.Y - currentPoint.Point.Y, 2));
+                        if (distance < eps)
+                        {
+                            break;
+                        }
                     }
-                    while (Math.Abs(joinPoint.Point.Y - currentPoint.Point.Y) > eps);
 
                     // join the two points together
                     if (joinPoint != null)
--- a/Source/OxyPlot/Series/ContourSeries.cs
+++ b/Source/OxyPlot/Series/ContourSeries.cs
@@ -492,20 +492,20 @@ namespace OxyPlot.Series
                     }
 
                     var maxX = currentPoint.Point.X + eps;
-                    var i2 = 1;
+                    var i2 = i + 1;
                     SegmentPoint joinPoint;
 
                     // search for a point with the same coordinates (within eps) as the current point
                     // as points are sorted by X, we typically only need to check the point immediately following the current point
                     do
                     {
-                        if (i + i2 >= points.Count)
+                        if (i2 >= points.Count)
                         {
                             joinPoint = null;
                             break;
                         }
 
-                        joinPoint = points[i + i2];
+                        joinPoint = points[i2];
                         i2++;
                         if (joinPoint.Join != null)
                         {
@@ -537,22 +537,21 @@ namespace OxyPlot.Series
                     }
 
                     var currentPoint = segmentPoint;
-                    var dataPoints = new List<DataPoint>();
 
-                    if (currentPoint.Join == null) // we are the first point of the contour
+                    // search for the beginning of the contour (or use the entry point if the contour is closed)
+                    while (currentPoint.Join != null)
                     {
-                        dataPoints.Add(currentPoint.Point);
-                        currentPoint = currentPoint.Partner;
-                        dataPoints.Add(currentPoint.Point);
-                    }
-                    else // we are somewhere in the middle of the contour, our partner might be the first, so add it first
-                    {
-                        dataPoints.Add(currentPoint.Partner.Point);
-                        dataPoints.Add(currentPoint.Point);
+                        currentPoint = currentPoint.Join.Partner;
+                        if (currentPoint == segmentPoint)
+                        {
+                            break;
+                        }
                     }
 
+                    var dataPoints = new List<DataPoint> { currentPoint.Point, currentPoint.Partner.Point };
                     currentPoint.Processed = true;
-                    currentPoint.Partner.Processed = true;
+                    currentPoint = currentPoint.Partner;
+                    currentPoint.Processed = true;
 
                     // follow the chain of joined points and add their coordinates until we find the last point of the contour (or complete a rotation)
                     while (currentPoint.Join != null)
--- a/Source/OxyPlot/Series/ContourSeries.cs
+++ b/Source/OxyPlot/Series/ContourSeries.cs
@@ -398,33 +398,6 @@ namespace OxyPlot.Series
         }
 
         /// <summary>
-        /// Determines if two values are close.
-        /// </summary>
-        /// <param name="x1">The first value.</param>
-        /// <param name="x2">The second value.</param>
-        /// <param name="eps">The squared tolerance.</param>
-        /// <returns>True if the values are close.</returns>
-        private static bool AreClose(double x1, double x2, double eps = 1e-6)
-        {
-            double dx = x1 - x2;
-            return dx * dx < eps;
-        }
-
-        /// <summary>
-        /// Determines if two points are close.
-        /// </summary>
-        /// <param name="p0">The first point.</param>
-        /// <param name="p1">The second point.</param>
-        /// <param name="eps">The squared tolerance.</param>
-        /// <returns>True if the points are close.</returns>
-        private static bool AreClose(DataPoint p0, DataPoint p1, double eps = 1e-6)
-        {
-            double dx = p0.X - p1.X;
-            double dy = p0.Y - p1.Y;
-            return (dx * dx) + (dy * dy) < eps;
-        }
-
-        /// <summary>
         /// Gets the index of item that is closest to the specified value.
         /// </summary>
         /// <param name="values">A list of values.</param>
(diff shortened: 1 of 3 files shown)


Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -486,100 +459,118 @@ private void AddContourLabels(Contour contour, ScreenPoint[] pts, ICollection<Co
...
+                    if (currentPoint.Join != null)
+                    {
+                        continue;
+                    }
 
-                    // Find a segment that is connected to the head of the contour
-                    segment1 = this.FindConnectedSegment(contourPoints[0], firstSegment.ContourLevel, eps, out reverse);
-                    if (segment1 != null)
+                    var maxX = currentPoint.Point.X + eps;
+                    var i2 = 1;
```
VisualMelon, line 495: I'd prefer that `i2 = i + 1` rather than being an offset (simplifies everything a little).

Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -486,100 +459,118 @@ private void AddContourLabels(Contour contour, ScreenPoint[] pts, ICollection<Co
...
+                p2.Partner = p1;
+                yield return p1;
+                yield return p2;
+            }
 
-                if (firstSegment != null)
+            foreach (var group in this.segments.GroupBy(p => p.ContourLevel))
+            {
+                var level = group.Key;
+                var points = group.SelectMany(GetPoints).OrderBy(p => p.Point.X).ToList();
```
VisualMelon, line 483: Should this be sorting by `x` and then `y`?

Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -486,100 +459,118 @@ private void AddContourLabels(Contour contour, ScreenPoint[] pts, ICollection<Co
...
+                            continue;
+                        }
+
+                        if (joinPoint.Point.X > maxX)
+                        {
+                            joinPoint = null;
+                            break;
+                        }
                     }
+                    while (Math.Abs(joinPoint.Point.Y - currentPoint.Point.Y) > eps);
```
VisualMelon, line 521: Any reason this isn't in the `if` at the top of the loop? I'm not sure I get why this is `>` rather than `<`.

Source/OxyPlot/Series/ContourSeries.cs
VisualMelon: Unless the values are expected to be identical, sorting by coordinates will never be perfect. I'd use a quad-tree for this, though that might be overkill.

Source/OxyPlot/Series/ContourSeries.cs
```diff
@@ -486,100 +459,118 @@ private void AddContourLabels(Contour contour, ScreenPoint[] pts, ICollection<Co
...
+                    var currentPoint = segmentPoint;
+                    var dataPoints = new List<DataPoint>();
+
+                    if (currentPoint.Join == null) // we are the first point of the contour
+                    {
+                        dataPoints.Add(currentPoint.Point);
+                        currentPoint = currentPoint.Partner;
+                        dataPoints.Add(currentPoint.Point);
+                    }
+                    else // we are somewhere in the middle of the contour, our partner might be the first, so add it first
```
VisualMelon, line 548: Not sure I understand this bit

VisualMelon (commented):
In principle I'm fully in favour of this change and it seems to work find in all the examples, but I'm not sure I understand the algorithm at the moment.

Source/OxyPlot/Series/ContourSeries.cs
Jonarw: Yes, sorting won't be perfect, but good enough for 'non-pathological' use cases - under normal circumstances sorting by X should pair corresponding points almost completely.

Source/OxyPlot/Series/ContourSeries.cs
Jonarw: In theory sorting by `y` additionally may help, but in practice I don't think it makes a difference; I wouldn't expect there to be any two identical `x` values, so there wouldn't be any ties to be resolved by additionally sorting by `y`.

Source/OxyPlot/Series/ContourSeries.cs
Jonarw: Well we go through the next couple of points after `currentPoint` _until_ we find one whose Y coordinate is within `eps` of the `currentPoint`s Y coordinate. That's what the `while` condition is checking for. Sure, this could be pulled into an `if` in the loop body, but I found it more concise that way.

VisualMelon: Ah, yeah, that makes sense.

Jonarw: What may be confusing is the fact that usually this loop will only run exactly once, because in most cases the point immediately following `currentPoint` will be the matching one (because they are already sorted by X).

Source/OxyPlot/Series/ContourSeries.cs
VisualMelon: Now that I understand the while properly, this makes more sense too.

Source/OxyPlot/Series/ContourSeries.cs
VisualMelon: I'd just completely missed that the work was done outside the loop. It's a bit confusing with all the exit conditions, but I can't think how to make it much clearer.

Source/OxyPlot/Series/ContourSeries.cs
Jonarw: Here we have two possibilities: 
1. The `currentPoint` is _not_ joined to any other point -> This will be the case for e.g. the first point of an open contour. In this case `currentPoint` will have to be the start point of our contour.

2. The `currentPoint` _is_ joined to another point -> This means that:
2.1 **Either** the contour is closed, so it doesn't matter if we add the `currentPoint` or its `Partner` first
2.2 **Or** the contour is not closed. In that case the `currentPoint`'s `Partner` _may_ be the start point of the contour -> in this case we would have to add the `currentPoint`'s `Partner` first

So in the second case, we just always add the `currentPoint`'s `Partner` first to be safe.

I do realize that this sounds really convoluted.. in my mind this made perfect sense earlier. I'll take another look tomorrow with fresh eyes and see if I can simplify this a bit. Or maybe the naming just isn't right, but I couldn't think of better alternatives to `Partner` and `Join`...

VisualMelon: `Partner` and `Join` are OK in my opinion. What I'm missing is the reason why it can't be the case that both `currentPoint` and `currentPoint.Partner` could be 'in the middle' of an open contour. The partner _might_ be first, but is that necessarily the case?

VisualMelon: I flipped `Peaks` around, so that there is an 'open' contour which ends with X large, and you can see the effect of this when you plot the start/end of each contour:

image

I think the best thing to do is to follow the contour to the end: if `currentPoint` gets back to `segmentPoint`, then you know it is a closed loop and, and can start anywhere; if you get to an end (i.e. a point that isn't joined to another point) then you start from the end.

Jonarw: Oh yes, of course you are right, I'll have to fix that.

Jonarw: Should be fixed now. I also slightly changed the `peaks` function used in the example to have all kinds of open contours included in the example.

VisualMelon: Oh no! You can't change peaks! It's the logo!

VisualMelon: New implementation looks good to me, and the issues seems to be fixed. I don't know if we do care about changing `peaks`, but the new one also doesn't cover the case of the algorithm finding the middle of an open contour initially: the open loops in the new example all 'start' at a low X value.

VisualMelon:
I have a slight concern about the choice of epsilon value; or rather, the use of a single epsilon value. Normally one would want an epsilon to be relative to some measure, but I'm not sure what that measure would be here. My only suggestion would be that the epsilon for a given X and Y value should be proportional to the change in X and Y values in the contour series itself. E.g. if you feed in a contour with X values [1, 2, 10, 11, 28, 30], then a point around 11 should have an epsilon proportional to min(11-10, 28-11).

While the new code is much better than the old stuff, I'm not sure what it is has changed that explains why the old code would broken, and whether the new code will work with e.g. very small or very large coordinates (and indeed, a combination of very small and very large deltas as you'd expect on a logarithmic axis).

Source/OxyPlot/Series/ContourSeries.cs
Jonarw: Haha, I'll try to come up with a function that will also cover contours which have their lowest X value somewhere in the middle and make that a separate example, so we can keep the original `peaks` :)

Jonarw:
I guess you have a point with the epsilon value... I'll try and come up with something, you suggestions does sound reasonable at least in theory.

Jonarw:
I had a go at this. Unfortunately I couldn't find a reasonable way to make your suggestion work... At the time I would need the epsilon I just don't know which is the next point of the contour. The next point in X order could be a completely unrelated point.

So I guess the next-best thing is to make the epsilon at least proportional to the point's coordinates. I think it makes sense to use the distance here as opposed to applying different epsilons to the two individual coordinates of a point.
My latest commit demonstrates this.

I am not 100% convinced that this is the best way and open to further suggestions, but I think it is at least reasonable.

VisualMelon:
Would there be any sense is making it proportional to the segment length? I have other things to do today, but I'll try to take a proper look again tomorrow.

VisualMelon:
Sorry, I completely missed that you'd already make them proportional to segment length (misunderstood your middle paragraph). I certainly can't think of any better way of doing this. In terms of the PR, it might be a good idea to add some more examples with log axes and large/small values, so that have some empirical evidence that this works as expected, but otherwise it all looks good to me.

Jonarw:
Well actually it is not proportional to the segment length, so you understood it correctly the first time ;).
As you can see [here](https://github.com/oxyplot/oxyplot/pull/1700/files#diff-07d9393d9381e2a893c137c6c1631c9667347843e6e5213e2cb2744202ed0e5dR495), the epsilon is proportional to the distance of the current point to (0,0). I think that makes sense, because in the end the epsilon is there to deal with floating-point noise, which I assume is also proportional to the absolute value of the number in many cases.

Making epsilon proportional to the segment length also sounds sensible, but I couldn't tell if it makes more sense. For most applications it probably won't matter.

I'll see if I can come up with some more examples.

VisualMelon:
Sorry, thought I'd already comment on this. The new example looked good to me, and I'm happy for this to be merged if you are.

Issue #24, opened 2014-08-14
Minimum number of labels on an axis


objorke:
**Imported from [http://oxyplot.codeplex.com/workitem/9975](http://oxyplot.codeplex.com/workitem/9975)**

**[objo](http://www.codeplex.com/site/users/view/objo)** 2012-08-09 00:17
There should never be less than two labels on the axes. 

See discussion <a href="http://oxyplot.codeplex.com/discussions/390008" rel="nofollow">
http://oxyplot.codeplex.com/discussions/390008</a>

Trass3r:
Does any-one have an idea how to implement this?

objorke:
First I think we need a couple of examples that reproduces the bug!
Your screen shots in #749 looks good, can you provide example code? I guess the height of the plot view is important - can this be reproduced in the example browser with the default window size?

Trass3r:
Yes depends on the plot height. Visible in the BindingObservableCollection example when clicking on load points with:

```patch
 .../MainWindow.xaml                                | 24 ++++++++++++++++++++--
 1 file changed, 22 insertions(+), 2 deletions(-)

diff --git a/Source/Examples/WPF/WpfExamples/Examples/BindingObservableCollectionDemo/MainWindow.xaml b/Source/Examples/WPF/WpfExamples/Examples/BindingObservableCollectionDemo/MainWindow.xaml
index 5cef092..10b96b7 100644
--- a/Source/Examples/WPF/WpfExamples/Examples/BindingObservableCollectionDemo/MainWindow.xaml
+++ b/Source/Examples/WPF/WpfExamples/Examples/BindingObservableCollectionDemo/MainWindow.xaml
@@ -1,11 +1,12 @@
 ﻿<Window x:Class="BindingObservableCollectionDemo.Window" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:oxy="clr-namespace:OxyPlot.Wpf;assembly=OxyPlot.Wpf"
         Title="BindingDemo" Width="640"
-        Height="480">
+        Height="380">
   <Grid>
     <Grid.RowDefinitions>
       <RowDefinition Height="Auto" />
       <RowDefinition Height="*" />
+      <RowDefinition Height="*" />
     </Grid.RowDefinitions>
     <Button Command="{Binding LoadCommand}" Content="Load next points" />
[...]

PR #1712, opened 2020-11-18
Add MinimumMajorIntervalCount and MaximumMajorIntervalCount Axis Properties (#24)

VisualMelon:
Addresses #24.

### Checklist

- Add `MinimumMajorIntervalCount` and `MaximumMajorIntervalCount` to `Axis` with defaults `2` and `double.MaxValue`
- Add `minIntervalCount` and `maxIntervalCount` parameters to the 2 `CalculateActualInterval` method overloads, and implement appropriate logic in all implementions
    - `maxIntervalCount` as an upper-bound on what is otherwise computed as the `range/intervalSize`
    - `minIntervalCount` is a lower bound on the projected interval count: has precedence over `maxIntervalCount` on edge cases (so in the `Min = Max` example, the thing takes the smallest interval that satisfies the minimum.
- Add 3 examples

### Concerns

 - This design hasn't been agreed, but I figured it would be easier for everyone if I just implemented it rather than trying to describe it as a proposal: any and all feedback welcome
 - The signature changes make this a breaking change: any code that overrides the methods will no longer compile.
 - What should the defaults be? I made the min 2 so that I could look through the example browser and look for weirdness, but this changes existing behaviour, so 'by default' we should make it `0` unless we can convinced ourselves the alternative is good:
    - I think `2` gives people a better experience by default: I'd rather my plots were untidy than uninterpretable
 - I've not fully reviewed the interaction with other related properties, but nothing seems obviously broken.
 - The maximum is not a hard maximum: it will be violated if there are no 'nice' values between it and the minimum.

### Other

 - Apart from being a breaking change, I don't see any issue with adding this if we make the defaults `0` and `double.MaxValues`: the behaviour will be unchanged.
 - Since #24 is considered a bug, this would be a candidate for 2.1 if it were not for the breaking change: could we remove the params in the signatures and just grab the properties manually for 2.1, and maybe do otherwise for 3?

@oxyplot/admins


Jonarw:
Sounds interesting! I'm a bit busy these days, but I'll try to take a closer look tomorrow.

Jonarw:
I think I like this change!
While #24 is labeled as a bug, I'd be inclined to consider this more as an enhancement rather than a bug fix, so I don't think we'd need to backport this to v2.1. Especially if we are considering making setting `MinimumMajorIntervalCount=2` by default (which I think would be a sensible idea).

Some other thoughts:
- Could `MaximumMajorIntervalCount` get another name that conveys its 'not-hard-maximum' nature? Something like `DesiredMaximumIntervalCount`, though that is quite a mouthful...
- One issue I found while playing around: With very short axes, something like this can happen: 
(Axis Examples > Tick Style = Outside, `MinimumMajorIntervalCount=2`, zoomed in a bunch):
image
Mind you, this is still better than `MinimumMajorIntervalCount=0`, which shows just the tick 100. But I think it would be more sensible to display only major ticks 50 and 100 here. On some zoom/pan levels this is the case, but for others we get more ticks than necessary I think (like in this example).

VisualMelon:
I like the symmetry in the names at the moment, but prefixing `Desired` would probably be a clearer.

I can't repro the small-axis issue: can you provide the exact code you were using? I agree what you describe would be better and is what I see when I set the min and leave the max free:

image

It would be nice to add this as a non-breaking change now and change the defaults later, but I in terms of wanting to get 2.1 out the door, I think you are right and we should hold off adding more complexity.

Jonarw:
> I can't repro the small-axis issue: can you provide the exact code you were using?

For me it happens every time when I
- open the example mentioned above
- drag the window small (the exact size doesn't matter)
- zoom in once with the mouse wheel
- pan the axis such that both 50 and 100 should be in range

Let me know if you can reproduce it this way, if not I'll try and come up with some code that reproduces it.

VisualMelon:
Ah, yeah, I remember this now (can repro also). I thought I'd addressed, but I can't think how, so I guess I didn't.

It's caused by the range changing slightly when panning: if the range is 100, then when it considers an interval of 50 it projects the number of ticks as being 100/50=2, and it is happy with that. But when you pan, the range can end up slightly either side of 100, and if it goes below 100, it rejects an interval of 50 (because for a range of <100, it is possible there could be only 1 tick depending on the offset).

This is an unsolvable problem in the sense that there will always be a range where it could go either way. We could ask `GetTickValues` to determine the actual number of ticks, but that make the behaviour more 'bouncy', as the number of ticks depends heavily on the offset.

VisualMelon:
I'm tempted to say we should merge this. The range issue is still there, but even the new behaviour is arguably better than the old behaviour (a single tick is completely useless), and we may have a better idea about how to handle edge cases if we see it working in reality.

Jonarw:
Agree this can be merged. Even though this one particular point could be improved in the future, this is definitely an improvement to the current behavior.

I'll not merge right away, I don't really have an overview over other open PRs this might interfere with. I'll try and catch up with the PRs in the coming days, but feel free to merge yourself!

Jonarw (approved):

PR #1736, opened 2021-01-11
Add control over how far from the series the tracker fires

### LINKED ISSUE #1731: Consider adding control over how far from the series the tracker fires
Hello!

### Consider adding control over how far from the series the tracker fires

It is hardcoded at the moment, we can see it here

https://github.com/oxyplot/oxyplot/blob/924cc3676f74e7902868c03a953135610fdf6cd2/Source/OxyPlot/PlotController/Manipulators/TrackerManipulator.cs#L144

or

https://github.com/oxyplot/oxyplot/blob/924cc3676f74e7902868c03a953135610fdf6cd2/Source/OxyPlot/PlotController/Manipulators/TouchTrackerManipulator.cs#L156

Also, I think this also applies to this

https://github.com/oxyplot/oxyplot/blob/924cc3676f74e7902868c03a953135610fdf6cd2/Source/OxyPlot/PlotController/Manipulators/TrackerManipulator.cs#L79

or

https://github.com/oxyplot/oxyplot/blob/924cc3676f74e7902868c03a953135610fdf6cd2/Source/OxyPlot/PlotController/Manipulators/TouchTrackerManipulator.cs#L102

rstm-sf:
Closes #1731 

### Checklist

- Add control over how far from the series the tracker fires
- Add checking distance for result between data points (breaking change?)

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 - Example for Issue #1716 showing poor tick spacing on DateTimeAxis with interval types of Weeks or Years
 - Example for label placement on BarSeries with non-zero BaseValue (#1726)
 - Add control over how far from the series the tracker fires (#1736)
+- Add option to check distance for result between data points (#1736)
 
 ### Changed
 
@@ -15,7 +16,6 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Zero-crossing axis bounds (#1708)
 - Incorrect label placement on BarSeries with non-zero BaseValue (#1726)
-- Add checking distance for result between data points (#1736)
 
 ## [2.1.0-Preview1] - 2020-10-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Example for Issue #1716 showing poor tick spacing on DateTimeAxis with interval types of Weeks or Years
 - Example for label placement on BarSeries with non-zero BaseValue (#1726)
+- Add control over how far from the series the tracker fires (#1736)
 
 ### Changed
 
@@ -14,6 +15,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Zero-crossing axis bounds (#1708)
 - Incorrect label placement on BarSeries with non-zero BaseValue (#1726)
+- Add checking distance for result between data points (#1736)
 
 ## [2.1.0-Preview1] - 2020-10-18
 
--- a/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
+++ b/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
@@ -67,5 +67,38 @@ namespace ExampleLibrary
             };
             return model;
         }
+
+        [Example("Specified distance of the tracker fires")]
+        public static Example TrackerFiresDistance()
+        {
+            var model = new PlotModel
+            {
+                Title = "Specified distance of the tracker fires",
+                Subtitle = "Press the left mouse button to test the tracker.",
+            };
+            model.Series.Add(new FunctionSeries(Math.Sin, 0, 10, 100));
+            
+            // create a new plot controller with default bindings
+            var plotController = new PlotController();
+
+            // remove a tracker command to the left mouse down event by default
+            plotController.Unbind(PlotCommands.SnapTrack);
+            
+            // add a tracker command to the left mouse down event with specified distance
+            plotController.BindMouseDown(
+                OxyMouseButton.Left,
+                new DelegatePlotCommand<OxyMouseDownEventArgs>((view, controller, args) =>
+                    controller.AddMouseManipulator(
+                        view,
+                        new TrackerManipulator(view)
+                        {
+                            Snap = true,
+                            PointsOnly = false,
+                            FiresDistance = 2.0,
+                        },
+                        args)));
+
+            return new Example(model, plotController);
+        }
     }
 }
--- a/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
+++ b/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
@@ -96,7 +96,7 @@ namespace ExampleLibrary
                             Snap = true,
                             PointsOnly = false,
                             FiresDistance = 2.0,
-                            IsCheckDistanceBetweenPoints = true,
+                            CheckDistanceBetweenPoints = true,
                         },
                         args)));
             plotController.BindTouchDown(
--- a/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
+++ b/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
@@ -22,7 +22,6 @@ namespace ExampleLibrary
             {
                 Title = "No tracker interpolation",
                 Subtitle = "Used for discrete values or scatter plots.",
-                TrackerFiresDistance = 4.0,
             };
             var l = new Legend
             {
@@ -32,16 +31,16 @@ namespace ExampleLibrary
             model.Legends.Add(l);
 
             var s1 = new LineSeries
-                         {
-                             Title = "Series 1",
-                             CanTrackerInterpolatePoints = false,
-                             Color = OxyColors.SkyBlue,
-                             MarkerType = MarkerType.Circle,
-                             MarkerSize = 6,
-                             MarkerStroke = OxyColors.White,
-                             MarkerFill = OxyColors.SkyBlue,
-                             MarkerStrokeThickness = 1.5
-                         };
+            {
+                Title = "Series 1",
+                CanTrackerInterpolatePoints = false,
+                Color = OxyColors.SkyBlue,
+                MarkerType = MarkerType.Circle,
+                MarkerSize = 6,
+                MarkerStroke = OxyColors.White,
+                MarkerFill = OxyColors.SkyBlue,
+                MarkerStrokeThickness = 1.5
+            };
             for (int i = 0; i < 63; i++)
             {
                 s1.Points.Add(new DataPoint((int)(Math.Sqrt(i) * Math.Cos(i * 0.1)), (int)(Math.Sqrt(i) * Math.Sin(i * 0.1))));
@@ -59,7 +58,6 @@ namespace ExampleLibrary
             {
                 Title = "Handling the TrackerChanged event",
                 Subtitle = "Press the left mouse button to test the tracker.",
-                TrackerFiresDistance = 4.0,
             };
             model.Series.Add(new FunctionSeries(Math.Sin, 0, 10, 100));
             model.TrackerChanged += (s, e) =>
--- a/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
+++ b/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
@@ -96,6 +96,7 @@ namespace ExampleLibrary
                             Snap = true,
                             PointsOnly = false,
                             FiresDistance = 2.0,
+                            IsCheckDistanceBetweenPoints = true,
                         },
                         args)));
             plotController.BindTouchDown(
--- a/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
+++ b/Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
@@ -81,16 +81,28 @@ namespace ExampleLibrary
             // create a new plot controller with default bindings
             var plotController = new PlotController();
 
-            // remove a tracker command to the left mouse down event by default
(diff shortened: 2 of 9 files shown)


rstm-sf:
Hello!

I do not know where it is better to define it:
- It seems to me impossible in the tracker
- In the series, we will have to determine the distance separately for each

rstm-sf:
> Add checking distance for result between data points (breaking change?)

It seems that this should be done as a separate task. In fact, this is the problem that I didn't like: because of this, the output is much larger than 20

VisualMelon (commented):
I think this should really be `TrackManipulator`, and the user should bind their own view commands for it; however, that's a painful prospect with the library as it stands: you'd have to consult `PlotController.cs` and `PlotCommands.cs` to provide an otherwise-consistent user experience. (At present, `PlotController` is great if you want to change everything, but small changes are no fun)

We could make this easier by adding helper methods to register 'default' commands with a given configuration, e.g. replace https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/PlotController/PlotController.cs#L50-L56 with a call to `static void RegisterDefaultTrackerCommands(double trackerFireDistance, double trackerInterpolateDistance, etc.)` or a version which takes a delegate that generates the command from the usual settings for `LockToInitialSeries`, `Snap`, and `PointsOnly` (so you can configure it however you like), or whatever: the point would be to do all the 'tracker' stuff in one place, so that it's easy to override.

The same can be done for the cursor keys (e.g. with configurable pan distance), the -/+ keys (configurable zoom factor), etc. By committing to using these methods as default, consumers will be able to use them to override existing behaviour easily, and know that there code will be compatible with future versions (or not compile), rather than having to C&P bit chunks of `PlotController.cs` to refit a sub-set of behaviour.

rstm-sf:
@VisualMelon thank you very much for the clarification and guidance on how best to do it! I will try to fix it this week

Source/OxyPlot/PlotController/Manipulators/TrackerManipulator.cs
```diff
@@ -139,23 +144,26 @@ private static TrackerHitResult GetNearestHit(Series.Series series, ScreenPoint
...
+                if (IsAcceptableResult(result, point, firesDistance))
+                {
+                    return result;
+                }
             }
 
             return null;
         }
+
+        private static bool IsAcceptableResult(TrackerHitResult result, ScreenPoint point, double firesDistance) =>
```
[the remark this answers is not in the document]
rstm-sf: I prefer "Is" instead of "Should" :)

rstm-sf:
> We could make this easier by adding helper methods to register 'default' commands with a given configuration, e.g. replace https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot/PlotController/PlotController.cs#L50-L56 with a call to static void RegisterDefaultTrackerCommands(double trackerFireDistance, double trackerInterpolateDistance, etc.) or a version which takes a delegate that generates the command from the usual settings for LockToInitialSeries, Snap, and PointsOnly (so you can configure it however you like), or whatever: the point would be to do all the 'tracker' stuff in one place, so that it's easy to override.
>
> The same can be done for the cursor keys (e.g. with configurable pan distance), the -/+ keys (configurable zoom factor), etc. By committing to using these methods as default, consumers will be able to use them to override existing behaviour easily, and know that there code will be compatible with future versions (or not compile), rather than having to C&P bit chunks of PlotController.cs to refit a sub-set of behaviour.

Perhaps this should be done as a separate issue? I don't feel like adding additional API unnecessarily :)

Source/Examples/ExampleLibrary/Examples/TrackerExamples.cs
```diff
@@ -59,5 +67,38 @@ public static PlotModel TrackerChangedEvent()
...
+            model.Series.Add(new FunctionSeries(Math.Sin, 0, 10, 100));
+            
+            // create a new plot controller with default bindings
+            var plotController = new PlotController();
+
+            // remove a tracker command to the left mouse down event by default
+            plotController.Unbind(PlotCommands.SnapTrack);
+            
+            // add a tracker command to the left mouse down event with specified distance
+            plotController.BindMouseDown(
```
VisualMelon, line 88: Could we replace the binding for `SnapTrackTouch` as well?

Source/OxyPlot/Utilities/TrackerHelper.cs
```diff
@@ -0,0 +1,48 @@
...
+                {
+                    return result;
+                }
+            }
+
+            // Check between data points (if possible)
+            if (!pointsOnly)
+            {
+                var result = series.GetNearestPoint(point, true);
+                if (IsTrackerOpen(result, point, firesDistance))
```
VisualMelon, line 36: Since it is breaking (and it's pretty jarring when you are used to the old behaviour), I think we should make the new behaviour optional, though I can't immediately think of a sensible property/parameter name.

CHANGELOG.md
```diff
@@ -14,6 +15,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Zero-crossing axis bounds (#1708)
 - Incorrect label placement on BarSeries with non-zero BaseValue (#1726)
+- Add checking distance for result between data points (#1736)
```
VisualMelon, line 18: If it is made optional, this will need to be an added feature rather than a fix.

VisualMelon (commented):
Apart from the breaking change - which I think we should change (though I agree the new behaviour makes more sense) - this all looks good.

Source/OxyPlot/Utilities/TrackerHelper.cs
rstm-sf: Perhaps we can make an additional property and mark it as `Obsolete`? I don't know to which version then keep it

But I am more for the fact that this is the wrong behavior

Source/OxyPlot/PlotController/Manipulators/TouchTrackerManipulator.cs
```diff
@@ -50,6 +52,17 @@ public TouchTrackerManipulator(IPlotView plotView)
...
 
+        /// <summary>
+        /// Gets or sets the distance from the series at which the tracker fires.
+        /// </summary>
+        public double FiresDistance { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to check distance when showing tracker between data points.
+        /// </summary>
+        [System.Obsolete("")]
```
rstm-sf, line 63: Which version is better to indicate?

Source/OxyPlot/Utilities/TrackerHelper.cs
VisualMelon: After mulling this over for a while, I think it might be best to provide a property that enables/disables the distance check within `Delta` for _both_ hit tests.
 - It means the beaviour is consistent for interpolated/non-interpolated plots
 - The ability to easily track a series once you've selected it is useful (it's familiar, and useful if you have a messy plot, e.g. I sometimes have to follow a single line out of 200 or so, and letting the tracker do the work for me makes life much better)
 - Users can still use the modifier keys if they want non-snapping or whatever behaviour

I would expect the default to be to disable the distance checks by default, at least for 2.1. I don't think marking the property Obsolete is a good idea, because their behaviour may be desirable.

rstm-sf: Did I understand correctly that it is better to remove the obsolete mark? Good. The only thing left to do is to make sure the property naming and documentation is well interpreted :)

VisualMelon: Yes, for now I think it's probably best to stick with what you have working, checking that the default behaviour is the same as the old behaviour. We can make additional changes in a separate PR. Would it be a problem for you if this wasn't in the - somewhat late (my fault) - 2.1 release? Because it is adding additional behaviour one way or another, we probably shouldn't include it. (It's also a feature that can be easily implemented without changes to the core library, so the pre-release implementation and examples should help anyone trying to do something similar).

rstm-sf: No, that won't be a problem. I figured out how to do a workaround by describing custom class for the manipulator. Thanks for help!

VisualMelon:
Sorry it's taken me so long to get back to this.

Am I right in thinking that the default behaviour is back to the original behaviour by default? If so, I don't think there is anything else to do, and I'll try to merge this over the weekend when I can take another proper look at it.

Source/OxyPlot/Utilities/TrackerHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+        /// </summary>
+        /// <param name="series">The series.</param>
+        /// <param name="point">The point.</param>
+        /// <param name="snap">Snap to points.</param>
+        /// <param name="pointsOnly">Check points only (no interpolation).</param>
+        /// <param name="firesDistance">The distance from the series at which the tracker fires</param>
+        /// <param name="isCheck"></param>
+        /// <returns>A tracker hit result.</returns>
+        /// <remarks>
+        /// TODO: With removing <see cref="TrackerManipulator.IsCheckDistanceBetweenPoints" />
```
rstm-sf, line 16: It seems that this todo is already redundant. I try to clean up by the weekend

rstm-sf:
```cs
if (!isCheck || IsTrackerOpen(result, point, firesDistance))
```

It uses lazy operations

Source/OxyPlot/Utilities/TrackerHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+    {
+        /// <summary>
+        /// Gets the nearest tracker hit.
+        /// </summary>
+        /// <param name="series">The series.</param>
+        /// <param name="point">The point.</param>
+        /// <param name="snap">Snap to points.</param>
+        /// <param name="pointsOnly">Check points only (no interpolation).</param>
+        /// <param name="firesDistance">The distance from the series at which the tracker fires</param>
+        /// <param name="isCheck"></param>
```
VisualMelon, line 13: Brief param comment needed here`: could the variable name be clearer (would `checkDistanceBetweenPoints` be objectionable?)

Source/OxyPlot/Utilities/TrackerHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    return result;
+                }
+            }
+
+            // Check between data points (if possible)
+            if (!pointsOnly)
+            {
+                var result = series.GetNearestPoint(point, true);
+                if (!isCheck || IsTrackerOpen(result, point, firesDistance))
```
VisualMelon, line 41: (Note: this is the only line that is different from the origin in `TrackManipulator.cs`)

Source/OxyPlot/Utilities/TrackerHelper.cs
VisualMelon: Should mention that it is ignored if `pointsOnly` is set.

Source/OxyPlot/Utilities/TrackerHelper.cs
```diff
@@ -0,0 +1,53 @@
...
+                {
+                    return result;
+                }
+            }
+
+            return null;
+        }
+
+        private static bool IsTrackerOpen(TrackerHitResult result, ScreenPoint point, double firesDistance) =>
```
VisualMelon, line 50: I'm still not wild about this name, because it feels like it is checking state: could it be `shouldTrackerOpen`? (private, so not a big deal)

Source/OxyPlot/PlotController/Manipulators/TrackerManipulator.cs
```diff
@@ -47,6 +49,16 @@ public TrackerManipulator(IPlotView plotView)
...
 
+        /// <summary>
+        /// Gets or sets the distance from the series at which the tracker fires.
+        /// </summary>
+        public double FiresDistance { get; set; }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether to check distance when showing tracker between data points.
+        /// </summary>
+        public bool IsCheckDistanceBetweenPoints { get; set; }
```
VisualMelon, line 60: Again, this name is a bit weird (and it's public). I know you don't like `Should`, but it sounds wrong to me with `Is`. Can I suggest `CheckDistanceBetweenPoints`? There hasn't been any strict adherence to prefixing boolean properties previously.

Source/OxyPlot/Utilities/TrackerHelper.cs
rstm-sf: Well, maybe I'm wrong and it doesn't read natively as I think. It's just that I'm used to checking the state with the prefix `Is` and only use `Should` in tests

VisualMelon: I see what you mean. Ah well, it's private. I see you've already changed it, but if you're happy, I'm happy.

VisualMelon:
@rstm-sf all looks good to me. I'll give the other maintainers a chance to look it over, and if there are no other issues I'll get this merged tomorrow.

rstm-sf:
Ok, Thanks!

PR #1746, opened 2021-02-19
ExtrapolationLineSeries

### LINKED ISSUE #1740: Line Series with style for extrapolated/intropolated regions.
I would suggest a new derived line series to indicate that parts of a curve have been extrapolated/intrapolated. This would be similar to the ThreeColorLine series but the visualization would depend on the horizontal coordinate and not the vertical one. 

I have added an [implementation](https://github.com/Doeharrrck/oxyplot/blob/develop/Source/OxyPlot/Series/ExtrapolationLineSeries.cs) and [examples](https://github.com/Doeharrrck/oxyplot/blob/develop/Source/Examples/ExampleLibrary/Series/ExtrapolationLineSeriesExamples.cs) to my fork of the project.

The examples look like this:
LineFit
Interpolation

Compared to a normal LineSeries, the class ExtrapolationLineSeries has three additional properties:
- Intervals: Gets or sets an IEnumerable<Tuple<double, double>> in which the curve is displayed in extra/interpolated style. Can be set to null. If not null, the intervals must be ascending and may not overlap. Otherwise an InvalidOperationException is thrown. Positive and negative infinity are also legal delimiters. 

- ExtrapolationColor: Gets or sets the color of the intervals in which the curve is visualized as extra/interpolated. In addition the line style is assigned only to these intervals while the rest of the line is in solid style.

- IgnoreExtraplotationForScaling: If set to true, the extrapolated intervals are not taken into account when the minima and maxima of the dataset are calculated. Hence, the extrapolated line will also not scale the axes.

Is it possible to contribute my work?

Doeharrrck:
Fixes #1740 .

### Checklist

- [ x] I have included examples or tests
- [ x] I have updated the change log
- [ x] I am listed in the CONTRIBUTORS file
- [ ?] I have cleaned up the commit history (use rebase and squash)

- I have added a series described in issue #1740 and also added several examples.
- The new classes are ExtrapolationLineSeries.cs, ExtrapolationLineSeriesExamples.cs and the support class DataRange.cs plus unit tests.
- I have not modified any existing code except for fixing a comment in RectangleItem.cs.
- I am not sure whether I correctly cleaned up the history because i am new to git.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,21 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- Example for Issue #1716 showing poor tick spacing on DateTimeAxis with interval types of Weeks or Years
+- Example for label placement on BarSeries with non-zero BaseValue (#1726)
+- ExtrapolationLineSeries as described in #1740 to allow to display pre-defined intervals of a line series with a different style than the rest.
+- Added DataRange, unit tests and examples as support for ExtrapolationLineSeries.
+- Add control over how far from the series the tracker fires (#1736)
+- Add option to check distance for result between data points (#1736)
 
 ### Changed
+- Updated Series.cd with ExtrapolationLineSeries and removed classes that do not exist anymore
 
 ### Removed
 
 ### Fixed
 - Zero-crossing axis bounds (#1708)
+- Incorrect label placement on BarSeries with non-zero BaseValue (#1726)
 
 ## [2.1.0-Preview1] - 2020-10-18
 
@@ -88,6 +96,7 @@ All notable changes to this project will be documented in this file.
 - Optimize clipping calls (#1661)
 - Mark CandleStickAndVolumeSeries as obsolete (#1661)
 - Implement StreamGeometry-based implementations of DrawEllipses, DrawLine, DrawLineSegments and DrawRectangle(s) which improves the rendering speed on WPF (#1673)
+- Change algorithm of ContourSeries.JoinContourSegments(). This should improve performance in most cases, but will cause labels to appear in different spots than before (#1685)
 
 ### Removed
 - Remove PlotModel.Legends (#644)
@@ -132,6 +141,7 @@ All notable changes to this project will be documented in this file.
 - ScreenMin and ScreenMax on Horizontal and Vertical Axes depends on plot bounds (#1652)
 - Windows Forms clipping last line of measured text (#1659)
 - Inconsistent Zooming behaviour (#1648)
+- ContourSeries produce fake connections (#1685)
 
 ## [2.0.0] - 2019-10-19
 ### Added 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Added
+
+### Changed
+
+### Removed
+
+### Fixed
+- Zero-crossing axis bounds (#1708)
+
 ## [2.1.0-Preview1] - 2020-10-18
 
 ### Added
--- a/Source/Examples/ExampleLibrary/Series/ExtrapolationLineSeriesExamples.cs
+++ b/Source/Examples/ExampleLibrary/Series/ExtrapolationLineSeriesExamples.cs
@@ -78,7 +78,7 @@ namespace ExampleLibrary.Series
             lineSeries.Points.AddRange(fitPoints);
             model.Series.Add(lineSeries);
 
-            Legend legend = new Legend
+            var legend = new Legend
             {
                 LegendPosition = LegendPosition.BottomRight,
             };
@@ -127,7 +127,7 @@ namespace ExampleLibrary.Series
 
             model.Series.Add(lineSeries);
 
-            Legend legend = new Legend
+            var legend = new Legend
             {
                 LegendPosition = LegendPosition.TopCenter,
             };
@@ -192,7 +192,7 @@ namespace ExampleLibrary.Series
 
             model.Series.Add(lineSeries);
 
-            Legend legend = new Legend
+            var legend = new Legend
             {
                 LegendPosition = LegendPosition.TopCenter,
             };
@@ -234,7 +234,7 @@ namespace ExampleLibrary.Series
 
             model.Series.Add(lineSeries);
 
-            Legend legend = new Legend
+            var legend = new Legend
             {
                 LegendPosition = LegendPosition.TopCenter,
             };
@@ -269,7 +269,8 @@ namespace ExampleLibrary.Series
                 .Range(0, 98)
                 .Select(x => new DataRange(1000 * x, (1000 * x) + 500));
 
-            lineSeries.Intervals.AddRange(intervals);
+            foreach(var interval in intervals)
+                lineSeries.Intervals.Add(interval);
 
             lineSeries.Intervals.Add(new DataRange(200_000, double.PositiveInfinity));
 
@@ -281,7 +282,7 @@ namespace ExampleLibrary.Series
 
             model.Series.Add(lineSeries);
 
-            Legend legend = new Legend
+            var legend = new Legend
             {
                 LegendPosition = LegendPosition.TopCenter,
             };
@@ -335,7 +336,7 @@ namespace ExampleLibrary.Series
 
             model.Series.Add(lineSeries);
 
-            Legend legend = new Legend
+            var legend = new Legend
             {
                 LegendPosition = LegendPosition.TopCenter,
             };
@@ -360,11 +361,11 @@ namespace ExampleLibrary.Series
                 throw new ArgumentException("at least two points required", nameof(points));
             }
 
-            double meanX = points.Select(p => p.X).Average();
-            double meanY = points.Select(p => p.Y).Average();
+            var meanX = points.Select(p => p.X).Average();
+            var meanY = points.Select(p => p.Y).Average();
 
-            double cov = Covariance(points, meanX, meanY);
-            double var2_x = Variance2(points.Select(p => p.X));
+            var cov = Covariance(points, meanX, meanY);
+            var var2_x = Variance2(points.Select(p => p.X));
 
             slope = cov / var2_x;
             intercept = meanY - (slope * meanX);
@@ -375,7 +376,7 @@ namespace ExampleLibrary.Series
         /// </summary>
         private static double Covariance(IEnumerable<ScatterPoint> points, double meanX, double meanY)
         {
-            double res = points.Sum(p => p.X * p.Y);
+            var res = points.Sum(p => p.X * p.Y);
 
             res -= points.Count() * meanX * meanY;
             res /= points.Count() - 1;
@@ -388,9 +389,9 @@ namespace ExampleLibrary.Series
         /// </summary>
         private static double Variance2(IEnumerable<double> values)
         {
-            double mean = values.Average();
+            var mean = values.Average();
 
-            double res = values.Sum(x => x * x);
+            var res = values.Sum(x => x * x);
 
             res -= values.Count() * mean * mean;
             res /= values.Count() - 1;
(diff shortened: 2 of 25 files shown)


CHANGELOG.md
```diff
@@ -6,8 +6,12 @@ All notable changes to this project will be documented in this file.
 ### Added
 - Example for Issue #1716 showing poor tick spacing on DateTimeAxis with interval types of Weeks or Years
 - Example for label placement on BarSeries with non-zero BaseValue (#1726)
+- ExtrapolationLineSeries and examples
```
VisualMelon, line 9: Should mention motivating issue (helps future maintainers find design discussion):

    - ExtrapolationLineSeries and examples (#1740)

Source/OxyPlot/Series/DataRange.cs
```diff
@@ -0,0 +1,111 @@
...
+            return CodeGenerator.FormatConstructor(this.GetType(), "{0},{1}", this.Minimum, this.Maximum);
+        }
+
+        /// <summary>
+        /// Returns a <see cref="string" /> that represents this instance.
+        /// </summary>
+        /// <returns>A <see cref="string" /> that represents this instance.</returns>
+        public override string ToString()
+        {
+            return $"{this.Minimum} {this.Maximum}";
```
VisualMelon, line 108: Would prefer something like$"[{min}, {max}]", so that it's clearly a closed interval (and distinguishes it from e.g. `ScreenPoint.ToString()`.

Source/OxyPlot/Series/DataRange.cs
```diff
@@ -0,0 +1,111 @@
...
+        /// </summary>
+        public static readonly DataRange Undefined = new DataRange(double.NaN, double.NaN);
+
+        private double? minimum;
+        private double? maximum;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DataRange"/> struct.
+        /// </summary>
+        /// <param name="min">lower bound.</param>
```
VisualMelon, line 30: Should use proper sentence case in param comments, e.g. "The inclusive lower bound."

Source/OxyPlot/Series/DataRange.cs
```diff
@@ -0,0 +1,111 @@
...
+    /// </summary>
+    public struct DataRange : ICodeGenerating
+    {
+        /// <summary>
+        /// The undefined data range.
+        /// </summary>
+        public static readonly DataRange Undefined = new DataRange(double.NaN, double.NaN);
+
+        private double? minimum;
+        private double? maximum;
```
VisualMelon, line 25: - Should be `readonly`.
 - I'd prefer these were non-nullable, and there was an explicit `isDefined` field, and suitable validation was performed in the constructor, but it's not a big deal

Source/OxyPlot/Series/DataRange.cs
```diff
@@ -0,0 +1,111 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DataRange"/> struct.
+        /// </summary>
+        /// <param name="min">lower bound.</param>
+        /// <param name="max">upper bound.</param>
+        public DataRange(double min, double max)
+        {
+            if (max < min)
```
VisualMelon, line 34: I'd prefer that we explicitly reject NaN values here (so that it is impossible to accidently construct an undefined interval), but I'd defer to @objorke for that decision. The rest of the implementation looks fine wrt. to the current behaviour of accepting NaNs.

Source/OxyPlot/Series/DataRange.cs
```diff
@@ -0,0 +1,111 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="DataRange.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Represents an interval defined by two doubles
```
VisualMelon, line 6: //   Represents an interval defined by two doubles.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
```diff
@@ -0,0 +1,322 @@
...
+        public List<DataRange> Intervals { get; } = new List<DataRange>();
+
+        /// <summary>
+        /// Gets the actual dash array for the extrapolated parts of the line.
+        /// </summary>
+        protected double[] ActualExtrapolationDashArray
+        {
+            get
+            {
+                return this.Dashes ?? this.ActualExtrapolationLineStyle.GetDashArray();
```
VisualMelon, line 95: This means it will use the 'normal' dash style if it was set: we should probably provide another `ExtrapolationLineDashes`.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
```diff
@@ -0,0 +1,322 @@
...
+            }
+        }
+
+        /// <summary>
+        ///  Overridden to show solid style.Renders the legend symbol for the line series on the
```
VisualMelon, line 100: Needs tidying a litte.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
```diff
@@ -0,0 +1,322 @@
...
+
+        /// <summary>
+        /// Updates the maximum and minimum values of the series.
+        /// </summary>
+        protected internal override void UpdateMaxMin()
+        {
+            if (this.IgnoreExtraplotationForScaling && this.orderedIntervals.Any())
+            {
+                this.MinX = this.Points
+                    .Where(p => !this.orderedIntervals.Any(i => i.Contains(p.X)))
```
VisualMelon, line 159: Probably not a big deal, but it would be nice to use a binary search for this, so that it scales better with large numbers of intervals.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
```diff
@@ -0,0 +1,322 @@
...
+            {
+                IOrderedEnumerable<DataRange> ordered = intervals.OrderBy(i => i.Minimum);
+
+                foreach (DataRange current in ordered)
+                {
+                    DataRange previous = orderedList.LastOrDefault();
+
+                    if (current.IntersectsWith(previous))
+                    {
+                        orderedList[orderedList.Count - 1] = new DataRange(previous.Minimum, current.Maximum);
```
VisualMelon, line 310: max values should be `Math.Max(previous.Maximum, current.Maximum)`, otherwise a wider interval may 'consume' a smaller interval.

VisualMelon (commented):
Regarding all the git stuff, basically the idea is that each PR should - in the end - be one trivially mergeable commit. You find more info here: https://github.com/oxyplot/oxyplot/blob/develop/.github/CONTRIBUTING.md#commit-your-changes

I think we can do the squash as part of the merge process, and you your branch doesn't currently need rebasing, though it may become necessary if another PR gets merged in before yours.

CHANGELOG.md
Doeharrrck: Done.

Source/OxyPlot/Series/DataRange.cs
Doeharrrck: Done.

Source/OxyPlot/Series/DataRange.cs
Doeharrrck: Done.

Source/OxyPlot/Series/DataRange.cs
Doeharrrck: Done.

Source/OxyPlot/Series/DataRange.cs
Doeharrrck: Done.

Source/OxyPlot/Series/DataRange.cs
Doeharrrck: Done.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
Doeharrrck: Done. Also added a new example to test it to the browser.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
Doeharrrck: Done.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
Doeharrrck: Done. I have added an example with 100,000 points and 1000 intervals to the browser.

Source/OxyPlot/Series/ExtrapolationLineSeries.cs
Doeharrrck: Fixed.

Source/OxyPlot/Series/DataRange.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        /// <summary>
+        /// Gets the upper bound of the data range.
+        /// </summary>
+        public double Maximum => this.maximum;
+
+        /// <summary>
+        /// Gets the difference between maximum and minimum.
+        /// </summary>
+        public double Range => this.Maximum - this.Minimum;
```
VisualMelon, line 64: Maybe not the best name... but I can't think of a better one.

Source/OxyPlot/Series/DataRange.cs
```diff
@@ -0,0 +1,119 @@
...
+
+        /// <summary>
+        /// Determines whether this data range is defined.
+        /// </summary>
+        /// <returns><c>false</c> if minimum or maximum is NaN, otherwise<c>true</c>.</returns>
```
VisualMelon, line 69: No longer true

VisualMelon:
I'll take another proper look over the weekend: thanks for your continued work!

Source/OxyPlot/Series/DataRange.cs
Doeharrrck: Measurement studio has a similar class named `Range` where the corresponding property is named 'Interval'. Another alternative might be 'Span'. I personally don't have a strong opinion about which name is best.

Source/Examples/ExampleLibrary/Series/ExtrapolationLineSeriesExamples.cs
```diff
@@ -0,0 +1,492 @@
...
+        /// </summary>
+        /// <returns>A <see cref="PlotModel" />.</returns>
+        [Example("Intersecting Intervals")]
+        public static PlotModel IntersectingIntervals()
+        {
+            var model = new PlotModel { Title = "Intersecting Intervals" };
+
+            var lineSeries = new ExtrapolationLineSeries
+            {
+                Title = "Overlap between 0 and 30, double-overlap between 10 and 20",
```
VisualMelon, line 163: Omits the `[-20, 20]` range.

Source/OxyPlot.Tests/Series/DataRangeTests.cs
```diff
@@ -0,0 +1,152 @@
...
+// </copyright>
+// <summary>
+//   Tests the <see cref="DataRange" /> struct.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Tests.Series
+{
+    using System;
+    using System.Collections.Generic;
```
VisualMelon, line 13: Unnecessary `usings`

Source/OxyPlot.Tests/Series/DataRangeTests.cs
```diff
@@ -0,0 +1,152 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="DataRangeTests.cs" company="OxyPlot">
+//   Copyright (c) 2014 OxyPlot contributors
+// </copyright>
+// <summary>
+//   Tests the <see cref="DataRange" /> struct.
+// </summary>
+// --------------------------------------------------------------------------------------------------------------------
+
+namespace OxyPlot.Tests.Series
```
VisualMelon, line 10: I don't think the `.Series` namespace is necessary: other series specific tests are just thrown in `OxyPlot.Tests`, and the `DataRange` is pretty generic anyway.

VisualMelon (approved):
Sorry this is a week late.

There are a couple of minor things in the tests and examples, but I think this is basically good to go. I'll wait for @oxyplot or @Jonarw to approve before merging because this adds quite a lot.

Doeharrrck:
Thank you, that's great! I think I have just addressed your last remarks.

Jonarw (commented):
Thanks for this PR and sorry for taking so long!
Looks very nice to me, I just have two small nitpicks - other than that, I think this is good to merge.

PR #1764, opened 2021-05-06
Fix LineAnnotation Text Placement on Reversed Axes (#1741)

### LINKED ISSUE #1741: LineAnnotation text clipping on reversed axes
### Steps to reproduce

1. View example `LineAnnotation` in an example browser, and reverse the axes.
2. The 'First' annotation is clipped

image

The text should be positioned within the plot bounds, as occurs on a non-reversed axes, but is not. This appears to be mitigated by uncommenting this line:

https://github.com/oxyplot/oxyplot/blob/61ffa99e16bb6f320326d117a0fd6e86bea71634/Source/OxyPlot/Annotations/TextualAnnotation.cs#L77

The line was introduced (as a comment) in https://github.com/oxyplot/oxyplot/pull/1435/files#diff-9b25104c8bfedc3c97747a9d1e89366896b9f1ed1e70fd1f2101691da6eb3a8eR77

@Jonarw do you recall if there was a particular reason this was disabled? There are other problems with annotations on line plots (e.g. the text doesn't clamp to the Y axes, which would be nice), so we should probably review them all at some point.

VisualMelon:
Fixes #1741 and an old bug where labels are rendered on LineAnnotation that are off the far side of the plot.

### Checklist

- [ ] I have included examples or tests -> existing examples suffice

- Don't reverse line positions on reversed axes
- Don't render `LineAnnotation`s where the max>min in either X or Y dimension

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Annotations/PathAnnotation.cs
+++ b/Source/OxyPlot/Annotations/PathAnnotation.cs
@@ -145,6 +145,11 @@ namespace OxyPlot.Annotations
 
             this.CalculateActualMinimumsMaximums();
 
+            if (ActualMinimumX > ActualMaximumX || ActualMinimumY > ActualMaximumY)
+            {
+                return;
+            }
+
             this.screenPoints = this.GetScreenPoints();
 
             const double MinimumSegmentLength = 4;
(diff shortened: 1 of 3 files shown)


Source/OxyPlot/Annotations/PathAnnotation.cs
```diff
@@ -145,6 +145,11 @@ public override void Render(IRenderContext rc)
 
             this.CalculateActualMinimumsMaximums();
 
+            if (ActualMinimumX > ActualMaximumX || ActualMinimumY > ActualMaximumY)
```
Jonarw, line 148: Could we have some `this` here?

Jonarw (approved):
Apart from a small nitpick (see comment), looks good to me!

VisualMelon:
Thanks @Jonarw !

That's all the issues I had intended to fix before releasing 2.1 proper, and I'll try to make a 2.1 branch in my fork soon which only has the bug-fix commits.

PR #1783, opened 2021-08-31
Update changelog for 2.1.0

VisualMelon:
Updates the changelog for 2.1.0.

Provisional release notes at https://gist.github.com/VisualMelon/1c7fb55e0e915900544dcaf118c8d6dc

Again, I can't remember how release branches work, and I need to re-test all the deployment scripts, but hopefully these texts are OK.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,32 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
-## Unreleased
-
-### Added
-- Example for Issue #1716 showing poor tick spacing on DateTimeAxis with interval types of Weeks or Years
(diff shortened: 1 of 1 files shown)


VisualMelon:
I've tested the scripts, so that's all fine.

I think we should hold off until we work out what to do with #1787 which seems like a pretty serious regression, but should be easy to characterise.

VisualMelon:
This has been updated again; any last concerns? If not, I'll merge this tomorrow, and then try to work out how to make the release/2.1.0 branch and all those things before pushing everything to NuGet.

Provisional release notes here: https://gist.github.com/VisualMelon/1c7fb55e0e915900544dcaf118c8d6dc

Issue #1785, opened 2021-09-03
Plots with SkiaSharp renderer are blurry after ScaleTransfom with scaling factor unequal to 1 


albtaeler:
### Description
We use OxyPlot in an application for plots on a page that looks like a paper page. In it we use a ScaleTransform to fit the pages to the window ("Fit width", "Fit page", "Zoom in", etc).
For performance reasons we tried to switch to SkiaSharp Render, unfortunately not with the expected result.

### Steps to reproduce

1. Create a simple plot with a OxyPlot.SkiaSharp.Wpf.PlotView
2. Add a LayoutTransform with a scaling factor unequal to 1 (eg. 1.2)  to the plot.

```
public partial class MainWindow : Window
    {
        public MainWindow()
        {
            InitializeComponent();
            var plotModel = new PlotModel { Title = "SkiaSharp Renderer" };
            plotModel.Axes.Add(new LinearAxis { Position = AxisPosition.Left });
            plotModel.Axes.Add(new LinearAxis { Position = AxisPosition.Bottom });
            plotModel.Series.Add(new FunctionSeries(Math.Sin, 0, 10, 1000));
            plotModel.Series.Add(new FunctionSeries(x => Math.Sin(x) / x, 0, 10, 1000));

            this.Plot = plotModel;
            this.DataContext = this;
            PlotGrid.LayoutTransform = new ScaleTransform(1.2, 1.2);
        }

        public PlotModel Plot { get; set; }
    }
```

Platform: x86 & x64
.NET version: 4.7.2
Latest OxyPlot Version (nuget packages build from develop brunch)

### Expected behaviour

Plots should be sharp and pixel accurate as with the standard canvas renderer.

Plot with standard canvas renderer and ScaleTransfor, scaling factor 1.2:
![image](https://user-images.githubusercontent.com/84440369/131969251-85ab3936-f0f8-4954-8897-ba8999a5aa82.png)

### Actual behaviour

The hole plot is blurry, including axes and plot text.

Plot with SkiaSharp renderer and ScaleTransfor, scaling factor 1.2:
![image](https://user-images.githubusercontent.com/84440369/131967675-b9c8d1c5-25c2-4d30-8f74-b1cd99dd28fa.png)

VisualMelon:
This is because the backing buffer has a low resolution.
 - Need to adjust DpiScale (`PlotView.UpdateDpi`)
 - Need to adjust buffer size (`OxySKElement.CreateSize`)

[...]

PR #1792, opened 2021-09-11
Add OxyPlot.SkiaShap.Wpf PlotView support for simple render transforms

VisualMelon:
A jab at addressing #1785.

### Checklist

- Adds a `GetRenderScale` method to `OxySKElement` which tried to detect the render transform on the control and produce a single value scaling
- The value is used to change the render buffer width/height and effective DPI
- Adds inline doc for OxySKElement

It seems to work in the simplest case (a uniform scaling), but I've not thoroughly checked the implementation of `GetRenderScale` (are the matmuls backwards?) and we probably need an example to go with it. Also need to see what happens when you rotate one of these, to check it doesn't do something completely ridiculous, though I don't think that's a scenario we should be trying to support.

The `GetRenderScale` method is intentionally virtual so that it can be overridden (e.g. so that a user could change it if it's assumptions are not appropriate for their use-case). It must be public so that `PlotView` can determine the change in effective DPI. Importantly, no attempt is made to support non-uniform scaling.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to this project will be documented in this file.
 - Add option to check distance for result between data points (#1736)
 - Legend.AllowUseFullExtent property to control whether legends should be able to use the full extent of the plot (#1743)
 - Legend.ShowInvisibleSeries property to control whether invisible series should be shown on the legend (#1730)
+- OxyPlot.SkiaSharp.Wpf PlotView support for simple render transforms (#1785)
 
 ### Changed
 - Updated Series.cd with ExtrapolationLineSeries and removed classes that do not exist anymore
--- a/Source/Examples/WPF/ExampleBrowser/MainWindowViewModel.cs
+++ b/Source/Examples/WPF/ExampleBrowser/MainWindowViewModel.cs
@@ -94,7 +94,7 @@ namespace ExampleBrowser
 
         private void CoerceRenderer()
         {
-            ((IPlotModel)this._PlotModel).AttachPlotView(null);
+            ((IPlotModel)this._PlotModel)?.AttachPlotView(null);
             this.RaisePropertyChanged(nameof(this.SkiaModel));
             this.RaisePropertyChanged(nameof(this.CanvasXamlModel));
             this.RaisePropertyChanged(nameof(this.CanvasModel));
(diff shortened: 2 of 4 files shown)


VisualMelon:
(Latest commit shouldn't be merged; just for testing)

I've pushed a change to the example browser so that it by default scales the plot up 3 times (pictures below), and you can change the scale with a slider for testing. Noteably, the non-Skia renderers do not handle text scaling well:

image

I'll take a look at how the text is managed with the non-Skia renderers, and see if we can do anything about that.

Skia looks pretty crisp (note: skia isn't doing sub-pixel rendering)

image

VisualMelon:
Canvas render text is fine when `TextRenderingMode` is set to `Ideal` (rather than `Display`). I'm not sure what that tells us, but it is configurable, so does provide a workaround.

VisualMelon:
Force pushed one without all the test edits; I've checked the matrices and such now and I think the internals are OK.

I'm open to suggestions for tidy examples we would include; doesn't seem sane to create a whole new project or to mess with the Example Browser.

Is it OK to include this in 2.1? It's a bit wild, but the whole point is to make Skia available, and this seems like an important scenario.

albtaeler:
Thanks!
I tested it with out plot application, the plots are quite sharp and crisp. This solves the issue #1785 for us. I hope you can merge and include it in 2.1.

VisualMelon:
@albtaeler thanks for testing this!

Jonarw:
> I'm open to suggestions for tidy examples we would include; doesn't seem sane to create a whole new project or to mess with the Example Browser.

I think this could be potentially a good place for such an example:
https://github.com/oxyplot/oxyplot/tree/develop/Source/Examples/WPF/WpfExamples

Though at the moment this project doesn't currently include any examples using the Skia renderer, which is kind of the point of this PR...

But to be pragmatic I think for now it would be ok to just add an example with a `ScaleTransform` to `WpfExamples` using just the Canvas renderer. For quick testing of the functionality we can just manually swap out the `PlotView`, and maybe we can come up with a way to include both renderers in `WpfExamples` in the future - but this is out of the scope of this PR I would say.

> Is it OK to include this in 2.1? It's a bit wild, but the whole point is to make Skia available, and this seems like an important scenario.

I agree, would be nice to include this in 2.1 as it's technically a bug fix.

VisualMelon:
(latest commit fixes DPI problems which I'd overlooked; can someone with a High DPI screen check this?)

@Jonarw I'd thought about adding to WpfExamples, but as you say, it doesn't ref the Skia project at the moment, and this PR doesn't change the Canvas behaviour (though we could add an example showing how to change the `TextRenderingMode`).

VisualMelon:
@Jonarw any complaint with merging this as-is and releasing 2.1 this weekend? (I have much more time now, so no excuses at my end)

Jonarw:
Sorry for not replying sooner. No objections from my side, go ahead :)

PR #1797, opened 2021-10-05
Fix "WPF- OxyPlot doesn't render inside a Popup" (#1796)

### LINKED ISSUE #1796: WPF- OxyPlot doesn't render inside a Popup (fix included)
Doing my best to format it, as I'm not a git expert by any means.

### Steps to reproduce

1. Put a plotview inside a WPF Popup
2. Display the popup
3. See a white graph

Platform: Windows
.NET version: 4.6.1 and others

### Expected behaviour

The graph should appear normally inside a popup

### Actual behaviour

The graph appears Empty

### Why does it happen ?

In [PlotViewBase](https://github.com/oxyplot/oxyplot/blob/release/v2.1.0/Source/OxyPlot.Wpf.Shared/PlotViewBase.cs) , you're checking if the graph is part of the visual tree by checking if the ancestor is a window :

```
       /// <summary>
        /// Gets a value indicating whether the <see cref="PlotViewBase"/> is connected to the visual tree.
        /// </summary>
        /// <returns><c>true</c> if the PlotViewBase is connected to the visual tree; <c>false</c> otherwise.</returns>
        private bool IsInVisualTree()
        {
            DependencyObject dpObject = this;
            while ((dpObject = VisualTreeHelper.GetParent(dpObject)) != null)
            {
                if (dpObject is Window)
                {
                    return true;
                }
            }

            return false;
        }
```

In the case of a WPF popup, the ancestor is a Popuproot : 
image

### The fix

We need to check if the ancestor is also a Popuproot. If so, it's in the visual tree. But as the popuproot is an internal type, you can't check against it easily. The best way I found is to check if the logical parent of the PopupRoot is a Popup : 
```
        /// <summary>
        /// Gets a value indicating whether the <see cref="PlotViewBase"/> is connected to the visual tree.
        /// </summary>
        /// <returns><c>true</c> if the PlotViewBase is connected to the visual tree; <c>false</c> otherwise.</returns>
        private bool IsInVisualTree()
        {
            DependencyObject dpObject = this;
            while ((dpObject = VisualTreeHelper.GetParent(dpObject)) != null)
            {
                if (dpObject is Window)
                {
                    return true;
                }

                //Check if the logical parent is a popup. If so, we found the popuproot
                var logicalRoot = LogicalTreeHelper.GetParent(dpObject);
                if (logicalRoot is Popup)
                {
                    // popup root found here
                    return true;
                }
            }

            return false;
        }
```

This isn't enough, as in [PlotView](https://github.com/oxyplot/oxyplot/blob/release/v2.1.0/Source/OxyPlot.Wpf/PlotView.cs), you're getting the window parent from the Visual tree using this call : 

```
        /// <summary>
        /// Returns a reference to the window object that hosts the dependency object in the visual tree.
        /// </summary>
        /// <returns> The host window from the visual tree.</returns>
        private Window GetAncestorWindowFromVisualTree(DependencyObject startElement)
        {
            DependencyObject parent = startElement;
            while (!(parent is Window))
            {
                if (parent == null) { break; }
                parent = VisualTreeHelper.GetParent(parent);
            }
            return parent as Window ?? Window.GetWindow(this);
        }
```

I find this Window check superfluous, as if you're there, you have already checked you're in the visual tree. You can thus just walk the visual tree and get the root here, no check needed, and you only need a Visual type in the call, not a Window type : 
```
        private Visual GetAncestorVisualFromVisualTree(DependencyObject startElement)
        {
            
            DependencyObject child = startElement;
            DependencyObject parent = VisualTreeHelper.GetParent(child);
            while (parent != null)
            {
                child = parent;
                parent = VisualTreeHelper.GetParent(child);
            }

            return child is Visual visualChild ? visualChild : Window.GetWindow(this);
        }
```

This makes Oxyplot work flawlessly in a popup.

Feel free to correct me and point any mistakes !

Following writing all this, I'll try to a PR - not sure I'll succeed though :D

Tetedeiench:
Fix "WPF- OxyPlot doesn't render inside a Popup" (#1796)

Fixes #1796 .

### Checklist

- [ X ] I have updated the change log
- [ X ] I am listed in the CONTRIBUTORS file
- [ X ] I have cleaned up the commit history (use rebase and squash)

- WPF- OxyPlot doesn't render inside a Popup (fix included) #1796
-
-

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+### Fixed
+- WPF - OxyPlot doesn't render inside a Popup (#1796)
+
 ## [2.1.0] - 2021-10-02
 
 ### Added
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -10,6 +10,7 @@
 
 # Please keep the list sorted.
 
+Adrien Mercier
 Alexei Shcherbakov
 Anders Musikka <anders@andersmusikka.se>
 Auriou
--- a/Source/OxyPlot.WPF.sln.DotSettings
+++ b/Source/OxyPlot.WPF.sln.DotSettings
@@ -529,6 +529,7 @@
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=StaticReadonly/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=TypeParameters/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="T" Suffix="" Style="AaBb" /&gt;</s:String>
 	<s:String x:Key="/Default/CodeStyle/Naming/CSharpNaming/PredefinedNamingRules/=TypesAndNamespaces/@EntryIndexedValue">&lt;Policy Inspect="True" Prefix="" Suffix="" Style="AaBb" /&gt;</s:String>
+	<s:Boolean x:Key="/Default/Environment/SettingsMigration/IsMigratorApplied/=JetBrains_002EdotCover_002EIde_002ECore_002EFilterManagement_002EModel_002ESolutionFilterSettingsManagerMigrateSettings/@EntryIndexedValue">True</s:Boolean>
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,9 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+### Fixed
```
VisualMelon, line 4: Should ideally have the unrelease/added/changed/fixed/removed sections like in https://raw.githubusercontent.com/oxyplot/oxyplot/StemSeriesActualMarker/CHANGELOG.md , but we can do that separately

VisualMelon (commented):
I've tested and the fix looks good to me.

We probably need an example (so that it's easier to test in future); I put a simple example together in https://github.com/VisualMelon/oxyplot/tree/bugfix-wpf-popup which we could use, but if you have a better idea for a use-case that involves popups that would be welcome!

VisualMelon:
Fixed Skia stuff in my branch, and added a pointless/fun popup pre-view for the example browser (so we can accidently test popups _and_ Skia scaling) at the same time, all the time)

PR #1799, opened 2021-10-08
Fix NullReference in SkiaSharp WPF renderer if UIElement has no Prese…

### LINKED ISSUE #1798: OxyPlot SkiaSharp renderer throws a null reference exception if the rendered UIElement has no PresentationSource
### Description

To export whole plot pages (like paper pages) as PNG we create a new ViewModel in our MVVM WPF application only for the export which is not displayed on the screen. We don't use the oxyplot pngexporter for this because the pages contains also textboxes and other elements.
After updating to the latest OxyPlot version and switching to SkiaSharp Render, this no longer works.
We get a NullreferenceException when exporting.

### Steps to reproduce

Create and render a PlotView with code like this:

```
var mainGrid = new Grid() { Height = 400, Width = 600 };
mainGrid.Children.Add(new PlotView() { Model = plotModel });
var bmp = FrameworkElementToBitmap(mainGrid);
WritePngToFile(@"ExportTestSkiaRenderer.png", bmp);
```
Use a  RenderTargetBitmap to render it as an image.
```
        private static BitmapSource FrameworkElementToBitmap(FrameworkElement frameworkElement)
        {

            var size = new Size(frameworkElement.Width , frameworkElement.Height );
            frameworkElement.Measure(size);
            frameworkElement.Arrange(new Rect(size));
            frameworkElement.UpdateLayout();
            var bmp = new RenderTargetBitmap((int)size.Width, (int)size.Height, 96.0, 96.0, PixelFormats.Pbgra32);
            bmp.Render(frameworkElement);
            return bmp;
        }


        private void WritePngToFile(string fileName, BitmapSource bitmapImage)
        {
            using (var fileStream = File.Create(fileName))
            using (var pngStream = new MemoryStream())
            {
                var encoder = new PngBitmapEncoder();
                encoder.Frames.Add(BitmapFrame.Create(bitmapImage));
                encoder.Save(pngStream);
                pngStream.WriteTo(fileStream);
            }
        }
```
This causes a NullReferenceException at:
```
var m = PresentationSource.FromVisual(this).CompositionTarget.TransformToDevice;            
scaleX = m.M11;
scaleY = m.M22;
```

and at 
```
// get window to screen offset
var visualOffset = this.TransformToAncestor(GetAncestorWindowFromVisualTree(this)).Transform(default);
```
in [OxySKElement.cs](https://github.com/oxyplot/oxyplot/blob/release/v2.1.0/Source/OxyPlot.SkiaSharp.Wpf/OxySKElement.cs)

Platform: x86 & x64
.NET version: 4.7.2
Latest OxyPlot Version (nuget packages build from develop brunch)

### Expected behaviour
Rendering should be possible as in older versions. 
Version f9cebf3b95cecc76edb5e7b03824337764c24caf from 2019-01-31 didn't have this issue.

### Actual behaviour
When exporting as described above we get a NullreferenceException.

### Possible Fix
Return the default value if the PresentationSource or the ancestor visual is null:

```
            var compositionTarget = PresentationSource.FromVisual(this)?.CompositionTarget;
            if (compositionTarget != null)
            {
                var m = compositionTarget.TransformToDevice;
                scaleX = m.M11;
                scaleY = m.M22;
            }
```

and

```
            // get window to screen offset
            var ancestor = GetAncestorWindowFromVisualTree(this);
            var visualOffset = ancestor!=null ? this.TransformToAncestor(ancestor).Transform(default) : default;

```

albtaeler:
…ntationSource  (#1798)

Fixes #1798 .

### Checklist

- Return/use the default values if the PresentationSource or the ancestor visual is null:

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
+++ b/Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
@@ -373,10 +373,11 @@ namespace OxyPlot.Wpf
         /// </summary>
         protected void Render()
         {
-            if (this.plotPresenter == null || this.renderContext == null || !(this.isInVisualTree = this.IsInVisualTree()))
+            if (this.plotPresenter == null || this.renderContext == null)
             {
                 return;
             }
+            this.isInVisualTree = this.IsInVisualTree();
 
             this.RenderOverride();
         }
--- a/Source/OxyPlot.Wpf/PlotView.cs
+++ b/Source/OxyPlot.Wpf/PlotView.cs
@@ -127,8 +127,8 @@ namespace OxyPlot.Wpf
         protected override double UpdateDpi()
         {
             var scale = base.UpdateDpi();
-            this.RenderContext.DpiScale = scale;
-            this.RenderContext.VisualOffset = this.TransformToAncestor(this.GetAncestorVisualFromVisualTree(this)).Transform(default);
+            var ancestor = this.GetAncestorVisualFromVisualTree(this);
+            this.RenderContext.VisualOffset = ancestor != null ? this.TransformToAncestor(ancestor).Transform(default) : default;
             return scale;
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,17 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## Unreleased
+
+### Added
+
+### Changed
+
+### Removed
+
 ### Fixed
 - WPF - OxyPlot doesn't render inside a Popup (#1796)
+- NullReference in SkiaSharp WPF renderer if UIElement has no PresentationSource  (#1798)
 
 ## [2.1.0] - 2021-10-02
 
--- a/Source/OxyPlot.SkiaSharp.Wpf/OxySKElement.cs
+++ b/Source/OxyPlot.SkiaSharp.Wpf/OxySKElement.cs
@@ -150,7 +150,8 @@ namespace OxyPlot.SkiaSharp.Wpf
             this.bitmap.Unlock();
 
             // get window to screen offset
-            var visualOffset = this.TransformToAncestor(GetAncestorWindowFromVisualTree(this)).Transform(default);
+            var ancestor = GetAncestorVisualFromVisualTree(this);
+            var visualOffset = ancestor != null ? this.TransformToAncestor(ancestor).Transform(default) : default;
 
             // calculate offset to physical pixels
             var offsetX = ((visualOffset.X * scaleX) % 1) / scaleX;
@@ -192,9 +193,13 @@ namespace OxyPlot.SkiaSharp.Wpf
                 return new SKSizeI((int)w, (int)h);
             }
 
-            var m = PresentationSource.FromVisual(this).CompositionTarget.TransformToDevice;
-            scaleX = m.M11;
-            scaleY = m.M22;
+            var compositionTarget = PresentationSource.FromVisual(this)?.CompositionTarget;
+            if (compositionTarget != null)
+            {
+                var m = compositionTarget.TransformToDevice;
+                scaleX = m.M11;
+                scaleY = m.M22;
+            }
             return new SKSizeI((int)(w * scaleX), (int)(h * scaleY));
 
             static bool IsPositive(double value)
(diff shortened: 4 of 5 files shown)


VisualMelon:
Thanks for putting this together. I can't take a proper look now, but hopefully I'll be able to get back to you on this in the next few days.

albtaeler:
Another NullReferenceException appeared during tests of our application with the latest Oxyplot version.
I had to add this check in [Plotview.cs](https://github.com/oxyplot/oxyplot/blob/develop/Source/OxyPlot.SkiaSharp.Wpf/PlotView.cs) SkElement_PaintSurface().
```
        private void SkElement_PaintSurface(object sender, SKPaintSurfaceEventArgs e)
        {
            if (this.plotPresenter == null || this.renderContext == null)
            {
                return;
            }
```
Hopefully you @VisualMelon  or another maintainer will have time to review it soon.

VisualMelon (approved):
All looks good. Some duplication in #1820; I suggest we merge this and then work out if there is anything in #1820 that is missing.

VisualMelon:
Merged now so that we can look at #1820 properly, and because this was checked with the WPF and has 3 issues requesting the fix.

Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
```diff
@@ -373,10 +373,11 @@ protected void OnModelChanged()
         /// </summary>
         protected void Render()
         {
-            if (this.plotPresenter == null || this.renderContext == null || !(this.isInVisualTree = this.IsInVisualTree()))
```
HavenDV, line 376: This raises questions too - it cannot throw a NullReferenceException. If it's just a refactoring - then the behavior is changed.

Source/OxyPlot.Wpf/PlotView.cs
```diff
@@ -127,8 +127,8 @@ protected override void RenderOverride()
         protected override double UpdateDpi()
         {
             var scale = base.UpdateDpi();
-            this.RenderContext.DpiScale = scale;
```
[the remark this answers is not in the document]
VisualMelon: Thanks for spotting this.

For Skia/WPF it's fine I believe as it's handled at a different level, but does look like a regression for non-Skia/WPF which is a bit hard to spot, but the most obvious thing is that axis lines are fuzzy when the scale isn't `1`.

@albtaeler do you recall if there was a particular reason for removing this, or did it just look redundant?

Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
VisualMelon: This was mentioned in https://github.com/oxyplot/oxyplot/issues/1798#issuecomment-964202536 but looking briefly now it does look a bit odd.

I'll try to find time to take another look at this (I should have some tests lying around somewhere....); my WPF knowledge is pretty limited, so I'd welcome any input from you and @albtaeler.

HavenDV: Didn't see the comment. Considering it, it's all right. The only question is - did this check play any role in performance?

albtaeler: I believe the only reason for this check was that this.TransformToAncestor() fails with a NullReferenceException without my changes. As mentioned in https://github.com/oxyplot/oxyplot/issues/1798#issuecomment-964202536 I changed this behavior, because rendering it's not possible if it isn't in the VisualTree, but sometimes needed.
I could not see any performance changes. It is important that the property `this.isInVisualTree = this.IsInVisualTree();` is set.

Source/OxyPlot.Wpf/PlotView.cs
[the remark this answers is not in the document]
VisualMelon: @albtaeler thanks for taking another look; feel free to open a PR putting this back in; hopefully I'll get around to it eventually otherwise.

Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
HavenDV: I was confused at this point only because I don't see a place where IsInVisualTree() could throw a NullReferenceException.

PR #1815, opened 2021-11-01
Links to issues in CHANGELOG.md

Lette:
### Checklist

- De-linted README.md and CHANGELOG.md, using provided markdownlint configuration.
- All references to issues in the CHANGELOG are now links to the respective issues.

If you think it's good, keep it. If not, no argument from me.
Thank you for this awesome library!

@oxyplot/admins

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -1,0 +1,5 @@
+{
+    "default": true,
+    "line-length": false,
+    "no-duplicate-heading": { "siblings_only": true }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,17 @@
 # Change Log
+
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
 ### Fixed
+
 - WPF - OxyPlot doesn't render inside a Popup (#1796)
 
 ## [2.1.0] - 2021-10-02
 
 ### Added
+
 - Made Legend Items clickable to toggle series visibility (#644)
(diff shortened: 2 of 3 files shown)


VisualMelon:
Links are certainly nice!

In-line links might be preferable to keep changes simple.

In any case, I won't merge this myself, but I certainly don't object to the changes.

Lette:
I considered inline links but went the other way because the same issues are referenced in multiple places...

I can easily change if you want to.

VisualMelon:
What does the markdown lint file achieve?

On reflection, I think separating the links out would be better; it'll be a little more effort for contributors and maintainers, but it'll keep the content uncluttered.

I'm not sure how this will work when it is merged, so I'll hang-fire on that for the moment (but otherwise approved).

Jonarw:
This PR came up in the discussion in #2012. Initially the links seemed like a nice idea to me. However, after thinking a bit more about it, I am not so sure anymore whether the comfort of these links outweighs the added maintenance effort. I don't have a strong opinion either way, probably leaning a bit against it. From what I understand from #2012, you were also a bit skeptical of this @VisualMelon?

VisualMelon:
@Jonarw From memory I thought I was more opposed to this than it seems I actually was, or indeed am now that I've looked at this properly again. Like you, I'm happy either way, but would probably lean towards leaving things as-is so this isn't another thing that contributors have to worry about.

Jonarw:
Yes, agreed, so I will close this for now. Thanks anyway @Lette!

PR #1819, opened 2021-11-07
Axis should always honor absolute minimum /maximum as well as minimum/maximum range properties

### LINKED ISSUE #1812: Absolute limits and MinimumRange not always honored
### Steps to reproduce

Create an empty model with a range-constrained axis but no data, like so:
```c#
var pm = new PlotModel();
pm.Axes.Add(new LinearAxis
{
    Position = AxisPosition.Bottom,
    AbsoluteMinimum = -200.0,
    AbsoluteMaximum = -100.0,
    MinimumRange = 10.0,
});
plot1.Model = pm;
```
I used the WindowsFormsDemo but many of the examples should be usable as well.

Platform: OxyPlot.WinForms on Windows 10 64 bit
.NET version: 4.8

### Expected behaviour

I would expect the axis to honour the absolute limits and the minimum range, so any of these visible ranges would be fine:

- `[-200.0, -100.0]`
- `[-200.0, -190.0]`
- `[-155.0, -145.0]`
- `[-110.0, -100.0]`

### Actual behaviour

Instead I get the range `[-100.0, -99.0]` which violates both the `AbsoluteMaximum` and the `MinimumRange` properties:

oxyplot_wrong_range

### Cause and proposed fix

The problem is the following code in `Axis.cs`:

```c#
protected virtual void CoerceActualMaxMin()
{
    // Coerce actual minimum
    if (double.IsNaN(this.ActualMinimum) || double.IsInfinity(this.ActualMinimum))
    {
        this.ActualMinimum = 0;
    }

    // Coerce actual maximum
    if (double.IsNaN(this.ActualMaximum) || double.IsInfinity(this.ActualMaximum))
    {
        this.ActualMaximum = 100;
    }
```

which uses values outside the allowed range. A possible fix would be to change this to:

```c#
        this.ActualMinimum = Math.Max(0, this.AbsoluteMinimum);
```
and 
```c#
        this.ActualMaximum = Math.Min(100, this.AbsoluteMaximum);
```
respectively. With these changes I get the initial display range `[-110.0, -100.0]` which is valid.

mghie:
Fixes #1812.

### Checklist

- [*] I have included examples or tests
- [*] I have updated the change log
- [*] I am listed in the CONTRIBUTORS file

- check of property consistency moved to beginning of `Axis.CoerceActualMaxMin()`
- calculation of `ActualMaximum` and `ActualMinimum` improved to always honor absolute minimum and maximum as well as minimum and maximum range
- tests added

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 - SkiaSharp.WPF - OxyPlot doesn't render inside an ElementHost (#1800)
 - NullReference in SkiaSharp WPF renderer if UIElement has no PresentationSource  (#1798)
 - WPF DPI Regression (#1799)
+- Axes not always honoring AbsoluteMinimum/AbsoluteMaximum and/or MinimumRange/MaximumRange properties (#1812)
 
 ## [2.1.0] - 2021-10-02
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -90,6 +90,7 @@ Memphisch <memphis@machzwo.de>
 Mendel Monteiro-Beckerman
 Menno Deij - van Rijswijk <m.deij@marin.nl>
 methdotnet
+Michael Hieke <mghie@gmx.net>
 Mikant <a.mikant@gmail.com>
 mirolev <miroslav.levicky@gmail.com>
 Mitch-Connor <acm@htri.net>
--- a/Source/OxyPlot.Tests/Axes/AxisTests.cs
+++ b/Source/OxyPlot.Tests/Axes/AxisTests.cs
@@ -518,6 +518,58 @@ namespace OxyPlot.Tests
         }
 
         /// <summary>
+        /// Tests the initial axis range, if absolute and minimum range are set and data is partially out-of-range
+        /// </summary>
+        [Test]
+        public void Axis_MinimumRange_AbsoluteRange_DataOutOfRange()
+        {
+            var plot = new PlotModel();
+            var yaxis = new LinearAxis()
+            {
+                Position = AxisPosition.Bottom,
+                AbsoluteMinimum = -100,
+                AbsoluteMaximum = 0,
+                MinimumRange = 100,
+            };
+
+            plot.Axes.Add(yaxis);
+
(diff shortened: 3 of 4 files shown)


VisualMelon:
Thanks for putting this together! I'm pretty busy for the next few days, but will try to take a proper look at Thursday.

Feel free to ping me if I don't get back to you or if you have any questions.

VisualMelon (dismissed):
I had some reservations about the behaviour around `NaNs` here, but I think that's an edge case we probably don't need to worry about for now given how limited the scenarios are where this behaviour would be applicable.

(e.g. the implicit NaN check on line 1602 in Axis isn't very obvious; might be worth making that and other implicit checks more explicit)

VisualMelon:
@Jonarw would you mind sanity checking this? I'm always nervous about merging this sort of change without a second review.

VisualMelon:
@mghie that last commit has gone wrong, I think: are you able to revert and rebase against the develop branch?

mghie:
@VisualMelon I did a revert and rebase (first time I did such a thing, hopefully the right way), does it look better now? I have to say that I'm not really sure what I'm doing here, learning along the way, please be patient...

VisualMelon:
No problem, I should have been clearer. What you have looks fine when I look at it locally, but I'm not sure what is going on with the github diff interface.

Instead of that, can you do a hard reset to commit f3fab4a6349a160d807d0d70d09ad2811b5d4422 (may want to make a backup first) and force-push that over the top? That should remove the new commits, and just leave you with the 4 proper commits.

You can achieve this by running:
    git reset --hard f3fab4a6349a160d807d0d70d09ad2811b5d4422
    (check that you still have all your changes e.g. with git log and git diff head~4)
    git push --force

Having done this locally, there will be a conflict with the changelog, but we should be able to fix that when we merge it. Rebasing would enable you to resolve this, but it's trivial so shouldn't be a problem: if there is a problem, we can deal with that when we get there.

VisualMelon (approved):

VisualMelon:
Thanks for rebasing as well!

mghie:
@VisualMelon Thanks for your support, I learned a few things today

PR #1821, opened 2021-11-09
Update ScatterSeries{T}.cs

wailandow:
if you do a live update of the chart, then you get an error "Collection was modified; enumeration operation may not execute", because the collection changes during the iteration

Fixes # .

### Checklist

-
-
-

@oxyplot/admins


VisualMelon:
The list copies are not atomic, so this only reduces the probability of something going wrong. To perform off-thread updates you must lock the `PlotModel.SyncRoot`. We've discussed mitigating this in the past, and I think that we should not make any attempt to reduce the probability of an exception being thrown (if anything, we should make it more likely, because off-thread operations can lead to unintended state).

PR #1831, opened 2021-12-01
Add property to customize tracker line thickness

mattico:
### Checklist

Added a property to customize the tracker line thickness.

Maybe we could have separate thickness properties for horizontal and vertical lines?
Maybe we could name it something else to specify that "line" refers to the lines projected on the X and Y axis from the tracker's position. I decided against this since e.g. HorizontalLineVisibility already just calls it "line".
Don't think this needs an example but I could add one.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+### Added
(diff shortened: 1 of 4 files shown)


VisualMelon (approved):
Looks good to me.

Issue #1835, opened 2021-12-25
Remove target frameworks v4 and v4.5


BobLd:
### Feature description
Remove target frameworks v4 and v4.5 so that the project can be build in Visual Studio 2022. From https://docs.microsoft.com/en-gb/dotnet/framework/install/guide-for-developers:
> Starting with Visual Studio 2022, Visual Studio no longer includes .NET Framework components for .NET Framework 4.0 - 4.5.1 because these versions are no longer supported. Visual Studio 2022 and later versions can't build apps that target .NET Framework 4.0 through .NET Framework 4.5.1. To continue building these apps, you can use Visual Studio 2019 or an earlier version.

VisualMelon:
Thanks for reporting this. Do we know what actually happens if you don't have .NET 4.0 installed? I'd have thought it would error for those targets, but should otherwise compile fine (I don't have a machine without .NET 4 so can't test myself at present).

BobLd:
As far as I see, in VS 2022, even with .NET 4.0 and  .NET 4.5 installed, it won't compiled.

I tried to (re)install .NET 4 to make sure I had it, but this is what I get (so I assume it's installed on my machine):
![image](https://user-images.githubusercontent.com/38405645/147384797-5ef1f2e5-f82b-48a9-8412-b468d112f846.png)

VisualMelon:
It'll be the SDK rather than the runtime; I don't know enough about how the toolchains work.

We could address this by making the target conditional (we don't want separate projects) or by dropping them altogether.

BobLd:
I cannot compile the solution at all... and you're right, sorry for that about sdk/runtime. But I think it is now relatively tricky to find the sdk...
![image](https://user-images.githubusercontent.com/38405645/147385186-c44d2400-a0d3-42fe-940c-d8adebb9a747.png)

VisualMelon:
Haha, don't worry about it.

Given the whole sln packs up we'd better do something about that. If we can find a suitable MSBUILD property to work out whether the frameworks are supported or not then a conditional in the projects would be nice; otherwise, it may be time to finally drop .NET 4.

BobLd:
haha I think it's the right time 😄

VisualMelon:
Can we get away with dropping 4.0 and bumping 4.5.1 to 4.5.2 (supported until April I think) for the time being so that VS is happy, and defer any further decisions until a later date?

BobLd:
[...]

PR #1836, opened 2021-12-27
Tidy up, modernise code and improve render performance

BobLd:
### Checklist

- [X] I have included examples or tests (no need, they already exist)

- Update `net40`/`net45` to `net452` cf #1835
- Tidy up and modernise code
- Tidy up `ContourSeries`, `DataPointSeries`, `ExtrapolationLineSeries`, `FunctionSeries`, `HeatMapSeries`, `HistogramSeries`, `ItemsSeries`, `LineSeries`, `RectangleSeries`, `ScatterErrorSeries`, `ScatterSeries`, `StairStepSeries`, `StemSeries`, `TwoColorAreaSeries`, `TwoColorLineSeries`, `XYAxisSeries`
- Improve render performance of `BoxPlotSeries` (legend only), `HighLowSeries` , `CandleStickAndVolumeSeries`, `CandleStickSeries`, `VolumeSeries`, `PieSeries` - instead of calling render every time in the loop, it's called only once

One element that would need a particular review is `Source/OxyPlot.Wpf/XpsExporter.cs` where I'm not sure that replacing `#if !NET40` by `#if !NETFRAMEWORK` is correct?

@VisualMelon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 ### Changed
-
+- Core - Update to net452, modernise code and improve rendering performance of financial series (#1836)
 ### Removed
 
 ### Fixed
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -17,6 +17,7 @@ Auriou
 Bartłomiej Szypelow <bszypelow@users.noreply.github.com>
 benjaminrupp
 Benoit Blanchon <>
+BobLd
 br
 brantheman
 Brannon King
--- a/Source/Examples/ExampleLibrary/ExampleLibrary.csproj
+++ b/Source/Examples/ExampleLibrary/ExampleLibrary.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;net45</TargetFrameworks>
+    <TargetFrameworks>netstandard1.0;netstandard2.0;net452</TargetFrameworks>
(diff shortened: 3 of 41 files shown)


VisualMelon:
At a glance this all looks good.

Apart from the conditional bit you've called out (not sure what the answer is right now), was this all (or mostly) automatic (e.g. VS Cleanup)?

BobLd:
It is yes, mainly applying VS Intellisense recommendation.

The actual changes are in `BoxPlotSeries` (legend only), `HighLowSeries` , `CandleStickAndVolumeSeries`, `CandleStickSeries`, `VolumeSeries`, `PieSeries`

VisualMelon:
Thanks for the annotations. I probably won't have time to look at this properly for a couple of days.

matsydoodles:
More food for thought but in future it might be better to split up changes like this into multiple PRs as code clean up inherently creates a lot of noise. I imagine this is getting squash committed?  Which means going back to view the change for the performance improvement gets lost in sea of changes...

On the performance changes did you perform any profiling to quantify the improvements at all?

BobLd:
@matsydoodles - agreed, it's quite messy right now. I'll create new PRs with the separate changes.

In term of performance improvements, I didn't quantify it but just used the examples available for WPF/Avalonia and compared the new version versus the old. Panning and zooming are much better and the plots can display more points.

If you know a way to quantify that, I'm more than happy to give it a go.

VisualMelon:
Keeping this PR to fairly mundane changes would be good (agree perf should be separate). This sort of PR is always a problem because it jumbles the history, so keeping it squashed and limited in scope is ideal.

Perf is always difficult to assess; historically we've avoided making changes that don't obviously improve matters, because while perf isn't a key objective of the project, scenario-specific regressions are always irritating.

BobLd:
closing to do separated PRs

Issue #236, opened 2014-11-06
Add ILabelProvider interface


objorke:
Use it where items series (e.g. `LineSeries`) wants a label from an item (e.g. render a label near a point).
When the item implements `ILabelProvider`, this should override the `LabelFormatString` property.
Implement similar to `IDataPointProvider`

objorke:
Most series contains an LabelFormatString property. I think this property can be moved to ItemsSeries...

objorke:
There could also be a `Func<object,string> LabelMapping` function, similar to the `Mapping` property.

mdhammad313:
@objorke, are we looking for any change or we are good with the PR of @tomaszkot

objorke:
Sorry, I am not active in this project at the moment.
@VisualMelon could you review the PR?
@oxyplot/admins

sachPundir:
> Most series contains a `LabelFormatString property`. I think this property can be moved to ` ItemsSeries`... Then it should be possible to create a method `string GetFormattedLabel(object item)` in `ItemsSeries`

From my review, it’s not most of the series — only a few specific series actually need the LabelFormatString property:

<img width="1203" height="1187" alt="Image" src="https://github.com/user-attachments/assets/df39e11d-d676-4062-ba95-780ae13e041e" />

Instead of placing this in ItemsSeries, I think XYAxisSeries is a more suitable location. However, many classes derived from XYAxisSeries do not require this property:

<img width="1359" height="1860" alt="Image" src="https://github.com/user-attachments/assets/90f73c5e-a696-405a-a8a9-b16fdc2729ba" />

Because of that, introducing an ILabelFormatProvider interface seems like the cleanest and most flexible solution.

If this approach sounds reasonable, please let me know. Once confirmed, I will prepare and submit a PR with the proposed solution.

PR #1838, opened 2022-01-15
So far only LabelFormatString moved to the base class, I would provide GetFormattedLabel method as  a separate commit

tomaszkot:
#236 Most series contains a LabelFormatString property. I think this property can be moved to ItemsSeries

Fixes # .

### Checklist

-
-
-

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file.
 ## [2.1.0] - 2021-10-02
 
 ### Added
+- It should be possible to create a method string GetFormattedLabel(object item) in ItemsSeries (#236)
 - Made Legend Items clickable to toggle series visibility (#644)
 - Added properties LegendKey and SeriesGroupName to Series, allowing grouping series between multiple legends and/or within same legend (#644)
 - OxyPlot.ImageSharp (#1188)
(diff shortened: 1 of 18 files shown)


tomaszkot:
Actually Unit Tests are shipped in the next bigger push containing GetFormattedLabel implementation.
I have used my other github account for the work (user ouadsecond), sorry about the confusion.

VisualMelon:
I'm not sure this is necessarily a good idea: the introduction of `GetInvalidOverloadUsed` is indicative that this isn't a solid abstraction; what's your thinking behind this change? Is this related to an open issues?

There has been discussion previously concerning inconsistencies between tracker formatters and methods like `GetValue` of which this reminds me, but I can't immediately find it.

tomaszkot:
It's based on that comment from https://github.com/oxyplot/oxyplot/issues/236   :

> _Most series contains a LabelFormatString property. I think this property can be moved to ItemsSeries...
> Then it should be possible to create a method string **GetFormattedLabel(object item)** in ItemsSeries__

Once I started to override GetFormattedLabel in derived classes I spotted some of them do not have LabelFormatString or need additional parameter to do formatting (for example  DataPoint point).  In such case there is a GetFormattedLabel added method with arguments needed in given specialization.

Jonarw:
Closing this according to the discussion in https://github.com/oxyplot/oxyplot/discussions/2012. If there is interest to continue with this topic, feel free to let us know in here or start a new PR.

PR #1857, opened 2022-02-23
SkiaSharp - Search for Fonts in EmbeddedResources

### LINKED ISSUE #1735: Unable to use custom font with SkiaSharp.Wpf
### Feature description

I should be able to use custom fonts from a local .ttf file using SkiaSharp.Wpf. 

Implementation suggestion:

In the current implementation of SkiaRenderContext, the DrawText method (and GetTextPaint method internally) use an SKTypeface cache (called typefaceCache). A method to add custom fonts to this cache would make custom fonts accessible. The SKTypeface.FromStream Method could be used in this process.

JensKrumsieck:
Adds EmbeddedResource based fonts to SkiaSharp Renderer as `SKTypeface.FromFamilyName` is not valid in e.g. WASM. 
Uses inspiration from https://www.blazorspread.net/blogview/82.
Tested with Blazor WASM but probably also Fixes #1735  .
Axis rendered with Montserrat Font in WASM
montserrat

### Checklist

- Added Dictionary for translating fontweights (int) to string as this is used for naming .ttf file when downloaded from e.g. GoogleFonts 
  ```csharp
  private readonly Dictionary<int, string> _fontWeights = new Dictionary<int, string>()
          {
              [100] = "Thin",
              [200] = "ExtraLight",
              [300] = "Light",
              [400] = "Regular",
              [500] = "Medium",
              [600] = "SemiBold",
              [700] = "Bold",
              [800] = "ExtraBold",
              [900] = "Black"
          };
  ```
- Search embedded resources if font ist not found by name
  ```csharp
  if (typeface.FamilyName != fontFamily) // requested font not found or is WASM
                  {
                      try
                      {
                          var assembly = Assembly.GetEntryAssembly(); // the executing program (the GUI Project (WPF, WASM, ...))
                          var weight = (_fontWeights.ContainsKey((int)fontWeight) ? _fontWeights[(int)fontWeight] : "Regular");
                          var filename = $"{fontFamily}-{weight}.ttf".ToLower();
                          Console.WriteLine($"Load Font {filename}");
  
                          var matches = assembly!.GetManifestResourceNames().Where(item => item.ToLower().EndsWith(filename));
                          if (!matches.Any()) matches = assembly!.GetManifestResourceNames().Where(item => item.ToLower().EndsWith(fontFamily + ".ttf"));
                          foreach (var item in matches)
                          {
                              var s = assembly.GetManifestResourceStream(item);
                              typeface = SKTypeface.FromStream(s);
                          }
                      }
                      catch
                      {
                          Console.WriteLine($"Requested Font {fontFamily} could not be found, falling back to {typeface.FamilyName}");
                      }
                  }
  ```
@oxyplot/admins

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -62,6 +62,7 @@ Ilya Skriblovsky <IlyaSkriblovsky@gmail.com>
 Iurii Gazin <archeg@gmail.com>
 Jānis Kiršteins <janis@janiskirsteins.org>
 jaykul
+Jens Krumsieck <j.krumsieck@outlook.de>
 Jeremy Koritzinsky
 Jeremie Magnette
 jezza323
--- a/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
+++ b/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
@@ -11,11 +11,11 @@ namespace OxyPlot.SkiaSharp
     using System;
     using System.Collections.Generic;
     using System.Linq;
-using System.Reflection;
+    using System.Reflection;
 
-/// <summary>
-/// Implements <see cref="IRenderContext" /> based on SkiaSharp.
-/// </summary>
+    /// <summary>
+    /// Implements <see cref="IRenderContext" /> based on SkiaSharp.
+    /// </summary>
     public class SkiaRenderContext : IRenderContext, IDisposable
     {
         private readonly Dictionary<FontDescriptor, SKShaper> shaperCache = new Dictionary<FontDescriptor, SKShaper>();
--- a/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
+++ b/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
@@ -11,16 +11,30 @@ namespace OxyPlot.SkiaSharp
     using System;
     using System.Collections.Generic;
     using System.Linq;
+using System.Reflection;
 
-    /// <summary>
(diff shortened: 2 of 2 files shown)


JensKrumsieck:
Regarding the CI Build error:

There is no `Assembly.GetEntryAssembly()`  in netstandard1.3, which is one of the TargetFrameworks of the SkiaSharp-Package. 
https://github.com/oxyplot/oxyplot/blob/e88c97fc1bc2c6da76ae45e2afd8531ee262ba89/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs#L873
The idea behind this line was to pack the Fonts into the UI Assemblies. 
To get rid of the Build error i see two possibilities:
1. Drop netstandard1.3 or increase to netstandard1.5 as `Assembly.GetEntryAssembly()` was added in the latter
2. Exclude this feature for netstandard1.3 with `#if NETSTANDARD1_5_OR_GREATER || NET452_OR_GREATER`

VisualMelon:
@JensKrumsieck thanks for that discussion; I'll try to take a proper look at this over the weekend

VisualMelon:
Sorry for taking so long to get back to this; I think excluding it for not-netstandard2.0 is the right thing to do (i.e. option 2).

If you're happy to make that change, then I'll approve this, and if none of the other maintainers object I'll merge a few days thereafter.

VisualMelon (approved):

VisualMelon:
Sorry for taking so long to get back to this.

PR #1864, opened 2022-03-02
Fix tracker label clipping 1863

### LINKED ISSUE #1863: WinForms ShowTracker Clipping PlotView Boundaries
### Expected behavior

I'd expect the ShowTracker function in OxyPlot.WindowsForms/PlotView.cs to offset the trackerLabel location if it is too close to the edge of the PlotView control containing it.  For example if the left most 30 pixels of the label would be outside the PlotView control and not visible, I'd expect the label to be offset 30 pixels to the right to compensate.  Vertically things get a little trickier due to potential overlap with the mouse cursor, but some cursor overlap is the lesser of two evils.

### Actual behavior

If the mouse is clicked/dragged to close to a PlotView control boundary the tracker label is clipped.

image

foxmja:
Fixes #1863.  First time contributing (to any open source project); documentation helped a lot, hopefully I didn't miss anything.

### Checklist

- [NA] I have included examples or tests

- Added some bounds checking to the WinForms PlotView trackerLabel control to keep it within the bounds of the parent control where it will remain visible.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file.
 - WPF DPI Regression (#1799)
 - Code generation for escape sequences in strings (#1824)
 - Axes not always honoring AbsoluteMinimum/AbsoluteMaximum and/or MinimumRange/MaximumRange properties (#1812)
+- WindowsForms tracker no longer clipping outside PlotView boundaries (#1863)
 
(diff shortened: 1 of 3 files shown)


VisualMelon (approved):
Looks good to me.

The WinForms tracker could do with a refit (default WPF tracker is somewhat clearer), but this is certainly an improvement on the current situation!

PR #1883, opened 2022-04-29
Improve StairStepSeries render performance

carlreinke:
### Checklist

1. Improve performance of `StairStepSeries`, particularly when zoomed in and X is monotonic or when consecutive points have equal Y components
2. `StairStepSeries` renders a horizontal line when a point with a valid X component and invalid Y component follows a valid point

Regarding 2, here is the behavior before the change:

And here is the behavior after:

Note that this does not change how `DataPoint.Undefined` is rendered:

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Series/StairStepSeries.cs
+++ b/Source/OxyPlot/Series/StairStepSeries.cs
@@ -156,47 +156,53 @@ namespace OxyPlot.Series
                                               : this.VerticalStrokeThickness;
 
             var actualColor = this.GetSelectableColor(this.ActualColor);
+            var edgeRenderingMode = this.EdgeRenderingMode.GetActual(EdgeRenderingMode.PreferSharpness);
 
-            Action<IList<ScreenPoint>, IList<ScreenPoint>> renderPoints = (lpts, mpts) =>
+            Action<IList<ScreenPoint>, IList<ScreenPoint>> renderPoints = (linePoints, markerPoints) =>
                 {
-                    // clip the line segments with the clipping rectangle
                     if (this.StrokeThickness > 0 && lineStyle != LineStyle.None)
                     {
+                        // TODO: Need ActualVerticalLineStyle; without it, when VerticalLineStyle is
+                        // Automatic, this code will not take the fast path even when it could.
                         if (!verticalStrokeThickness.Equals(this.StrokeThickness) || this.VerticalLineStyle != lineStyle)
                         {
-                            // TODO: change to array
-                            var hlpts = new List<ScreenPoint>();
-                            var vlpts = new List<ScreenPoint>();
-                            for (int i = 0; i + 2 < lpts.Count; i += 2)
+                            var hLinePoints = new List<ScreenPoint>(linePoints.Count);
+                            var vLinePoints = new List<ScreenPoint>(linePoints.Count);
+                            if (linePoints.Count >= 2)
                             {
-                                hlpts.Add(lpts[i]);
-                                hlpts.Add(lpts[i + 1]);
-                                vlpts.Add(lpts[i + 1]);
-                                vlpts.Add(lpts[i + 2]);
+                                hLinePoints.Add(linePoints[0]);
+                                hLinePoints.Add(linePoints[1]);
+                                for (int i = 1; i + 2 < linePoints.Count; i += 2)
+                                {
+                                    vLinePoints.Add(linePoints[i]);
+                                    vLinePoints.Add(linePoints[i + 1]);
+                                    hLinePoints.Add(linePoints[i + 1]);
+                                    hLinePoints.Add(linePoints[i + 2]);
+                                }
                             }
 
                             rc.DrawLineSegments(
-                                hlpts,
+                                hLinePoints,
                                 actualColor,
                                 this.StrokeThickness,
-                                this.EdgeRenderingMode.GetActual(EdgeRenderingMode.PreferSharpness),
+                                edgeRenderingMode,
                                 dashArray,
                                 this.LineJoin);
                             rc.DrawLineSegments(
-                                vlpts,
+                                vLinePoints,
                                 actualColor,
                                 verticalStrokeThickness,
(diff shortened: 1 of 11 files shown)


Source/OxyPlot/Series/StairStepSeries.cs
```diff
@@ -216,41 +222,146 @@ public override void Render(IRenderContext rc)
...
+                    {
+                        // Vertical line.
+                        linePoints.Add(Transform(previousPoint));
+
+                        // Horizontal line.
+                        linePoints.Add(Transform(previousPoint));
+
+                        xIncreased = false;
+                        xDecreased = false;
+                    }
```
VisualMelon: What's the idea here? Should this be projecting two different points rather than the previous point twice?

Does it need to clear both increased and decreased?

Source/OxyPlot/Series/StairStepSeries.cs
```diff
@@ -156,47 +156,53 @@ public override void Render(IRenderContext rc)
...
             var actualColor = this.GetSelectableColor(this.ActualColor);
+            var edgeRenderingMode = this.EdgeRenderingMode.GetActual(EdgeRenderingMode.PreferSharpness);
 
-            Action<IList<ScreenPoint>, IList<ScreenPoint>> renderPoints = (lpts, mpts) =>
+            Action<IList<ScreenPoint>, IList<ScreenPoint>> renderPoints = (linePoints, markerPoints) =>
                 {
-                    // clip the line segments with the clipping rectangle
                     if (this.StrokeThickness > 0 && lineStyle != LineStyle.None)
                     {
+                        // TODO: Need ActualVerticalLineStyle; without it, when VerticalLineStyle is
```
VisualMelon, line 165: What's this comment about? Is this an outstanding issue?

Source/OxyPlot/Series/StairStepSeries.cs
```diff
@@ -216,41 +222,146 @@ public override void Render(IRenderContext rc)
...
+                for (i = validOffset; i < endOffset; ++i)
                 {
-                    renderPoints(linePoints, markerPoints);
-                    linePoints.Clear();
-                    markerPoints.Clear();
-                    previousY = double.NaN;
-                    continue;
+                    var point = points[i];
+
+                    // For performance we want to draw as few lines as possible.  Assuming sane
```
VisualMelon, line 257: Might want to qualify 'sane axis transforms' as 'orthogonal and monotonic'

VisualMelon (changes requested):
Thanks for submitting this! The optimisation for monotone series is very welcome.

Could you add an example showing that the marker behaviour is unchanged when the lines are coalesced? (call it 'repeated values' or something)

The valid X/invalid Y is a bit weird, but it shouldn't mess with most logic, so probably OK if it's a useful scenario.

Source/OxyPlot/Series/StairStepSeries.cs
carlreinke: Yes, though not one that's been reported, as far as I'm aware.  It's just an additional opportunity for performance improvement that I wanted to call out, but one that I wasn't planning to tackle.

Source/OxyPlot/Series/StairStepSeries.cs
carlreinke: Yeah, so... I wasn't sure what kind of transformation it needs to be in order to make dropping points out of a horizontal line valid.  An affine transformation?  A linear transformation?  It's been a while since I studied linear algebra...

Source/OxyPlot/Series/StairStepSeries.cs
carlreinke: It's the same point twice because the number of points in the line has to be even to make the horizontal/vertical line styling work.

You're right -- there's a bug in here related to clearing both flags.  I'll fix it.

carlreinke: Fixed.

carlreinke:
> Could you add an example showing that the marker behaviour is unchanged when the lines are coalesced? (call it 'repeated values' or something)

I'm not sure it's a particularly interesting example -- looking at the loop over the points, you can see that the loop body will always reach the statement that adds the point to the markers (i.e. the loop has no `break` or `continue`).  I can add such an example, but I'm not sure how to make it interesting.  Actually, the "non-monotonic X" example is also not very interesting, but I hit enough snags that I kinda couldn't afford to not add something to show that it still works.  Not breaking marker rendering was easy, so I'm not sure adding an example for it is very useful.

> The valid X/invalid Y is a bit weird, but it shouldn't mess with most logic, so probably OK if it's a useful scenario.

I wanted to show a series of values, each of which is valid for a limited period of time.  If a value expires without a new value being provided, the horizontal line shows the time period that the value was valid (with no marker at the end since there isn't a value there).

Source/OxyPlot/Series/StairStepSeries.cs
VisualMelon: As far as I can tell, it just needs to be that a change in data-X effects not change in screen-Y (and vice versa: orthogonal) and that increasing the value in data-space only increases or decreases the value in screen space (can't decrease in one place and increase in the other: monotonic). The only axes we have which violate this are those for Polar plots, which have never been supported on `StairStepSeries`.

VisualMelon (approved):
Don't worry about making examples exciting: they don't have to be interesting. Apart from providing some compensation for the inadequate documentation, they also help to make sure we don't break things in future (e.g. we should show edge-cases in examples), and in that sense, they want to be as un-interesting as possible, showing exactly one thing. (It's common to have a sub-title that explains what you are meant to see, so that you can see immediately if there is an issue).

As usual, I'll give the other maintainers time to look at this before merging.

It would be nice to break out the addition of the editorconfig into a separate PR (thanks for adding that; I've been holding off adding the StyleCop analyser, so this will cover some of that), but I'm not fussed because it doesn't change the runtime behaviour.

VisualMelon:
Sorry for taking so long to get back to this.

PR #1884, opened 2022-04-29
Add plot manipulator axis preference

carlreinke:
### Checklist

- Add `AxisPreference` to `PlotManipulator`.  When set, if an action could apply to a preferred axis and also to another axis (such has an action performed in the plot area) then the manipulator operates on only the preferred axis.  If an action applies only to a non-preferred axis (such as an action performed in the axis area of a non-preferred axis) then the manipulator operates on that axis, as usual.

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/PlotController/Manipulators/PlotManipulator.cs
+++ b/Source/OxyPlot/PlotController/Manipulators/PlotManipulator.cs
@@ -34,6 +34,13 @@ namespace OxyPlot
         public IPlotView PlotView { get; private set; }
 
         /// <summary>
+        /// Gets or sets the axis that the manipulator will prefer to operate on.  The default is
+        /// <see cref="AxisPreference.None"/>.
+        /// </summary>
+        /// <value>The axis preference.</value>
+        public AxisPreference AxisPreference { get; set; }
+
+        /// <summary>
         /// Gets or sets the X axis.
         /// </summary>
         /// <value>The X axis.</value>
@@ -77,6 +84,21 @@ namespace OxyPlot
             if (this.PlotView.ActualModel != null)
             {
                 this.PlotView.ActualModel.GetAxesFromPoint(position, out xaxis, out yaxis);
+
+                if (this.AxisPreference != AxisPreference.None &&
+                    this.PlotView.ActualModel.PlotArea.Contains(position))
+                {
+                    if (this.AxisPreference == AxisPreference.X)
+                    {
+                        if (xaxis != null)
+                            yaxis = null;
+                    }
+                    else if (this.AxisPreference == AxisPreference.Y)
+                    {
+                        if (yaxis != null)
+                            xaxis = null;
+                    }
+                }
             }
             else
             {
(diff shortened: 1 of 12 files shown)


Source/OxyPlot/PlotController/Manipulators/PlotManipulator.cs
```diff
@@ -77,6 +84,21 @@ protected void AssignAxes(ScreenPoint position)
             if (this.PlotView.ActualModel != null)
             {
                 this.PlotView.ActualModel.GetAxesFromPoint(position, out xaxis, out yaxis);
+
+                if (AxisPreference != AxisPreference.None &&
```
VisualMelon: Member access should use `this`: if you can write a style-cop pass that would be great.

VisualMelon:
My only reservation here, is that it would be nice if the plot manipulators were not tied to a single X and Y axis; it's common e.g. to want to pan _all_ axes, which have been carefully lined up at the outset (probably keeping the main X and Y axes, which in practise are useful even if you want to operate on many, and would help with backwards compatibility).

_If_ there were a change in future to support this, I would imagine adding `All` to your `enum`, and then populating a list of axes instead. `None` would, then, sound rather like _no axes_ rather than _no preference_.

Could we change `None` to `Default`?

Source/OxyPlot/PlotController/Manipulators/PlotManipulator.cs
carlreinke: I added a rule in .editorconfig for this.

carlreinke:
If plot manipulators weren't tied to a single X and/or Y axis then you'd have to make a breaking change in `PlotManipulator<T>` since it exposes `XAxis` and `YAxis` properties.  If you made that breaking change, redesigning how the axis preference works would be part of that.  (Though I think that's out of scope for this PR.)

I did consider `Default` and also `All`.

The problem with `Default` is that it doesn't say much about what the behavior will be -- you get the default preference, but what is the default preference?  To a developer who is familiar with OxyPlot, it's probably clear that you get the behavior that you're used to.  To a new developer, it's maybe not so clear.

The problem with `All` is that it's not really a preference.  (Compare: `All` "The manipulator prefers all of the axes." vs `None` "The manipulator doesn't prefer any one axis over the others.").  Then again, it could be thought of as "I'd prefer that all the axes were manipulated." vs "I'd prefer that none of the axes were manipulated."  But clearly manipulating none of the axes isn't useful, so you'd probably dismiss that interpretation.

So I ended up with `None`.

I originally started off with the name `AxisBias` rather than `AxisPreference`.  In that case the default could be named `Unbiased`.

Maybe there is a better name that hasn't been considered yet?  Let me know what you prefer.

VisualMelon (approved):
Regarding the XAxis and YAxis properties, yes, they'd need to stay for compatibility, and in practise this is useful when implementing manipulators that operate on multiple axes, so I'm not worried about that. (also agree that's not an issue for today)

I agree `Default` is pretty meaningless, but I think the name is always going to be pretty meaningless, because it could cover any type of default axis (e.g. axes for Polar plots): the current code just sort of grabs whatever it can (as you say, this might mean something to a familiar user, but not a newcomer).

I think the problems with the names is we're running by slightly different interpretations of 'preference': `None` can be read as "I'd prefer no axes". That said, I can't think of a better name offhand, so approving now; perhaps another maintainer will have a better idea, or something will come to us in the next few days before merging.

Note: you may need to rebase this or the other branch when one is merged, but hopefully not.

Issue #39, opened 2014-08-14
Complete XAML support


objorke:
Add XAML support for all types of annotations, axes and series for 
- [ ] WPF
- [ ] UWP
- [ ] Xamarin Forms
- [ ] Avalonia

Include all properties. 

Create a tool that generates this code automatically.

PR #1902, opened 2022-07-07
refactor: Refactored to use dependency-property-generator.

HavenDV:
Fixes # .

### Checklist

- Switching to DependencyPropertyGenerator(according changes in H.OxyPlot - https://github.com/HavenDV/H.OxyPlot/commit/486ef866508b64611738518736db8f01a5132de3)
-
-

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 ### Changed
 - Update net40 and net45 to net452 (#1835)
 - Change default `MinimumSegmentLength` to `2` and remove limits for series and annotations with simple geometry (#1853)
+- Switching to DependencyPropertyGenerator (#1902)
 
 ### Removed
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -143,3 +143,4 @@ LauXjpn <laucomm@gmail.com>
 R. Usamentiaga
 Dmytro Shaurin
 Rustam Sayfutdinov
+Konstantin Stukov <havendv@gmail.com>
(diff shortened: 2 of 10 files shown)


HavenDV:
You can verify that the generated code is equivalent by opening:
image

VisualMelon:
I'm not sure if this is a good idea: these generators are great for RAD and keeping things clean, but here we are losing the comments (though none of them in this case are all that inspiring, and indeed some are wrong) and it all seems a bit much for not much simplification in this case (most of these files have had about 3 changes since they were laid down in 2014; OxyPlot.Contrib would be a different matter).

This is not to say I'm opposed to this change, but I probably shan't merge it by myself because it adds additional third-party dependencies.

HavenDV:
I agree, there are pros and cons here, so feel free to close this PR if it doesn't seem right for this library.
xml documentation can also be added via this generator, either via the Description property (which will also generate an attribute for the designer) or raw. I didn't add this because it seemed to me that there was documentation here that didn't say anything. But I missed the point that there is some documentation here that makes sense.

HavenDV:
I've updated the PR to be fully consistent with the current documentation:
For DependencyProperties "Identifies" documentation is generated automatically.
For regular properties, the value is taken from the Description attribute.

VisualMelon:
Having the comments is a big plus for me. I'm certainly not going to close this, as one of the other maintainers may have a more positive reactive and I don't want to bury it, but as it does add 'third party' dependencies I'd want to defer to @objorke .

objorke:
Thanks @HavenDV for creating the PR, but I am also negative to merging this PR.

Dependencies should be avoided, both for maintenance and security perspectives. I see how dependency-property-generator can simplify code, but I don't think it adds enough value to bring in a dependency, even if it is a developmentDependency.

See also #39, I think that issue is similar - syncing dependency properties to the "core" model.

HavenDV:
Thank you for reviewing this PR. It's ok that it's rejected, your reasons are quite understandable.
P.S. Regarding the specified issue - the generator just made the dependency property code "cross-platform" (more precisely, it generated its own unique code for each platform, but keeping the same behavior between them)

PR #1903, opened 2022-07-07
refactor: Added nullable annotations to Utilities classes of Core library

HavenDV:
I want to start gradually adding nullable annotations in small chunks so that it's easier to test and accept a PR. It's just markup, I'm not changing the behavior.
In this PR I added 
```cs
        protected XmlWriterBase()
        {
            this.w = XmlWriter.Create(new StringBuilder());
        }
```
because all methods of this class do not check w for null. Here were alternatives to use ! to ignore or add null checks to each of the methods.
I will provide the next PR after accepting this. I don't list myself as a contributor or as a changelist, because that's only part of the change.

Fixes # .

### Checklist

- Added nullable annotations to Utilities classes of Core library
-
-

@oxyplot/admins


VisualMelon:
We had some discussion about this a while back: I think this approach of starting with the ancillary parts is a good idea.

VisualMelon:
Re. `XmlWriterBase..ctor()`, I think we should just remove it, or replace it with a parameter that takes an `XmlWriter`. I think there is no way to actually get at the contents if it is initialised with a new string builder.

In `ReflectedPath`, both arrays (`infos` and `reflectedTypes`) should probably have nullable elements and be initialised as such in the ctor.

Otherwise, looks good to me.

HavenDV:
By the way, you may not know - github allows you to comment on PR right next to the code:
image

HavenDV:
About infos: Then here comes the error because GetRuntimeProperty can return null
image

VisualMelon:
Re inline comments: I did... but I also forgot, so thanks for the reminder ;)

Re. the arrays, I probably wasn't clear: I think they should both have nullable elements, i.e. `infos` should be as it is in your current, `reflectedTypes` should be `Type?[]`, and the initialisation in the ctor should be updated to reflect this (bit surprised/annoyed it doesn't error on the assignment with non-nullable, TBH, but arrays are always going to be a sticking point for nullable, and I guess this is a gap for the sake of dev sanity more than anything else).

HavenDV:
I agree, this is very strange, ideally we should create an issue about this in one of the dotnet repositories (or search, it probably already exists)

VisualMelon (changes requested):
Can you add yourself to CONTRIBUTERS.txt in this PR, and then I'll get this merged.

VisualMelon (approved):

HavenDV:
https://github.com/dotnet/csharplang/issues/6368#issuecomment-1214180877

VisualMelon:
Yeah, Arrays are always going to be weird. No big deal.

PR #1913, opened 2022-08-13
Add PathAnnotation.TextBackground properties (#1900)

### LINKED ISSUE #1900: `Background` property is missing on `PathAnnotation`s and `Stoke` properties are ignored
### Steps to reproduce

1. Make a `LineAnnotation` with some text.
2. Set `StrokeThickness` to a non-zero value and `Stroke` to an `OxyColor`.
3. Observe that you can't set a `Background` like with `TextAnnotation`
4. At runtime, observe that the `Stoke` and `StrokeThickness` properties are ignored.

Platform: Windows 11
.NET version: 6.0.6
OxyPlot version: 2.1.0+ custom-built from e3c42f72

### Expected behaviour

I can set a `Background` property on my `LineAnnotation` and the `Stoke` and `StrokeThickness` properties aren't ignored.

### Actual behaviour

I can't set a `Background`, and the stroke ones are ignored at runtime (no border box is drawn)

VisualMelon:
Fixes #1900 by adding properties to `PathAnnotation` that match the functionality of `TextAnnotation`. Decided to not try to put everything in `TextualAnnotation` because there would be naming issues, as the same names are used in `PathAnnotation` and `TextAnnotation` already for different purposes.

### Checklist

- Add four new properties to `PathAnnotation`:
  - `TextBackground`
  - `TextBackgroundStroke`
  - `TextBackgroundStrokeThickness`
  - `TextBackgroundPadding`

@colejonson66 suggested `Border` for the stroke properties as an alternative to `Background`, which may be a better name.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file.
 - Example of issue with AreaSeries tracker (#1982)
 - Example for CategoryAxis with custom MajorStep and uncentered ticks (#1971)
 - BarSeries.LabelAngle property (#1870)
+- Border properties on PathAnnotation to match functionality in TextAnnotation (#1900)
 
 ### Changed
 - Make consistent BaseValue and BaseLine across BarSeries, LinearBarSeries, and HistogramSeries
@@ -42,7 +43,6 @@ All notable changes to this project will be documented in this file.
 - Add `AxisPreference` to `PlotManipulator`
 - Add MinimumMajorIntervalCount and MaximumMajorIntervalCount Axis Properties (#24)
 - Add VisualStudioToolsManifest.xml to add components to the Visual Studio Designer toolbox (#1446)
-- Add TextBorder properties on PathAnnotation to match functionality in TextAnnotation (#1900)
 
 ### Changed
 - Change default `MinimumSegmentLength` to `2` and remove limits for series and annotations with simple geometry (#1853)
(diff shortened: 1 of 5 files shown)


VisualMelon:
`ShapeAnnotation` would be another candidate for adding such properties if we can settle on names.

VisualMelon:
@oxyplot/admins does this API look OK? Not wild about `ChangeOpacity` now that I look at it again, but can remove that and keep the other changes otherwise.

Any preference between `Border`/`Background` for naming per ermakrs from @colejonson66 ?

Jonarw:
`Border` sounds more adequate at first, however the property `TextBackground` doesn't make sense when renamed to `TextBorder`. I guess we'd have to do `TextBorderBackground` then. `TextBorderStroke`, `TextBorderStrokeThickness` and `TextBorderPadding` would be fine however.

Currently I would tend to leaving it at `Background`, but could be convinced otherwise.

`ChangeOpacity` is fine in my opinion, `ChangeAlpha` could also work.

VisualMelon:
Looking at this quickly now, I'm tempted again by `TextBorderStroke` and `TextBorderThickness`, leaving everything else as is. If that's agreeable, then I'll make that change and I think we should get this merged.

Jonarw:
Sounds good to me!

VisualMelon:
Then I don't like `TextBackgroundPadding`... I think `TextBorderBackground` is OK, so I've changed to `Border` everything (rather than `TextBackground` and rebased. I don't know if you're convinced by the new naming, however.

Jonarw (approved):
All fine with me, if you're happy I think we can merge this :)

PR #1916, opened 2022-08-15
fix: Fixed binding and DataContext issue in WPF ExampleBrowser.

HavenDV:
Fixes # .

### Checklist

- I moved the initialization of the ViewModel into a xaml file, so the IDE's xaml parser is able to understand the DataContext type and issue warnings about incorrect bindings
- Accordingly, I corrected incorrect bindings. I have not checked other applications for this error.
- Sorry for the xaml formatting change, this is due to https://github.com/Xavalon/XamlStyler, I can submit a PR that will provide a configuration (https://github.com/Xavalon/XamlStyler/wiki/External-Configurations) for it according to your requirements.

@oxyplot/admins

### DIFF
--- a/Source/Examples/WPF/ExampleBrowser/MainWindow.xaml
+++ b/Source/Examples/WPF/ExampleBrowser/MainWindow.xaml
@@ -1,18 +1,30 @@
-﻿<Window x:Class="ExampleBrowser.MainWindow"
-        xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
-        xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-        xmlns:oxyWpf="clr-namespace:OxyPlot.Wpf;assembly=OxyPlot.Wpf" 
-        xmlns:oxySkia="clr-namespace:OxyPlot.SkiaSharp.Wpf;assembly=OxyPlot.SkiaSharp.Wpf" 
-        xmlns:local="clr-namespace:ExampleBrowser" 
-        Title="OxyPlot.WPF Example Browser"
-        Height="720" Width="1280" Icon="OxyPlot_64.png">
+﻿<Window
+    x:Class="ExampleBrowser.MainWindow"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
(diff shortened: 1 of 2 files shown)


VisualMelon:
(Comprehensive?) Version without formatting changes (for reference): https://github.com/oxyplot/oxyplot/compare/develop...VisualMelon:oxyplot:wpfexamplebrowserbinding

I don't think the current code conforms to any particular standard at present beyond tabbing to 4 spaces. I've not done commercial XAML for years, so not sure what the standards people tend to go with are these days if we wanted to standardise on anything. Personally I'm not wild about limits on attributes-per-line and never liked omitting the space before `/>`, but there is a mix of everything in the sources.

HavenDV:
Thanks for making a version without xaml changes. I think I should close this PR in favor of yours.

HavenDV:
I use XamlStyler mainly for the sake of uniformity and styling automation. It just saves you time because it works when you save with Ctrl + S

VisualMelon:
OK, I'll merge my version now; not sure if we can expect contributors to install a third-party extension, but may be worth defining a standard in any case.

HavenDV:
It is optional. Here is the style I usually use in my projects: https://github.com/HavenDV/H.OxyPlot/blob/main/settings.xamlstyler
XamlStyler will use it if it is present in the solution folder (or in the root folder of the repository, but I'm not sure because I always store .sln files in the root of the repository)
If contributors don't have this extension, they won't be affected in any way by having this file, except that their new xaml code might not follow certain rules.
Also, perhaps, after adding this file, it makes sense to format all files of the repository using a console utility.

VisualMelon:
Yeah, that makes sense. Adding such rules along with a PR that formats core (non-example) files accordingly would seem a reasonable thing to do. I personally don't like rules like "one attribute per line" (though I don't see a way to disable that in XamlStyle) or ordering things by name automatically (as a compulsive saver, that would drive me crazy), but nothing else looks objectionable in this PR at least, and I'm assuming this is from the same format you linked?

HavenDV:
The PR had default settings from here: https://github.com/Xavalon/XamlStyler/wiki/External-Configurations
My file disables attribute reordering and also wraps `/>` on a new line.
As I understand it, you need to change the following settings to achieve what you want:
```
"MaxAttributesPerLine": 0,
"EnableAttributeReordering": false,
```

PR #1925, opened 2022-08-25
Replaced unsafe value < double.Epsilon expressions

### LINKED ISSUE #1924: 0 < double.Epsilon checks return false
Several expressions in OxyPlot make a zero value check using the idiom: `value < double.Epsilon`.
For example the range check in CalculateActualInterval in Axis.cs

The problem is that `0 < double.Epsilon` is not a safe comparison in .NET. It may evaluate to false, if double.Epsilon is expressed as 0 by the compiler.
The consequence in Axis.CalculateActualInterval is that it gets stuck in an infinite while(true) loop later in the method.

```
            // This is the actual code. It does not throw if range == 0
            if (Math.Abs(range) < double.Epsilon)
            {
                throw new ArgumentException("Range cannot be zero.", "range");
            }

            // This does not throw either
            if (range < double.Epsilon && range > -double.Epsilon)
            {
                throw new ArgumentException("Range cannot be zero.", "range");
            }

               // This throws!!!
            if (range == 0)
            {
                throw new ArgumentException("Range cannot be zero.", "range");
            }
```

The issue is observed with Visual Studio Pro 2022 17.2.6 using an x64 build, target .NET 4.8.

Setting ActualMaximum to ActualMinimum and forcing a CalculateActualInterval will reproduce the issue.
ActualMaximum / ActualMinimum needs to be set indirectly, not quite sure how. Perhaps Zoom(0) will do the trick.

Suggestion:
Change all `value < double.Epsilon` to simply `value == 0` or at least `value <= double.Epsilon`

The `value < double.Epsilon` idiom occurs several places in the codebase.

### Steps to reproduce

1. Setting ActualMaximum to ActualMinimum and forcing a CalculateActualInterval will reproduce the issue.
ActualMaximum / ActualMinimum needs to be set indirectly, not quite sure how. Perhaps Zoom(0) will do the trick.

Platform:
.NET version: 4.8

### Expected behaviour

Exception thrown for range == 0

### Actual behaviour

Oxyplot gets stuck in infinite loop.

pantosh:
Fixes #1924.

All `Math.Abs(value) < double.Epsilon` checks replaced with `<=` since `0 < double.Epsilon` cannot be trusted to evaluate true on all architectures.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file.
 - WindowsForms tracker no longer clipping outside PlotView boundaries (#1863)
 - Histogram now rendering properly when using logarithmic Y axis (#740) 
 - Fix ExampleLibrary build errors in certain code pages (#1890)
+- Fix for double.Epsilon zero check that fails on some architectures (#1924)
 
 ## [2.1.0] - 2021-10-02
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -120,6 +120,7 @@ Scott W Harden <swharden@gmail.com>
 Shankar Mathiah Nanjundan <shankar.ooty@hotmail.com>
 Shun-ichi Goto <shunichi.goto@gmail.com>
 Soarc <gor.rustamyan@gmail.com>
+Soren Holstebroe <sh@pantoinspect.com>
 Stefan Rado <oxyplot@sradonia.net>
 stefan-schweiger
 Steve Hoelzer <shoelzer@gmail.com>
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -1449,7 +1449,7 @@ namespace OxyPlot.Axes
                 a1 -= this.MaximumDataMargin * marginSign;
             }
 
-            if (this.ActualMaximum - this.ActualMinimum < double.Epsilon)
+            if (this.ActualMaximum - this.ActualMinimum <= double.Epsilon)
             {
                 this.ActualMaximum = this.ActualMinimum + 1;
             }
@@ -1740,7 +1740,7 @@ namespace OxyPlot.Axes
             var actualMaximum = this.DataMaximum;
             double range = this.DataMaximum - this.DataMinimum;
 
-            if (range < double.Epsilon)
+            if (range <= double.Epsilon)
             {
                 double zeroRange = this.DataMaximum > 0 ? this.DataMaximum : 1;
                 actualMaximum += zeroRange * 0.5;
@@ -1769,7 +1769,7 @@ namespace OxyPlot.Axes
             var actualMinimum = this.DataMinimum;
             double range = this.DataMaximum - this.DataMinimum;
 
-            if (range < double.Epsilon)
+            if (range <= double.Epsilon)
             {
                 double zeroRange = this.DataMaximum > 0 ? this.DataMaximum : 1;
                 actualMinimum -= zeroRange * 0.5;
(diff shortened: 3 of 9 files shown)


VisualMelon (approved):

VisualMelon:
Azure failure can be ignored

VisualMelon:
I'll leave this for the moment to let the other maintainers take a look, but if there are no objections I'll merged it after a few days.

pantosh:
I tried to minimize the double.Epsilon changes to those that would surely fail on archs where 0 < double.Epsilon is false, however it should be considered to remove all usages of double.Epsilon. See the discussion in #1924.

The largest breaking risk is in the HistogramSeries where it looks like double.Epsilon is used as some non-zero marker for min/max. That is risky on its own and should be changed, since it may evaluate to zero.

VisualMelon:
For Historgram series, we probably need to detect the log(0) conditions explicitly and deal with it. Notes in #1886

Jonarw:
Thanks for highlighting this issue! This is something I absolutely was not aware of.
Considering that there are seemingly no guarantees regarding `double.Epsilon`, I'd be inclined to eliminate all its usages across the code base. In most cases it should be fine to just do `<= 0`, right? Except for the `HistogramSeries`, which we would have to consider separately, this shouldn't really break anything, right? And even if it does break something, at least it would break _consistently_, which I think would be better than breaking _sometimes_.

VisualMelon:
I'd agree with @Jonarw that `<= 0` would be the most direct translation, and looking again it's less likely to create confusion down the line.

(note: rebasing should stop azure whinging)

pantosh:
The changes in this PR were the most conservative I could make, i.e. replacing < epsilon with <= epsilon.
There are a handful of other epsilon usages that I didn't touch, because of the risk of double.Epsilon having some semantically meaning as a non-zero marker (i.e. in histogram).

In conclusion, double.Epsilon is not a safe constant to use. For predicates that check if two calculations are mathematically the same "within floating point rounding error", some other epsilon must be used, i.e. 1e-50. Obviously this has its own problem, if the user happens to use data at that scale. Floats are not always easy to work with in a general purpose library.

VisualMelon:
I think what @Jonarw and I are saying is that substituting `<= 0` for `< double.Epsilon` will always produce the same behaviour as when epsilon is the smallest positive representable value (per the docs and my previous understanding), and I think it will be less likely to confuse matters in future: `<= epsilon` looks - as you point out - like a conscious (but failed) attempt to cleanly handle edge cases and won't be consistent across platforms.

I think the only truly problematic instance in the codebase is histogram (i.e. where just fixing it in this PR and forgetting about it won't cut it): in the other cases, there is no obvious epsilon to use, and picking one would just provide borderline-wrong behaviour for borderline data, so reducing them to `<= 0` and losing the mention of `Epsilon` (which could serve as a marker) is absolutely fine. If issues do emerge then we can address those directly down the line, and I think ensuring the same behaviour (as expected and across platforms) and keeping the code simple is the right way to go here.

VisualMelon:
(all that said, consistency probably isn't a big deal (not something we can solve), given the platforms will be inherently inconsistent anyway, and in the specific case of epsilon comparison, `<= 0` and `<= epsilon` will be the same, if we ignore all the other subnormals)

VisualMelon:
@pantosh sorry about my indecisiveness on this; if you can change the instances of `<= epsilon` to `<= 0`, then I'll get this quickly merged.

As an aside, I had a bash at the histogram issue at some point but wasn't happy with it; will have another go once this and some of the modernisation PRs have gone through.

VisualMelon:
@pantosh are you happy to make the change described by @Jonarw and myself above? (that is, change `<= epsilon` to `<= 0`).

If you don't have time, let me know, and I'll pick it up.

VisualMelon:
@pantosh Work merged in #1997 : thanks for reporting this and suggesting the fix; wanted to get on with the merge to made the final changes discussed in a separate branch.

PR #1927, opened 2022-08-25
DataPointSeries.GetItem is using cached items source points.

### LINKED ISSUE #1926: DatePointSeries ignores cached item source points
The GetItem method in DataPointSeries never uses cached item source points which can result in a huge performance penalty for simple operations like getting nearest point for mouse movement.

### Steps to reproduce

1. Create an item source class that implements IEnumerable<DataPoint>.
2. Make a log entry or breakpoint in the enumerator.
3. Attach the item source to a line series and move the mouse arround.

Platform:
.NET version:

### Expected behaviour
Mouse movement should only result in a single iteration of the items source.

### Actual behaviour
The item source iterator is called quite a lot.

pantosh:
Removed items source null check in DataPointSeries.GetItem since it prevented the use of cached items source points resulting in a huge performance penalty for large or slow items source data sets.

Fixes #1926 .

Null check for items source removed since it is already checked for in ActualPoints, but ActualPoints also returns cached points if available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file.
 - WindowsForms tracker no longer clipping outside PlotView boundaries (#1863)
 - Histogram now rendering properly when using logarithmic Y axis (#740) 
 - Fix ExampleLibrary build errors in certain code pages (#1890)
+- DataPointSeries.GetItem uses cached items source points if available.
 
 ## [2.1.0] - 2021-10-02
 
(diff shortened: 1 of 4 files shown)


VisualMelon:
This is a tad non-trivial, so I'll wait to test it until I have more time.

pantosh:
The items source invalidation design lacks some sophistication as also observed in #1928.
It would be more intuitive if caching of items sources were invalidated on the items source level rather than on the PlotModel.
Caching of items should be separated from their use instead of all the if itemssource == null checks, which do seem to cause crashes and performance issues.

pantosh:
This change did have some negative consequences: Custom labels referred with TrackerFormatString did not show.
This means that this is not pull ready yet.
I still think there are some serious performance issues with data sources containing many plots.

Jonarw:
Closing this according to the discussion in https://github.com/oxyplot/oxyplot/discussions/2012. If there is interest to continue with this topic, feel free to let us know in here or start a new PR.

PR #1935, opened 2022-09-13
refactor: Added nullable annotations to Annotations classes of Core library.

HavenDV:
### Checklist

-
-
-

@oxyplot/admins


VisualMelon:
Sorry I've not got around to looking at this yet, I'll try to find time tomorrow to go through it properly.

Source/OxyPlot/Annotations/FunctionAnnotation.cs
```diff
@@ -36,7 +38,7 @@ public FunctionAnnotation()
         /// <summary>
         /// Gets or sets the y=f(x) equation when Type is Equation.
         /// </summary>
-        public Func<double, double> Equation { get; set; }
+        public Func<double, double>? Equation { get; set; }
```
VisualMelon, line 41: This is the sort of thing where I got unstuck when I tried to fit nullables: things like this and `ImageAnnotation.ImageSource` would ideally be `required` and non-nullable, but that would present a breaking change, forcing assignment up-front. I think we have to make them nullable, but ideally, we'd also signal that they must be assigned for the object to function. Long term, we might judge that it makes sense to make these required and non-nullable, but the whole class of incrementally fillable types is problematic.

For the time being, however, these have to be nullable to avoid a breaking change, because there are no sane defaults, and defaulting to `null!` would just create confusion.

Source/OxyPlot/Annotations/ImageAnnotation.cs
```diff
@@ -120,17 +122,29 @@ public override void Render(IRenderContext rc)
 
             if (double.IsNaN(width) && double.IsNaN(height))
             {
+                if (this.ImageSource == null)
+                {
+                    throw new InvalidOperationException($"{nameof(this.ImageSource)} is null.");
+                }
```
VisualMelon: Can we move these up to the top of the method and make the exception message more descriptive? (that said, not sure what to suggest, something like "ImageSource must be non-null before rendering" would just be confusing)

Source/OxyPlot/Annotations/LineAnnotation.cs
```diff
@@ -96,7 +98,7 @@ protected override IList<ScreenPoint> GetScreenPoints()
                     points.Add(new DataPoint(this.ActualMinimumX, fx(this.ActualMinimumX)));
                     points.Add(new DataPoint(this.ActualMaximumX, fx(this.ActualMaximumX)));
                 }
-                else
+                else if (fy != null)
```
VisualMelon, line 101: Probably ought to throw if both are none, to catch any internal implementation errors.

Source/OxyPlot/Annotations/PathAnnotation.cs
```diff
@@ -182,6 +184,14 @@ public override void Render(IRenderContext rc)
 
             this.GetActualTextAlignment(out var ha, out var va);
 
+            if (this.XAxis == null)
+            {
+                throw new InvalidOperationException($"{nameof(this.XAxis)} is null.");
+            }
+            if (this.YAxis == null)
+            {
+                throw new InvalidOperationException($"{nameof(this.XAxis)} is null.");
```
VisualMelon: Typo: YAxis

Source/OxyPlot/Annotations/PathAnnotation.cs
```diff
@@ -291,6 +301,15 @@ protected override HitTestResult HitTestOverride(HitTestArguments args)
         /// </summary>
         protected virtual void CalculateActualMinimumsMaximums()
         {
+            if (this.XAxis == null)
+            {
+                throw new InvalidOperationException($"{nameof(this.XAxis)} is null.");
+            }
+            if (this.YAxis == null)
+            {
+                throw new InvalidOperationException($"{nameof(this.XAxis)} is null.");
```
VisualMelon: Typo: YAxis

Source/OxyPlot/Annotations/TextualAnnotation.cs
```diff
@@ -31,7 +33,7 @@ protected TextualAnnotation()
         /// Gets or sets the annotation text.
         /// </summary>
         /// <value>The text.</value>
-        public string Text { get; set; }
+        public string? Text { get; set; }
```
VisualMelon, line 36: File will need further changes when `MathRenderingExtensions` is updated: might be worth doing that now?

VisualMelon (changes requested):

VisualMelon:
Main thing to work out, I think, is how we are going to signal properties that have no default but must be assigned, e.g. `FunctionAnnotation.Equation` (see above) and `ImageAnnotation.Source`: both in the documentation, and when we throw exceptions because they are not assigned by the time they are rendered etc. Any thoughts welcome: worth figuring this out now then we can apply it elsewhere.

Pinging @Jonarw because he reviewed my original attempts at this.

Source/OxyPlot/Annotations/FunctionAnnotation.cs
HavenDV: I want to clarify that I am not changing anything. Right now the default is null, and nullable enable just exposes that.
But I can set the default value as `value => value` lambda.

VisualMelon: @HavenDV sorry, serious typo in my message, should have read "For the time being, however, these have to be nullable to avoid a breaking change"

I think any non-null default will just cover over issues in consumer code.

HavenDV:
For now, I've just settled on simply throwing `InvalidOperationException("{name} is null.")`. This is not the best solution, but it is much better than the possible NRE. I didn't bother to add a detailed description because the current code is working and well tested and these events most likely won't happen. In addition, I do not always understand the context of the code in order to make a clear description.

Source/OxyPlot/Annotations/ImageAnnotation.cs
HavenDV: image
Please check my post below.
Also, I didn't add an exception to the beginning of the method because the ImageSource is not needed if GetVector returns non-NaN X and Y values. But I didn't look below where the RenderingExtensions.DrawImage call happens, which requires a non-null ImageSource value and which hasn't yet been marked with nullable enable

VisualMelon:
Context I'm imagining is a user who is assembling an object from different places (so can't just throw all the properties together at once without needing a builder or shadow type of some kind) who may fail to set the value of one of these required-but-with-no-default properties: they may either not realise that they are meant to assign the value, or it may not be assigning it without realising (or indeed may be assinging it null without realising). As you say, hopefully OxyPlot itself isn't going to assign e.g. `FunctionAnnotation.Function` to null unilaterally, so ideally we want an exception that makes it clear that it's user error. `InvalidOperationException` is the right type (if we don't want our own), but I just wonder if we can convey it more strongly in the message: this property is null, and shouldn't be null by the time the plot element is expected to do something (and it's your problem as the consumer). Like I said, I can't suggest a better wording off hand, but hopefully that explains where I'm coming from.

Source/OxyPlot/Annotations/ImageAnnotation.cs
VisualMelon: Sorry, I didn't realise that none of the render context stuff has been done yet. The image source is essential to the image annotation, and should be required for the call to `RenderingExtensions.DrawImage`, so it's OK to throw under any circumstance. Maybe like `MathRenderingExtensions` it makes sense to annotate `RenderingExtensions` and `IRenderContext` now (should be pretty simple, since there's no state lingering).

Source/OxyPlot/Rendering/RenderContext/RenderingExtensions.cs
```diff
@@ -93,8 +95,8 @@ public static EdgeRenderingMode GetActual(this EdgeRenderingMode edgeRenderingMo
             EdgeRenderingMode edgeRenderingMode,
             double[] dashArray,
```
VisualMelon: `dashArray` should be nullable (`GetDashArray` in `LineStyleHelper` can return `null`)

Source/OxyPlot/Annotations/TextualAnnotation.cs
```diff
@@ -33,7 +33,7 @@ protected TextualAnnotation()
         /// Gets or sets the annotation text.
         /// </summary>
         /// <value>The text.</value>
-        public string? Text { get; set; }
+        public string Text { get; set; } = string.Empty;
```
VisualMelon: This should remain as `string? Text` initialised to `null`

(Sorry, my original comment wasn't very clear: with it nullable we need to add checks when calling into `MathRenderingExtensions`)

VisualMelon (changes requested):
Couple of small things, otherwise I think this is all good.

Source/OxyPlot/Annotations/TextualAnnotation.cs
HavenDV: This will entail a whole series of checks during the rendering stage.
image

VisualMelon: I think that might be inevitable. I'll take a closer look now, but I don't think we can change the default to non-null in a hurry.

HavenDV: I only see problems with this if users with Nullable disable can sometimes have null instead of this value. But this will result in an NRE for them anyway.
While exceptions will let the user know that they forgot to set the Text property, the very use of TextualAnnotation implies this.

VisualMelon: It's the same problem as `ImageAnnotation.ImageSource` etc.: it has to be set, but there's almost 10 years of code that might not be initialising the object all at once, and may even be using `null` to signal to different parts of code that it needs to be changed: changing the default at this point will cover-over failing to assign the value, and be a silent code-breaking change for some users. Perhaps we can try to gauge how big a deal this will be?

Regarding the last point, `PointAnnotation` and friends don't actually require the text property to be set; however, the behaviour won't change for `PointAnnotation` with your proposal, but it may change the behaviour of consuming code. It may just be the old versions of everything we're using, but I'm surprised by the error in that case, as the usage is within an `if (!String.IsNullOrEmpty)` check.

Looking back at old posts, I think my previous remarks express my current feelings rather well:

> I think the main benefit we might see from this will just be in informing us where we should be throwing more helpful errors. Short of breaking every plot element in two, I don't think there is too much we can achieve to tighten the API, so it will mostly be limited to expressing which things may be null, which dilutes the meaning for those which are expected to be null.

The problem is that OxyPlot is basically a bundle of mutable objects without defaults (which means that `null` signals something to the consumer). This (`TextualAnnotation.Text`) is obviously a case where I'm advocating for making something nullable because it might happen to be null, rather than because it is intended to be null.

@Jonarw has previously described the possibility of just leaving nullable off for certain types; I'm going to do some reading/experimentation to see if this might be a possibility in these cases. I've previously been opposed to this, but I could certainly buy the benefit of having nullable annotations on all methods but omitting it on a class of properties if it's not too misleading for the consumer. _In particular, we could leave these unannotated now, and then annotate them e.g. when we have `required`: benefit of `required` is that it expresses the intention well, means simple and correct code will continue to work fine, and will noisily break old code, so is OK for a major release._

I'm torn between respecting the type system - which tells me to keep these nullable and use documentation to talk to the user instead - and trying to express the intention that these should be initialised to a non-null value by the user - which tells me to make these non-null and just initialise them to `null` anyway (is there any precedent for this in other libs?). I'm definitely opposed to adding non-null defaults for everything at the moment, though in this specific case, perhaps `string.Empty` would be OK..

Sorry, no decisions in there... I'd like to take a bit more time to think about this. My instinct still is to keep them nullable, and add more useful exceptions and inline documentation. Feel free to dump any vaguely relevant thoughts you might have here.

VisualMelon:
@HavenDV thanks for pushing that (if nothing else makes it easier to talk about): what's your reckoning? (`NotNullWhen` appears in netstandard2.1)

HavenDV:
> @HavenDV thanks for pushing that (if nothing else makes it easier to talk about): what's your reckoning? (`NotNullWhen` appears in netstandard2.1)

Sorry, I didn't understand what you mean, English is not my native language

VisualMelon:
Sorry, I'm not sure that's even good English to be honest.

Just wanted to know your opinion on your latest commit; obviously it's not pretty, but I don't think we can do better while continuing to support netstandard1.0 if we do make `Text` nullable.

Source/OxyPlot/Annotations/TextualAnnotation.cs
VisualMelon: Pinging @Jonarw again should be want to jump in on this

HavenDV:
I've been doing this since the early days of nullable enable and I'm already used to it, so it looks ok to me.

VisualMelon:
Cool; I suspect most other strings where we'll need these checks will be left as nullable, so shouldn't be wide-spread.

VisualMelon:
I've not done a complete review again, but I think if you're happy with the current state of this PR, then I am also happy. If so, will merge once 2.1.1 is out of the way (so hopefully next weekend).

VisualMelon (approved):

HavenDV:
Hi. Any news?

VisualMelon:
2.1.1 became 2.1.2 when I made a mistake, but let's ignore that and merge this.

PR #1936, opened 2022-09-18
Bump msbuild tool version for github actions

VisualMelon:
Hopefully this will magically fix the CI failure now that we use reproducible builds

### Checklist

Bump msbuild tool version for github actions to 1.1

@oxyplot/admins

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -5,11 +5,13 @@ jobs:
     runs-on: windows-latest
     steps:
     - uses: actions/checkout@v1
-    
+
     - name: Setup .NET Core
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.x
+        dotnet-version: | 
+          3.1.x
+          6.0.x
     
     - name: Build
       run: dotnet build Source/OxyPlot.CI.sln --configuration Release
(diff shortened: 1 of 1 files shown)


mattico:
Would it make sense to run GitHub Actions on PRs as well, so we don't only notice the failing CI after it gets merged?

VisualMelon:
Yes, ideally. I know they run for me when I push them to my own repo (I even get an email every time!), so I know that none of attempts solve the problem (See https://github.com/VisualMelon/oxyplot/actions/runs/3076290473 ). What confuses me at the moment is why github actions fails but the others don't seem to care; suggestions welcome. @HavenDV suggested bumping the tests to .NET 6 (I don't think we need to bump the package?) which I'd be fine with.

VisualMelon:
Looks like it's happy if you provide it with .NET 6 for the build, and .NET Core 3.1 for the tests. Will squash and merge soon unless someone has a better idea. (Successful build: https://github.com/VisualMelon/oxyplot/actions/runs/3077709157/jobs/4972758009 )

VisualMelon:
@mattico that seems to be working; hopefully it won't cause any rate limiting etc.

Is it useful for others to have the workflow run when they push to their own branches? I don't think (until today, maybe) it's ever picked up on anything for me, so I think having everything in one place for PRs is probably better.

PR #1937, opened 2022-09-18
Improve Project Files

mattico:
### Checklist

- As a follow-up to #1930, adds Directory.Build.props to Solution Items folders
- Add EnableWindowsTargeting to enable building .NET Framework libs from Linux and Mac
- Factor out common project properties to Directory.Build.props now that we have one of those
- A drive-by fix to some SkiaSharp deprecated functions
- Update supported TargetFrameworks to remove EoL versions and default to .NET 6.0 where possible.

@oxyplot/admins


VisualMelon:
This looks great, thanks for putting this together. I'll try to find time to look it over properly tomorrow.

Source/Directory.Build.targets
```diff
@@ -0,0 +1,10 @@
+<Project>
+  <ItemGroup Condition="$(GeneratePackageOnBuild)">
+    <None Include="$(SolutionDir)..\Icons\OxyPlot_128.png" Pack="true" PackagePath="\" />
+  </ItemGroup>
+  <PropertyGroup Condition="$(GeneratePackageOnBuild)">
+    <PackageLicenseExpression>MIT</PackageLicenseExpression>
+    <PackageProjectUrl>https://oxyplot.github.io/</PackageProjectUrl>
+    <PackageIcon>OxyPlot_128.png</PackageIcon>
+  </PropertyGroup>
+</Project>
```
mattico: The `Build.targets` file is executed towards the end of the build process, so we can depend on variables set in the individual projects. In here we have common properties for our published packages.

mattico:
@VisualMelon no rush! Thanks for taking a look.

VisualMelon:
I think we should include a comment somewhere prominent that we now require .NET 5 or .NET 6 to build the libraries, though they still target .NET Core 3.1 and need it for testing: that could create some real confusion for people trying to get it up and running the first time. To keep thing simple, may make sense to add net6.0 targets to everything now as @HavenDV suggested (not necessarily in this PR) so that basic dev tasks can all be performed with .NET 6.

VisualMelon (approved):

Jonarw (approved):
We should consider upgrading things like example browser to .NET 6 anyway, as .NET 3.1 runs out of support at the end of this year anyway (if I remember correctly). But this does not need to be part of this PR.

Source/Directory.Build.props
```diff
@@ -4,4 +4,10 @@
     <PackageReference Include="DotNet.ReproducibleBuilds" Version="1.1.1" PrivateAssets="All"/>
     <PackageReference Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" PrivateAssets="All" />
```
mattico: @Jonarw from what I understand the ReferenceAssemblies included in this NuGet package mean that we don't need to have any targeting packs installed.

VisualMelon: Certainly seems we don't need them to build the assemblies, but I think we do need to keep the example and tests projects current as they determine the runtime actually used by devs and the CI (and for the framework runtimes you need a targeting pack, I think)

Jonarw: Hmm, that didn't seem to be enough in my case. I was unable to build before installing the targeting pack manually. Of course now that I have installed it I can't reproduce the issue anymore.
In any case, I think upgrading to .NET 4.6.2 makes sense anyway, as the older versions have been out of support since April 2022.

mattico:
Okay, I updated all the targetframeworks so we're only shipping for supported targets: .NET Framework 4.6.2, .NET Standard 2.0, and .NET 6.0. I had to fix a few nullable warnings to work on .NET 6.0.

While running the tests I changed some of the tests to use `TestCaseSource` so the cases can run in parallel and so you can see which test cases are failing and debug them individually.

Source/OxyPlot.Wpf.Tests/ExampleLibraryTests.cs
```diff
@@ -9,68 +9,58 @@
...
+        const string DestinationDirectory = "ExampleLibrary.Actual";
+        const string BaselineDirectory = "ExampleLibrary.Baseline";
+        const string DiffDirectory = "ExampleLibrary.Diff";
+
+        [OneTimeSetUp]
+        public void Init()
+        {
+            Directory.CreateDirectory(BaselineDirectory);
+            Directory.CreateDirectory(DestinationDirectory);
+            Directory.CreateDirectory(DiffDirectory);
```
mattico, line 30: Aside: should this delete the existing dirs? Some of the examples use random data so they error if you run the tests multiple times without clearing out the baseline dir. Could also seed the rng in tests but that's not exposed from the examples currently.

Jonarw:
Hi @mattico!
First of all, sorry that this PR has been open for so long. We would like to get this merged finally! Would you be interested in rebasing / sorting out the merge conflicts that have accumulated? Alternatively, would it be alright with you if I took care of that?
Thanks for this contribution in any case!

mattico:
@Jonarw no worries!

Rebased. The merge conflicts were only from the changelog. Feel free to update the PR in the future if necessary.

Source/Examples/ExampleLibrary/ExampleLibrary.csproj
```diff
@@ -1,27 +1,18 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
     <TargetFrameworks>netstandard1.0;netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net462;net6.0;net6.0-windows</TargetFrameworks>
+    <TargetFrameworks>net462;net6.0;net6.0-windows</TargetFrameworks>
```
VisualMelon: Extra line

Source/Examples/PerformanceTest/PerformanceTest.csproj
```diff
@@ -1,6 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <TargetFramework>net462</TargetFramework>
+        <TargetFrameworks>net6.0-windows</TargetFrameworks>
```
VisualMelon: Extra line

Source/OxyPlot.ImageSharp/OxyPlot.ImageSharp.csproj
```diff
@@ -22,6 +16,9 @@
     <ItemGroup>
         <PackageReference Include="SixLabors.ImageSharp" Version="2.1.3" />
         <PackageReference Include="SixLabors.ImageSharp.Drawing" Version="1.0.0-beta15" />
+        <PackageReference Include="SixLabors.ImageSharp" Version="1.0.0-beta0007" />
+        <PackageReference Include="SixLabors.ImageSharp.Drawing" Version="1.0.0-beta0007" />
+        <PackageReference Include="SixLabors.Shapes.Text" Version="1.0.0-beta0009" />
```
VisualMelon: Not sure where these came from ;)

Source/OxyPlot.Pdf.Tests/OxyPlot.Pdf.Tests.csproj
```diff
@@ -1,19 +1,18 @@
...
     <PropertyGroup>
-        <TargetFramework>net462</TargetFramework>
+        <TargetFrameworks>net462</TargetFrameworks>
         <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
         <Description>OxyPlot Pdf unit tests</Description>
-        <PackageLicenseExpression>MIT</PackageLicenseExpression>
-        <Copyright>OxyPlot contributors</Copyright>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <NoWarn>CS0618</NoWarn>
+        <NoWarn>CS0618,NU1701</NoWarn>
```
VisualMelon: OK for now I think; we should investigate PDF Sharp again: looks like a net6.0 package was released 3 month ago.

Source/OxyPlot.Pdf/OxyPlot.Pdf.csproj
```diff
@@ -1,24 +1,18 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net462</TargetFramework>
+    <TargetFrameworks>net462;net6.0-windows</TargetFrameworks>
+    <TargetFrameworks>net462</TargetFrameworks>
```
VisualMelon, line 3: Extra lines

Source/OxyPlot.SkiaSharp.Wpf/OxyPlot.SkiaSharp.Wpf.csproj
```diff
@@ -1,17 +1,12 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net462;netcoreapp3.1</TargetFrameworks>
```
VisualMelon: Extra line

Source/OxyPlot.SkiaSharp/OxyPlot.SkiaSharp.csproj
```diff
@@ -4,14 +4,12 @@
     <TargetFrameworks>net462;netstandard1.3;netstandard2.0</TargetFrameworks>
     <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <Copyright>Copyright (c) 2020 OxyPlot contributors</Copyright>
+    <TargetFrameworks>net452;netstandard1.3;netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>net462;netstandard2.0</TargetFrameworks>
```
VisualMelon, line 6: Extra line; would be nice to keep 1.3 if possible I can't remember the situation with Skia

Source/OxyPlot.WindowsForms/OxyPlot.WindowsForms.csproj
```diff
@@ -1,27 +1,19 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <TargetFrameworks>net462;netcoreapp3.1</TargetFrameworks>
```
VisualMelon: Extra line

Source/OxyPlot.Wpf/OxyPlot.Wpf.csproj
```diff
@@ -1,26 +1,18 @@
-﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop">
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>net462;netcoreapp3.1</TargetFrameworks>
```
VisualMelon: Extra line

Source/OxyPlot/Imaging/Png/PngDecoder.cs
```diff
@@ -11,6 +11,9 @@ namespace OxyPlot
 {
     using System;
     using System.IO;
+#if NET
```
VisualMelon, line 14: (means net core or net5.0 or greater)

Source/OxyPlot/OxyPlot.csproj
```diff
@@ -1,26 +1,18 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
         <TargetFrameworks>netstandard1.0;netstandard2.0;net462</TargetFrameworks>
+        <TargetFrameworks>netstandard2.0;net462;net6.0</TargetFrameworks>
```
VisualMelon: Can we keep netstandard1.0?

VisualMelon (changes requested):
Looks like the target framework lines may have gone wrong in the rebase

(Thanks for coming back to this!)

Source/Examples/ExampleLibrary/ExampleLibrary.csproj
mattico: Okay wow, the Git diff algorithm was not helpful this time.

Source/OxyPlot/OxyPlot.csproj
mattico: The motivation to remove it was that every platform which supports 1.0 but not 2.0 is long EOL, and the .NET Standard 1.0 is distributed as a granular set of packages so it makes package restores take a lot longer.

I don't particularly care.

VisualMelon: You've convinced me.

Source/OxyPlot.SkiaSharp/OxyPlot.SkiaSharp.csproj
VisualMelon: Same deal as 1.0 so happy to lose it

Source/Examples/ExampleLibrary/ExampleLibrary.csproj
VisualMelon: I suspect it's because I changed most of these for the 2.1.2 release

Source/OxyPlot.Wpf.Tests/ExampleLibraryTests.cs
VisualMelon: Baseline should be left alone; the others could be deleted, but it's not a problem I don't think.

Source/OxyPlot.Wpf/XpsExporter.cs
```diff
@@ -9,7 +9,6 @@
 
 namespace OxyPlot.Wpf
 {
-#if !NETFRAMEWORK
```
mattico, line 12: cc @VisualMelon

I may be missing some context but this appears to work by fixing the condition in the csproj.

VisualMelon: I think we can scrub the condition in the csproj, but need to keep this, but I'm not familiar with this stuff.

From what I can gather from mindless googling is that the references were necessary in netcore2.0, but aren't now.

Source/Examples/ExampleLibrary/ExampleLibrary.csproj
```diff
@@ -1,27 +1,16 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-
   <PropertyGroup>
-    <TargetFrameworks>netstandard1.0;netstandard2.0;net462</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net462;net6.0;net6.0-windows</TargetFrameworks>
```
VisualMelon: Is there a benefit to having net6.0 and net6.0-windows?

mattico: To support building examples on non-windows platforms, I assume.

Hard to remember my thought process from last November.

VisualMelon: It should be OK to just have `net6.0`, as the example library doesn't depend on any windows APIs (builds and all that). Publish produces separate packages for net6.0 and net6.0-windows so wouldn't be a problem, but would suggest dropping net6.0-windows where we don't need it.

.github/workflows/dotnetcore.yml
```diff
@@ -14,12 +14,10 @@ jobs:
     steps:
     - uses: actions/checkout@v1
```
HavenDV: Use v3

.github/workflows/dotnetcore.yml
```diff
@@ -14,12 +14,10 @@ jobs:
...
 
-    - name: Setup .NET Core
+    - name: Setup .NET
       uses: actions/setup-dotnet@v1
```
HavenDV: Use v3

Source/OxyPlot.Pdf/OxyPlot.Pdf.csproj
```diff
@@ -1,24 +1,16 @@
...
     <PackageTags>pdf plotting plot charting chart</PackageTags>
-    <RepositoryType>git</RepositoryType>
-    <RepositoryUrl>https://github.com/oxyplot/oxyplot.git</RepositoryUrl>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
     <GenerateDocumentationFile>True</GenerateDocumentationFile>
-    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
     <SignAssembly>True</SignAssembly>
     <AssemblyOriginatorKeyFile>OxyPlot.Pdf.snk</AssemblyOriginatorKeyFile>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <NoWarn>NU1701</NoWarn>
```
HavenDV: Replace to `<NoWarn>$(NoWarn);NU1701<NoWarn>` to prevent override previous value

HavenDV:
I would also recommend enabling net7.0 TargetFramework wherever possible to activate the output of the most up-to-date code analyzers. It will also allow using the latest APIs via #if

Source/OxyPlot.Pdf.Tests/OxyPlot.Pdf.Tests.csproj
HavenDV: Also use `$(NoWarn)`

Source/OxyPlot.ImageSharp/OxyPlot.ImageSharp.csproj
```diff
@@ -7,7 +7,7 @@
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
         <GenerateDocumentationFile>True</GenerateDocumentationFile>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
-        <NoWarn>NU5104</NoWarn>
+        <NoWarn>$(NoWarn),NU5104</NoWarn>
```
HavenDV: use `;` as separator

Source/Directory.Build.targets
HavenDV: Typically properties and Items are added to the `Directory.Build.props` file. `Directory.Build.targets` is used for custom targets. But in general, it doesn't matter.

You can also add a README for all libraries:
```
  <PropertyGroup Label="Nuget">
    <PackageReadmeFile>README.md</PackageReadmeFile>
  </PropertyGroup>

  <ItemGroup Label="Nuget">
    <None Include="$(MSBuildThisFileDirectory)../../README.md" Pack="true" PackagePath="\" Visible="false" />
  </ItemGroup>
```

I also recommend using `$(MSBuildThisFileDirectory)` wherever possible instead of `$(SolutionDir)`. `$(SolutionDir)` is not set when running dotnet build on a specific .csproj file

HavenDV:
Thank you very much for the quick changes.

Source/OxyPlot/OxyPlot.csproj
```diff
@@ -1,26 +1,17 @@
...
-        <GenerateDocumentationFile>True</GenerateDocumentationFile>
+        <GenerateDocumentationFile>true</GenerateDocumentationFile>
         <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
-        <SignAssembly>True</SignAssembly>
+        <SignAssembly>true</SignAssembly>
         <AssemblyOriginatorKeyFile>OxyPlot.snk</AssemblyOriginatorKeyFile>
         <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
         <LangVersion>8</LangVersion>
-        <NoWarn>618</NoWarn>
+        <NoWarn>$(NoWarn);CS0618;CS8767</NoWarn>
```
VisualMelon: Would prefer we disable CS8767 only in the one place it lights up (`CompareHelper`) with `#pragma warning disable/restore CS8767`. Without changing to C#9 (which I don't think is supported on net framework) I can't find a way to make the code happy.

HavenDV: > Without changing to C#9 (which I don't think is supported on net framework) I can't find a way to make the code happy.

.Net Framework supports C# 9. You can also always use PolySharp - https://github.com/Sergio0694/PolySharp

VisualMelon: @HavenDV do you have a reference for C#9 support in net framework? I searched but could only find comments about net 5.0 and above, which I took to mean not net framework, but keen to hear otherwise.

HavenDV: I have many projects that include net451 or net462 TargetFramework and for which I install the latest version of the language. Everything works correctly. But, for example, to support records, you need to add special classes to the project (or use PolySharp, which will do the same)

VisualMelon: This states it is not supported (in 4.8 at least), but I still can't find a clear definitive reference https://github.com/dotnet/docs/issues/26587#issuecomment-947693187

HavenDV: Not everything on the internet is correct.
Probably the easiest way, if you don't believe me, is to ask the community of other C# developers
https://discord.gg/uwp - there are many developers here, including those from Microsoft. Despite the name, it's not just about UWP

Jonarw: I can confirm this, I always use the latest version of C# even on a project where I am stuck on .NET Framework. Some stuff works right out of the box, for some stuff you need to supply the runtime features (didn't know about PolySharp, thanks for the link!), and some stuff doesn't work at all. But, importantly, it doesn't break the build, even if not all features are supported.
I think we could consider bumping our `LangVersion` to 11. Doesn't have to be part of this PR though (but can, if you are motivated @mattico).

VisualMelon: I'm tempted to suggest we remove the `ComparisonHelper` API and rewrite the only usage thereof (which is confusing and depends on implementation details in `List<T>.BinarySearch`). For the time being, it's probably simplest just to disable the warning locally, which won't create a burden anywhere else.

VisualMelon: Quoting the [MSDN](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#:~:text=Choosing%20a%20language%20version%20newer%20than%20the%20default%20can%20cause%20hard%20to%20diagnose%20compile%2Dtime%20and%20runtime%20errors.)
> Choosing a language version newer than the default can cause hard to diagnose compile-time and runtime errors.

Certainly OxyPlot isn't doing anything exciting run-time wise, and I'm not aware of any situations where it actually goes wrong, but I'd still be uneasy using something that isn't officially supported.

Jonarw: With our `LangVersion` at 8 we are actually already in unsupported territory. I suspect MS writes this as they want a) people to move away from .NET Framework and b) don't want to be held responsible for any misuses.

But I agree, let's keep it simple for now, the `LangVersion` might be a discussion for another day. Let's disable the warning locally and keep everything else as is.

VisualMelon: The question for me regarding bumping the language version is really whether it can create silent bugs. Continuing the search, I can't find any evidence that this has occurred, and most discussion is just concerning using newer features.

All that said, it looks like C# 8 isn't strictly supported either, so I should probably stop worrying.

VisualMelon: (Sorry for the confusing messages: I've managed to miss both of @Jonarw 's message before sending mine somehow)

VisualMelon (approved):

VisualMelon:
All looks good to me!

VisualMelon:
@HavenDV anything else from your side? (And thank you for chiming in)

HavenDV:
```
    <PropertyGroup Label="Analyzers">
        <EnableNETAnalyzers>true</EnableNETAnalyzers>
        <AnalysisLevel>latest</AnalysisLevel>
        <AnalysisMode>All</AnalysisMode>
    </PropertyGroup>
```
I usually add this to Directory.Build.props.
But in fact, this is likely to generate a lot of warnings and is suitable for a single PR or a series of PRs.

VisualMelon:
@Jonarw I'm happy for this to be merged if you are

Jonarw:
Thanks @mattico for this PR (and for your patience)!

PR #1962, opened 2022-11-26
Release/v2.1.1

VisualMelon:
Proposal for 2.1.1 release: includes bug fixes, security fixed, and a couple of tiny features.

Reverted the targets so that we don't change the release targets for the point release. Will re-apply the reverted commit after the release.

Pushed the core packages to https://www.myget.org/feed/Packages/melonoxyplottesting for testing...

@oxyplot/admins


VisualMelon:
Not off to a great start: pushed the wrong package to nuget for OxyPlot-core by accident (now unlisted) and having trouble running the tests... may need to bump NUnit.

Issue #1419, opened 2019-11-03
Hit testing for axes


objorke:
Add support for hit-testing on axes

objorke:
See #1415

objorke:
@silviubogan You have implemented this in your fork, can you create a PR for this change only?

https://github.com/oxyplot/oxyplot/pull/1415/files#diff-2c8edecb3bf6753459205e53722c4c22R1758

silviubogan:
@objorke I think it works (I just copied and pasted existing code) although I did not test [the code](https://github.com/silviubogan/oxyplot/commit/1e6738f33b7a46eb38a1d5c26f43fc1961b266a6), a reason for this is that I do not know for what hit testing for axes will be used excepting tooltips. Please let me know if you agree with the creation of the pull request based on the commit from the link above. Thank you.

objorke:
I would prefer if the hit testing code was in the `Axis` class, and not the `PlotModel`, similar to the `Series` and `Annotations`. I think the `GetAxesFromPoint` has slightly different usage (it is used for zoom and pan handlers I think) where it also needs to find axes when clicking inside the "plot area". I would prefer not adding the overload on `GetAxesFromPoint`.

Yes, tooltips is probably the first need for hit testing on axes, but I think it can also be useful if you have a user interface where you can click on an axis to select it and edit min/max values.

silviubogan:
@objorke I think I have remade it in the way you suggested. There is [just one commit](https://github.com/silviubogan/oxyplot/commit/31449bd0ce5c07d17b2eb3b414d015db551e3da5). I have also tested it with [this *example*](https://gist.github.com/silviubogan/027519a356c90265e79e8b98f30df32e). Please let me know if I can make the PR (I will include the *example* in the PR only if you suggest this). Thank you.

objorke:
Can you create a branch and include both the example and the changes?

Do we need to add all the position tier logic in the hit test code? can the axis simply keep a bounds rectangle (that it updates during rendering) that we can check against? 
I think the hit test code could be something like "return this.axisBounds is not empty && this.axisBounds 
 contains args.Point"

silviubogan:
@objorke 

> Can you create a branch and include both the example and the changes?

Yes.

> Do we need to add all the position tier logic in the hit test code?

No.

> can the axis simply keep a bounds rectangle (that it updates during rendering) that we can check against?
> I think the hit test code could be something like "return this.axisBounds is not empty && this.axisBounds
[...]

PR #1966, opened 2022-12-06
Add AxisArea, AxisLineArea and AxisTitleArea property in Axis class (…

### LINKED ISSUE #1949: Axis area and the axis title area property in axis class
Hi,

I'm building an WPF application for analyzing measurement data. The frontend used OxyPlot to display the data. The user is able to select any part of the plot to change the properties over the suitable controls in the menu bar. By selecting a plot part (i.e. series, legends, annotations) it is highlighted with a dotted border with a padding. I can select almost everything but the axes, as the axes area seams not to be captured as area? Or am I blind? :-)

### Feature description

It would help to have a area property in the axis class for the axis area and the axis title area to be able to check a screenpoint is within.

Best regards
Chris

ChrisCC6:
Fixes #1949  .

### Checklist

- Add AxisArea, AxisLineArea and AxisTitleArea property in Axis class (#1949)
- Implement setting added area properties in axis renderer classes HorizontalAndVerticalAxisRenderer, AngleAxisRenderer and MagnitudeAxisRenderer
- Add OxyCircle and OxyAnnulus class
- Add IShape interface for OxyRect, OxyCircle and OxyAnnulus
- Add Union function to OxyRect class

@oxyplot/admins

### DIFF
--- a/Source/Examples/ExampleLibrary/Axes/AxisAreaExample.cs
+++ b/Source/Examples/ExampleLibrary/Axes/AxisAreaExample.cs
@@ -1,0 +1,158 @@
+﻿using OxyPlot;
+using OxyPlot.Axes;
+using OxyPlot.Series;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace ExampleLibrary.Axes
+{
+    [Examples("AxisArea"), Tags("Axes")]
+    public class AxisAreaExample
+    {
+        [Example("AxisArea LinearAxis")]
+        public static PlotModel AxisAreaLinearAxis()
+        {
+            var model = GetLinearAxisPlotModel();
+            model.Annotations.Add(new RenderingCapabilities.DelegateAnnotation(context =>
+            {
+                foreach (var axis in model.Axes)
+                {
+                    context.DrawRectangle((OxyRect)axis.AxisArea, OxyColors.Undefined, OxyColor.FromAColor(128, OxyColors.Red), 1, EdgeRenderingMode.Automatic);
+                    context.DrawRectangle((OxyRect)axis.TitleArea, OxyColors.Undefined, OxyColor.FromAColor(128, OxyColors.Green), 1, EdgeRenderingMode.Automatic);
+                    context.DrawRectangle((OxyRect)axis.AxisLineArea, OxyColors.Undefined, OxyColor.FromAColor(128, OxyColors.Blue), 1, EdgeRenderingMode.Automatic);
+                }
+            }));
+            // Subscribe to the mouse down event on the line series
+            model.MouseDown += (s, e) =>
+            {
+                ResetAxesFormat(model);
+                if (e.ChangedButton == OxyMouseButton.Left)
+                {
+                    if (model.Axes.SingleOrDefault(axis => axis.TitleArea.Contains(e.Position.X, e.Position.Y)) is Axis clickedAxis1)
+                    {
+                        clickedAxis1.TitleFontWeight = 800;
+                        clickedAxis1.TitleColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                    }
+                    else if (model.Axes.SingleOrDefault(axis => axis.AxisLineArea.Contains(e.Position.X, e.Position.Y)) is Axis clickedAxis2)
+                    {
+                        clickedAxis2.FontWeight = 800;
+                        clickedAxis2.TextColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                    }
+                    else if (model.Axes.SingleOrDefault(axis => axis.AxisArea.Contains(e.Position.X, e.Position.Y)) is Axis clickedAxis3)
+                    {
+                        clickedAxis3.FontWeight = 800;
+                        clickedAxis3.TitleFontWeight = 800;
+                        clickedAxis3.TitleColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                        clickedAxis3.TextColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                    }
+                }
+                model.InvalidatePlot(false);
+            };
+            return model;
+        }
+
+        [Example("AxisArea AngleAxis")]
+        public static PlotModel AxisAreaAngleAxis()
+        {
+            var model = GetAngleAxisPlotModel();
+            model.Annotations.Add(new RenderingCapabilities.DelegateAnnotation(context =>
+            {
+                context.DrawRectangle((OxyRect)model.Axes[1].AxisArea, OxyColors.Undefined, OxyColor.FromAColor(255, OxyColors.Red), 1, EdgeRenderingMode.Automatic);
+                var annulus = (OxyAnnulus)model.Axes[0].AxisArea;
+                context.DrawCircle(annulus.Center, annulus.InnerRadius, OxyColors.Undefined, OxyColor.FromAColor(255, OxyColors.Blue), 1, EdgeRenderingMode.Automatic);
+                context.DrawCircle(annulus.Center, annulus.OuterRadius, OxyColors.Undefined, OxyColor.FromAColor(255, OxyColors.Blue), 1, EdgeRenderingMode.Automatic);
+            }));
+                // Subscribe to the mouse down event on the line series
+            model.MouseDown += (s, e) =>
+            {
+                ResetAxesFormat(model);
+                if (e.ChangedButton == OxyMouseButton.Left)
+                {
+                    if (model.Axes.SingleOrDefault(axis => axis.TitleArea.Contains(e.Position.X, e.Position.Y)) is Axis clickedAxis1)
+                    {
+                        clickedAxis1.TitleFontWeight = 800;
+                        clickedAxis1.TitleColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                    }
+                    else if (model.Axes.SingleOrDefault(axis => axis.AxisLineArea.Contains(e.Position.X, e.Position.Y)) is Axis clickedAxis2)
+                    {
+                        clickedAxis2.FontWeight = 800;
+                        clickedAxis2.TextColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                    }
+                    else if (model.Axes.SingleOrDefault(axis => axis.AxisArea.Contains(e.Position.X, e.Position.Y)) is Axis clickedAxis3)
+                    {
+                        clickedAxis3.FontWeight = 800;
+                        clickedAxis3.TitleFontWeight = 800;
+                        clickedAxis3.TitleColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                        clickedAxis3.TextColor = OxyColor.FromAColor(255, OxyColors.Purple);
+                    }
+                }
+                model.InvalidatePlot(false);
+            };
+            return model;
+        }
+
+        private static void ResetAxesFormat(PlotModel model)
+        {
+            foreach (var axis in model.Axes)
+            {
+                axis.TitleFontWeight = 400;
+                axis.FontWeight = 400;
+                axis.TitleColor = OxyColor.FromAColor(255, OxyColors.Black);
+                axis.TextColor = OxyColor.FromAColor(255, OxyColors.Black);
+            }
+        }
+
+        private static PlotModel GetLinearAxisPlotModel()
+        {
+            var model = new PlotModel();
+            model.Axes.Add(new LinearAxis { Maximum = 36, Minimum = 0, Title = "km/h" });
+            model.Axes.Add(new LinearAxis { Maximum = 10, Minimum = 0, Position = AxisPosition.Right, Title = "m/s" });
+            model.Axes.Add(new LinearAxis
+            {
+                Maximum = 10,
+                Minimum = 0,
+                Position = AxisPosition.Bottom,
+                Title = "meter"
+            });
+            model.Axes.Add(new LinearAxis
+            {
+                Maximum = 10000,
+                Minimum = 0,
+                Position = AxisPosition.Top,
+                Title = "millimeter"
+            });
+            return model;
+        }
+
+        private static PlotModel GetAngleAxisPlotModel()
+        {
+            var model = new PlotModel
+            {
+                PlotType = PlotType.Polar,
+                PlotAreaBorderThickness = new OxyThickness(0),
+            };
+            model.Axes.Add(
+                new AngleAxis
+                {
+                    MajorStep = Math.PI / 4,
+                    MinorStep = Math.PI / 16,
+                    MajorGridlineStyle = LineStyle.Solid,
+                    MinorGridlineStyle = LineStyle.Solid,
+                    FormatAsFractions = true,
+                    FractionUnit = Math.PI,
+                    FractionUnitSymbol = "π",
+                    Minimum = 0,
+                    Maximum = 2 * Math.PI
+                });
+            model.Axes.Add(new MagnitudeAxis
+            {
+                MajorGridlineStyle = LineStyle.Solid,
+                MinorGridlineStyle = LineStyle.Solid,
+            });
+            model.Series.Add(new FunctionSeries(t => t, t => t, 0, Math.PI * 6, 0.01));
+            return model;
+        }
+    }
+}
--- a/Source/Examples/WPF/WpfExamples/App.xaml
+++ b/Source/Examples/WPF/WpfExamples/App.xaml
@@ -1,7 +1,7 @@
 ﻿<Application x:Class="WpfExamples.App"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             StartupUri="MainWindow.xaml">
+             StartupUri="Examples/AxisAreaDemo/MainWindow.xaml">
     <Application.Resources>
          
     </Application.Resources>
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -133,6 +133,10 @@ namespace OxyPlot.Axes
 
             this.DataMaximum = double.NaN;
             this.DataMinimum = double.NaN;
+
+            this.TitleArea = new EmptyShape();
+            this.AxisArea = new EmptyShape();
+            this.AxisLineArea = new EmptyShape();
         }
 
         /// <summary>
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -641,17 +641,17 @@ namespace OxyPlot.Axes
         /// <summary>
         /// The area of the axis in screen coordinates.
         /// </summary>
-        public IOxyRegion AxisArea { get; internal set; }
+        public IOxyRegion AxisArea { get; internal protected set; }
 
         /// <summary>
         /// The area of the axis line in screen coordinates.
         /// </summary>
-        public IOxyRegion AxisLineArea { get; internal set; }
+        public IOxyRegion AxisLineArea { get; internal protected set; }
 
         /// <summary>
         /// The area of the axis title in screen coordinates.
         /// </summary>
-        public IOxyRegion TitleArea { get; internal set; }
+        public IOxyRegion TitleArea { get; internal protected set; }
 
         /// <summary>
         /// Gets or sets the position tier max shift.
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -635,6 +635,21 @@ namespace OxyPlot.Axes
         public OxyThickness DesiredMargin { get; protected set; }
 
         /// <summary>
+        /// The area of the axis in screen coordinates.
+        /// </summary>
+        public IShape AxisArea { get; internal set; }
+
+        /// <summary>
+        /// The area of the axis line in screen coordinates.
+        /// </summary>
+        public IShape AxisLineArea { get; internal set; }
+
+        /// <summary>
+        /// The area of the axis title in screen coordinates.
+        /// </summary>
+        public IShape TitleArea { get; internal set; }
+
+        /// <summary>
         /// Gets or sets the position tier max shift.
         /// </summary>
         internal double PositionTierMaxShift { get; set; }
--- a/Source/OxyPlot/Axes/Axis.cs
+++ b/Source/OxyPlot/Axes/Axis.cs
@@ -641,17 +641,17 @@ namespace OxyPlot.Axes
         /// <summary>
         /// The area of the axis in screen coordinates.
         /// </summary>
-        public IShape AxisArea { get; internal set; }
+        public IOxyRegion AxisArea { get; internal set; }
 
         /// <summary>
         /// The area of the axis line in screen coordinates.
         /// </summary>
-        public IShape AxisLineArea { get; internal set; }
+        public IOxyRegion AxisLineArea { get; internal set; }
 
         /// <summary>
         /// The area of the axis title in screen coordinates.
         /// </summary>
-        public IShape TitleArea { get; internal set; }
+        public IOxyRegion TitleArea { get; internal set; }
 
         /// <summary>
         /// Gets or sets the position tier max shift.
--- a/Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
+++ b/Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
@@ -52,7 +52,7 @@ namespace OxyPlot.Axes
             var axisLength = Math.Abs(scaledEndAngle - scaledStartAngle);
             var eps = axis.MinorStep * 1e-3;
 
-            ScreenPoint someScreenPoint = ScreenPoint.Undefined;
+            ScreenPoint? innerRadiusTangentScreenPoint = null;
             double maxTextLength = double.MinValue;
 
             if (this.MinorPen != null)
@@ -62,7 +62,7 @@ namespace OxyPlot.Axes
                     .Take(tickCount + 1)
                     .Select(x => magnitudeAxis.Transform(magnitudeAxis.ClipMaximum, x, axis));
 
-                someScreenPoint = screenPoints.FirstOrDefault();
+                innerRadiusTangentScreenPoint = screenPoints.FirstOrDefault();
                 foreach (var screenPoint in screenPoints)
                 {
                     this.RenderContext.DrawLine(magnitudeAxis.MidPoint.x, magnitudeAxis.MidPoint.y, screenPoint.x, screenPoint.y, this.MinorPen, axis.EdgeRenderingMode);
@@ -82,8 +82,8 @@ namespace OxyPlot.Axes
                     .Take(majorTickCount)
                     .Select(x => magnitudeAxis.Transform(magnitudeAxis.ClipMaximum, x, axis));
 
-                if (someScreenPoint.Equals(ScreenPoint.Undefined) && screenPoints.Count() > 0)
-                    someScreenPoint = screenPoints.FirstOrDefault();
+                if (innerRadiusTangentScreenPoint == null && screenPoints.Count() > 0)
+                    innerRadiusTangentScreenPoint = screenPoints.FirstOrDefault();
                 foreach (var point in screenPoints)
                 {
                     this.RenderContext.DrawLine(magnitudeAxis.MidPoint.x, magnitudeAxis.MidPoint.y, point.x, point.y, this.MajorPen, axis.EdgeRenderingMode);
@@ -125,9 +125,9 @@ namespace OxyPlot.Axes
                     pt, text, axis.ActualTextColor, axis.ActualFont, axis.ActualFontSize, axis.ActualFontWeight, angle, ha, va);
             }
 
-            if (!someScreenPoint.Equals(ScreenPoint.Undefined))
+            if (innerRadiusTangentScreenPoint != null)
             {
-                double r1 = magnitudeAxis.MidPoint.DistanceTo(someScreenPoint);
+                double r1 = magnitudeAxis.MidPoint.DistanceTo(innerRadiusTangentScreenPoint.Value);
                 double r2 = r1 + maxTextLength + axis.AxisTickToLabelDistance;
                 if (r2 < 5)
                     r2 = r1 + 5;
@@ -136,8 +136,8 @@ namespace OxyPlot.Axes
             }
             else
             {
-                axis.AxisLineArea = new OxyCircle(0, 0, 0);
-                axis.AxisArea = new OxyCircle(0, 0, 0);
+                axis.AxisLineArea = new EmptyShape();
+                axis.AxisArea = new EmptyShape();
             }
         }
     }
--- a/Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
+++ b/Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
@@ -52,7 +52,7 @@ namespace OxyPlot.Axes
             var axisLength = Math.Abs(scaledEndAngle - scaledStartAngle);
             var eps = axis.MinorStep * 1e-3;
 
-            ScreenPoint? innerRadiusTangentScreenPoint = null;
+            ScreenPoint? innerTangentPoint = null;
             double maxTextLength = double.MinValue;
 
             if (this.MinorPen != null)
@@ -62,7 +62,7 @@ namespace OxyPlot.Axes
                     .Take(tickCount + 1)
                     .Select(x => magnitudeAxis.Transform(magnitudeAxis.ClipMaximum, x, axis));
 
-                innerRadiusTangentScreenPoint = screenPoints.FirstOrDefault();
+                innerTangentPoint = screenPoints.FirstOrDefault();
                 foreach (var screenPoint in screenPoints)
                 {
                     this.RenderContext.DrawLine(magnitudeAxis.MidPoint.x, magnitudeAxis.MidPoint.y, screenPoint.x, screenPoint.y, this.MinorPen, axis.EdgeRenderingMode);
@@ -82,8 +82,8 @@ namespace OxyPlot.Axes
                     .Take(majorTickCount)
                     .Select(x => magnitudeAxis.Transform(magnitudeAxis.ClipMaximum, x, axis));
 
-                if (innerRadiusTangentScreenPoint == null && screenPoints.Count() > 0)
-                    innerRadiusTangentScreenPoint = screenPoints.FirstOrDefault();
+                if (innerTangentPoint == null && screenPoints.Count() > 0)
+                    innerTangentPoint = screenPoints.FirstOrDefault();
                 foreach (var point in screenPoints)
                 {
                     this.RenderContext.DrawLine(magnitudeAxis.MidPoint.x, magnitudeAxis.MidPoint.y, point.x, point.y, this.MajorPen, axis.EdgeRenderingMode);
@@ -125,9 +125,9 @@ namespace OxyPlot.Axes
                     pt, text, axis.ActualTextColor, axis.ActualFont, axis.ActualFontSize, axis.ActualFontWeight, angle, ha, va);
             }
 
-            if (innerRadiusTangentScreenPoint != null)
+            if (innerTangentPoint != null)
             {
-                double r1 = magnitudeAxis.MidPoint.DistanceTo(innerRadiusTangentScreenPoint.Value);
+                double r1 = magnitudeAxis.MidPoint.DistanceTo(innerTangentPoint.Value);
                 double r2 = r1 + maxTextLength + axis.AxisTickToLabelDistance;
                 if (r2 < 5)
                     r2 = r1 + 5;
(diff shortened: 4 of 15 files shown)


Source/OxyPlot/Rendering/OxyCircle.cs
```diff
@@ -0,0 +1,197 @@
...
+            return new OxyCircle(this.Center.X + offsetX, this.Center.Y + offsetY, this.Radius);
+        }
+
+        /// <summary>
+        /// Intersects this <see cref="OxyCircle"/> with another <see cref="OxyCircle"/> with an equal center point.
+        /// </summary>
+        /// <param name="circle">The other <see cref="OxyCircle"/>.</param>
+        /// <returns>The intersection between this <see cref="OxyCircle"/> and the other <see cref="OxyCircle"/>.</returns>
+        /// <exception cref="System.ArgumentException">circle;The center should be equal to this center.</exception>
+        public OxyAnnulus Intersect(OxyCircle circle)
```
VisualMelon: I think `Intersect` is the wrong word here, though I'm not entirely sure what the idea of this operation is

Source/OxyPlot/Rendering/OxyCircle.cs
```diff
@@ -0,0 +1,197 @@
...
+            builder.Append(")");
+            return builder.ToString();
+        }
+
+        /// <summary>
+        /// Returns a circle that is expanded or shrunk by the specified radius amounts.
+        /// </summary>
+        /// <param name="dr">The amount by which to expand or shrink the radius of the circle.</param>
+        /// <returns>The expanded/shrunk <see cref="OxyCircle" />.</returns>
+        public OxyCircle Inflate(double dr)
```
VisualMelon: Not used; would omit for the time being as the inflate could be misunderstood as a factor.

VisualMelon: (Similarly for annulus)

Source/OxyPlot/Rendering/OxyAnnulus.cs
```diff
@@ -0,0 +1,229 @@
...
+        /// </summary>
+        /// <param name="center">The center location of the hypoid</param>
+        /// <param name="r1">The inner radius of the hypoid</param>
+        /// <param name="r2">The outer radius of the hypoid</param>
+        /// <exception cref="System.ArgumentOutOfRangeException">r1;The inner radius should not be negative.</exception>
+        /// <exception cref="System.ArgumentOutOfRangeException">r2;The outer radius should not be equal or smaller as the inner radius.</exception>
+        public OxyAnnulus(ScreenPoint center, double r1, double r2)
+        {
+            if (r1 < 0)
+                throw new ArgumentOutOfRangeException("r1", "The inner radius should not be negative.");
```
VisualMelon: Should use `nameof(r1)` etc. to minimise refactoring burden

VisualMelon: Would prefer `innerRadius`, `outerRadius` as the parameter names themselves

Source/OxyPlot/Rendering/OxyAnnulus.cs
```diff
@@ -0,0 +1,229 @@
...
+        public OxyAnnulus(double x, double y, double r1, double r2) : this(new ScreenPoint(x, y), r1, r2)
+        {
+        }
+
+        /// <summary>
+        /// Determines whether the specified point is inside the hypoid.
+        /// </summary>
+        /// <param name="x">The x coordinate.</param>
+        /// <param name="y">The y coordinate.</param>
+        /// <returns><c>true</c> if the hypoid contains the specified point; otherwise, <c>false</c>.</returns>
```
VisualMelon: hypoid -> annulus

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -19,6 +19,11 @@ namespace OxyPlot.Axes
     /// </summary>
     public class HorizontalAndVerticalAxisRenderer : AxisRendererBase
     {
+        private ScreenPoint pt0;
```
VisualMelon: These should be locals

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
```diff
@@ -51,13 +51,18 @@ public override void Render(Axis axis, int pass)
 
             var axisLength = Math.Abs(scaledEndAngle - scaledStartAngle);
             var eps = axis.MinorStep * 1e-3;
+
+            ScreenPoint someScreenPoint = ScreenPoint.Undefined;
```
VisualMelon: This could do with a better name

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
```diff
@@ -51,13 +51,18 @@ public override void Render(Axis axis, int pass)
...
+            double maxTextLength = double.MinValue;
+
             if (this.MinorPen != null)
             {
                 var tickCount = Math.Abs((int)(axisLength / axis.ActualMinorStep));
                 var screenPoints = this.MinorTickValues
                     .Take(tickCount + 1)
                     .Select(x => magnitudeAxis.Transform(magnitudeAxis.ClipMaximum, x, axis));
 
+                someScreenPoint = screenPoints.FirstOrDefault();
```
VisualMelon: Not tested, but note that the default with 0,0, not an undefined point, as it looks is expected: looks like this needs to be revised (or I could missing the point)

Source/OxyPlot/Rendering/OxyCircle.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <param name="r">The radius of the circle</param>
+        /// <exception cref="System.ArgumentOutOfRangeException">r;The radius should not be negative.</exception>
+        public OxyCircle(ScreenPoint center, double r)
+        {
+            if (r < 0)
+                throw new ArgumentOutOfRangeException("r", "The radius should not be negative.");
+
+            this.Center = new ScreenPoint(center.X, center.Y);
+            this.Radius = r;
+
```
VisualMelon: random space

Source/OxyPlot/Rendering/OxyCircle.cs
```diff
@@ -0,0 +1,197 @@
...
+        /// <value>The most right point.</value>
+        public ScreenPoint Right => new ScreenPoint(this.Center.X + this.Radius, this.Center.Y);
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OxyCircle"/> structure that has a specified center and a radius
+        /// </summary>
+        /// <param name="center">The center location of the circle</param>
+        /// <param name="r">The radius of the circle</param>
+        /// <exception cref="System.ArgumentOutOfRangeException">r;The radius should not be negative.</exception>
+        public OxyCircle(ScreenPoint center, double r)
```
VisualMelon: Prefer `radius` over `r`

Source/OxyPlot/Rendering/OxyCircle.cs
```diff
@@ -0,0 +1,197 @@
...
+    /// Describes the center and the radius of a circle.
+    /// </summary>
+    public struct OxyCircle : IShape, IEquatable<OxyCircle>, IFormattable
+    {
+
+        /// <summary>
+        /// Gets the center point of the circle.
+        /// </summary>
+        /// <value>The center.</value>
+        public ScreenPoint Center { get; set; }
```
VisualMelon: Should be getter-only

Source/OxyPlot/Rendering/OxyCircle.cs
```diff
@@ -0,0 +1,197 @@
...
+
+        /// <summary>
+        /// Gets the radius of the circle
+        /// </summary>
+        public double Radius { get; set; }
```
VisualMelon: As above

Source/OxyPlot/Rendering/IShape.cs
```diff
@@ -0,0 +1,36 @@
...
+
+        /// <summary>
+        /// Returns a <see cref="System.String" /> that represents this instance.
+        /// </summary>
+        /// <param name="format">The format.</param>
+        /// <param name="formatProvider">The format provider.</param>
+        /// <returns>
+        /// A <see cref="System.String" /> that represents this instance.
+        /// </returns>
+        public string ToString(string format, IFormatProvider formatProvider);
```
VisualMelon: redundant `public`; random space below this; can we just derive `IFormattable`

Source/OxyPlot/Axes/Rendering/MagnitudeAxisRenderer.cs
```diff
@@ -40,6 +40,9 @@ public override void Render(Axis axis, int pass)
 
             var angleAxis = this.Plot.DefaultAngleAxis;
 
+            angleAxis.AxisLineArea = new OxyRect(0, 0, 0, 0);
```
VisualMelon: Worth adding a null IShape?

VisualMelon (commented):
Not a complete review, just found a few minutes to take a look.

Source/OxyPlot/Axes/Rendering/MagnitudeAxisRenderer.cs
VisualMelon: (i.e. a concrete type that represents a zero-area space)

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: I used the class OxyRect as template for OxyCircle and adapted the functionality to "circles"... Might be deleted.

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: as above

Source/OxyPlot/Rendering/OxyAnnulus.cs
ChrisCC6: agreed & changed

Source/OxyPlot/Rendering/OxyAnnulus.cs
ChrisCC6: done

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
ChrisCC6: agreed & changed

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
ChrisCC6: maybe

`ScreenPoint? innerRadiusTangentScreenPoint = null;`

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: agreed & changed

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: agreed & changed

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: changed to private setter

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: as above

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
VisualMelon: `innermostPoint` would make sense; may need to check if reversing the axis messes with this (e.g. setting `StartPosition=1` and `EndPosition=0`)

Source/OxyPlot/Rendering/OxyCircle.cs
VisualMelon: I think keep the APIs simple for the time being: it's easy to add more methods when we need them.

Source/OxyPlot/Rendering/IShape.cs
ChrisCC6: redundant `public` & random space is deleted;

regarding "can we just derive `IFormattable`", I'm not really sure what you mean

VisualMelon: `IFormattable` already has a method with the same signature, so could just extend it in `IShape` 

https://learn.microsoft.com/en-us/dotnet/api/system.iformattable.tostring?view=net-7.0

Source/OxyPlot/Rendering/OxyCircle.cs
VisualMelon: Feel free to push new commits: will update the PR so we can discharge little things like this.

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: ok, deleted

Source/OxyPlot/Axes/Rendering/MagnitudeAxisRenderer.cs
ChrisCC6: I'm currently struggling how to do this. Could you give me a hint?

VisualMelon: I just mean a boring implementation of `IShape` where `bool Contains(double x, double y);` always returns `false`; call it `EmptyShape` or `NullShape` or something (suggestions for better names welcome): should be nicer than using zero-dimension Rects.

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
ChrisCC6: Is it not the case that all points  of the angle axis have the same distance to the center? Am I getting it wrong right now?

VisualMelon: Ah, yes of course: sorry, my mistake.

Source/OxyPlot/Axes/Rendering/MagnitudeAxisRenderer.cs
ChrisCC6: ok, done

VisualMelon: Incidentally, I'm thinking `IOxyRegion` might be a better name than `IShape`, but that's easy to change once the everything else is nailed down. Thoughts welcome.

VisualMelon:
Thanks for updating: I'll try to review everything again soon.

Source/OxyPlot/Rendering/IShape.cs
ChrisCC6: I deleted the two `ToString()` functions

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
ChrisCC6: ... but I admit: `innerRadiusTangentScreenPoint` is not nice to read...
optional: `innerTangentPoint` ?

Source/OxyPlot/Axes/Rendering/MagnitudeAxisRenderer.cs
ChrisCC6: Sounds good... Sold

Source/OxyPlot/Rendering/EmptyShape.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// Describes an empty shape
+    /// </summary>
+    public struct EmptyShape : IOxyRegion
+    {
+        /// <summary>
+        /// A specified point will never be within an empty shape.
+        /// </summary>
+        /// <param name="x"></param>
+        /// <param name="y"></param>
+        /// <returns>false</returns>
```
VisualMelon: Can use `<inheritdoc/>` for these

Source/OxyPlot/Rendering/OxyAnnulus.cs
```diff
@@ -0,0 +1,207 @@
...
+
+        /// <summary>
+        /// Gets the inner radius of the annulus
+        /// </summary>
+        public double InnerRadius { get; set; }
+
+        /// <summary>
+        /// Gets the outer radius of the annulus
+        /// </summary>
+        public double OuterRadius { get; set; }
```
VisualMelon: These should be getter only (i.e. `{ get; }`)

Source/OxyPlot/Rendering/OxyCircle.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <summary>
+        /// Gets the center point of the circle.
+        /// </summary>
+        /// <value>The center.</value>
+        public ScreenPoint Center { get; private set; }
+
+        /// <summary>
+        /// Gets the radius of the circle
+        /// </summary>
+        public double Radius { get; private set; }
```
VisualMelon: These should be getter only (i.e. `{ get; }`) (only assigned in the constructor, so don't need a private setter)

Source/OxyPlot/Axes/Axis.cs
```diff
@@ -634,6 +638,21 @@ public double Scale
...
+        /// <summary>
+        /// The area of the axis line in screen coordinates.
+        /// </summary>
+        public IOxyRegion AxisLineArea { get; internal set; }
+
+        /// <summary>
+        /// The area of the axis title in screen coordinates.
+        /// </summary>
+        public IOxyRegion TitleArea { get; internal set; }
+
```
VisualMelon: Probably need `internal protected` setter

Source/OxyPlot/Rendering/OxyAnnulus.cs
```diff
@@ -0,0 +1,207 @@
+﻿// --------------------------------------------------------------------------------------------------------------------
+// <copyright file="OxyCircle.cs" company="OxyPlot">
```
VisualMelon: Circle -> Annulus

Source/OxyPlot/Rendering/OxyAnnulus.cs
```diff
@@ -0,0 +1,207 @@
...
+            if (outerRadius <= innerRadius)
+                throw new ArgumentOutOfRangeException(nameof(outerRadius), "The outer radius should not be equal or smaller as the inner radius.");
+
+            this.Center = new ScreenPoint(center.X, center.Y);
+            this.InnerRadius = innerRadius;
+            this.OuterRadius = outerRadius;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="OxyAnnulus"/> structure that has a specified center a inner and outer radius
```
VisualMelon: "has a specified center a inner and outer radius" -> "has the specified center, inner radius, and outer radius"

Source/OxyPlot/Rendering/OxyAnnulus.cs
```diff
@@ -0,0 +1,207 @@
...
+        /// </summary>
+        /// <param name="x">The x-coordinate location of the center of the circle</param>
+        /// <param name="y">The x-coordinate location of the center of the circle</param>
```
VisualMelon: circle -> annulus

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -248,6 +249,9 @@ protected static void SnapTo(double target, ref double v, double eps = 0.5)
         /// <param name="titlePosition">The title position.</param>
         protected virtual void RenderAxisTitle(Axis axis, double titlePosition)
         {
+            ScreenPoint pt0 = new ScreenPoint(0, 0);
+            ScreenPoint pt1 = new ScreenPoint(0, 0);
```
VisualMelon: can we move these closer to where they are used, and can we rename them to `areaTopLeft` and `areaBottomRight` or similar?

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -19,6 +19,7 @@ namespace OxyPlot.Axes
     /// </summary>
     public class HorizontalAndVerticalAxisRenderer : AxisRendererBase
     {
+
```
VisualMelon: random space

VisualMelon:
This is looking good now (few typos, but otherwise public API looks nice), and provides a nice foundation we can build on in future (can imagine providing regions/areas (maybe that implies some name changes) for other elements).... but I'm not sure for what purposes at the minute.

I'm struggling to think of a nice simple example we can include in the example library to help demonstrate this functionality... it is obviously useful for your custom interaction example, but as it can't be a substitute for the terrifying `PlotModel.GetAxes` implementation as I was originally thinking, and the event system is deprecated, I think we're going to struggle to find a use that fits in with the generally supported scenarios in the library.

Suggestions very welcome.

Location information is useful for user interaction, and for positioning other UI elements, e.g. custom annotations.

Source/OxyPlot/Rendering/EmptyShape.cs
ChrisCC6: done

Source/OxyPlot/Rendering/OxyAnnulus.cs
ChrisCC6: changed to private setters

Source/OxyPlot/Axes/Axis.cs
ChrisCC6: changed

Source/OxyPlot/Rendering/OxyAnnulus.cs
ChrisCC6: done

Source/OxyPlot/Rendering/OxyAnnulus.cs
ChrisCC6: done

Source/OxyPlot/Rendering/OxyAnnulus.cs
ChrisCC6: done

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
ChrisCC6: deleted

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
ChrisCC6: done

Source/OxyPlot/Rendering/OxyAnnulus.cs
VisualMelon: I'd prefer these were strictly immutable after construction like `OxyRect`: am I missing a reason why they need a setter?

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
VisualMelon: Given the points are all on the circumference, can we just derive one using `magnitudeAxis.Transform(magnitudeAxis.ClipMaximum, <some sensible value>, axis))`?

Actually, that inspires a name: `pointOnInnerCircumference` (long, but explicit, and hopefully should be clear if I've missed the point)

ChrisCC6:
> I'm struggling to think of a nice simple example we can include in the example library to help demonstrate this functionality... it is obviously useful for your custom interaction example, but as it can't be a substitute for the terrifying `PlotModel.GetAxes` implementation as I was originally thinking, and the event system is deprecated, I think we're going to struggle to find a use that fits in with the generally supported scenarios in the library.
> 
> Suggestions very welcome.
> 
> Location information is useful for user interaction, and for positioning other UI elements, e.g. custom annotations.

Need to find some time, but I'll try to create an example for this. Maybe like I use it (maybe not that simple):
I use wrapper classes around axes, series and annotations, overwriting the `Render` method to draw a dotted border depending on a `bool IsSelected` property.
Also a custom implementation for `MouseManipulator`. Here I check if one area contains the MouseUpPosition, set the `IsSelected = true` and request a new rendering. At the same time I send a message to notify a parent class representing a analyze that the selected object is changed. A public property for the selected item (my wrapper classes) is bound to my ribbon toolbar. This is how I make the plot and all elements editable via controls to have adaptable for the user on runtime.
Example for selected text:
image

I hope you can follow my thoughts. For an example, I'll try to strip it down to some basic functionality.

Source/OxyPlot/Rendering/OxyAnnulus.cs
ChrisCC6: That's ok for me. I'm used to do it that way. Maybe something I should think about... Refactoring maybe takes longer in some cases, but provides some not neglectable safety.
Maybe I'll use this concept more in the future.

Source/OxyPlot/Rendering/OxyCircle.cs
ChrisCC6: as above

Source/OxyPlot/Rendering/IOxyRegion.cs
```diff
@@ -0,0 +1,19 @@
...
+    public interface IOxyRegion
+    {
+        /// <summary>
+        /// Determines whether the specified point is inside the shape.
+        /// </summary>
+        /// <param name="x">The x coordinate.</param>
+        /// <param name="y">The y coordinate.</param>
+        /// <returns><c>true</c> if the shape contains the specified point; otherwise, <c>false</c>.</returns>
+        bool Contains(double x, double y);
+
```
VisualMelon: Empty line

VisualMelon (approved):
I think we should merge this. It's non-trivial, so should merge sooner than later so it can be tested before any release.

Still keen to have a nice example, and possibly incorporating a nicer hit-testing API for Axes (per #1419), but unless you think they'd warrant significant changes, I think we should get this in soon.

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -274,6 +274,40 @@ protected virtual void RenderAxisTitle(Axis axis, double titlePosition)
...
+                    areaTopRight = new ScreenPoint(lpt.X - titleSize.Width * 0.5, lpt.Y + titleSize.Height);
+                    areaBottomLeft = new ScreenPoint(lpt.X + titleSize.Width * 0.5, lpt.Y);
+                    break;
+                case AxisPosition.Bottom:
+                    areaTopRight = new ScreenPoint(lpt.x - titleSize.Width * 0.5, lpt.Y);
+                    areaBottomLeft = new ScreenPoint(lpt.X + titleSize.Width * 0.5, lpt.Y - titleSize.Height);
+                    break;
+
+            }
+            // TODO: ensure min area of 5px x X to capture clicks, o this above acc. to axis position
```
Jonarw: Did this TODO get resolved?

Jonarw (commented):
I think it would be important to have at least some example to demonstrate and verify the functionality. I hacked something together very quickly, which can be used for 'normal' Axes:

```
        [Example("AxisArea")]
        public static PlotModel AxisArea()
        {
            var plotModel = FourAxes();
            plotModel.Annotations.Add(new RenderingCapabilities.DelegateAnnotation(context =>
            {
                foreach (var axis in plotModel.Axes)
                {
                    context.DrawRectangle((OxyRect)axis.AxisArea, OxyColors.Undefined, OxyColor.FromAColor(128, OxyColors.Red), 1, EdgeRenderingMode.Automatic);
                    context.DrawRectangle((OxyRect)axis.TitleArea, OxyColors.Undefined, OxyColor.FromAColor(128, OxyColors.Green), 1, EdgeRenderingMode.Automatic);
                    context.DrawRectangle((OxyRect)axis.AxisLineArea, OxyColors.Undefined, OxyColor.FromAColor(128, OxyColors.Blue), 1, EdgeRenderingMode.Automatic);

                }
            }));

            return plotModel;
        }
```

This gives me this output, which seems all in order:
image


If I understand correctly, this should also work for `AngleAxis`, returning an `Annulus`? I tried something based on the `ArchimedeanSpiral` example, but I received `EmptyShape` for the `AxisShape`. Is this intended, or am I misunderstanding something?

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
VisualMelon: @ChrisCC6 this and comment code below could do with being addressed. If you don't have time to work on this let us know and I'll try to take a look into this and @Jonarw 's questions because I'm keen for this PR to go through.

Source/OxyPlot/Rendering/IOxyRegion.cs
ChrisCC6: Empty Line deleted

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
ChrisCC6: @VisualMelon I added two examples based on @Jonarw's code above and added the AxisArea to be set for the MagnitudeAxisRenderer.
The TODO is resolved as well.

VisualMelon: Thanks indeed! Shall try to take a proper look myself later today.

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -274,6 +274,40 @@ protected virtual void RenderAxisTitle(Axis axis, double titlePosition)
 
             var lpt = this.GetAxisTitlePositionAndAlignment(axis, titlePosition, ref angle, ref halign, ref valign);
 
+            var titleSize = this.RenderContext.MeasureText(axis.ActualTitle, axis.ActualTitleFont, axis.TitleFontSize, axis.TitleFontWeight);
```
VisualMelon, line 277: Needs to be `var titleSize = this.RenderContext.MeasureText(axis.ActualTitle, axis.ActualTitleFont, axis.ActualTitleFontSize, axis.TitleFontWeight);` (**Actual**TitleFontSize)

Source/Examples/ExampleLibrary/Axes/AxisAreaExample.cs
```diff
@@ -0,0 +1,158 @@
+﻿using OxyPlot;
```
VisualMelon, line 1: Needs copyright header (see AxisExample.cs for example), and filename should be `AxisAreaExamples.cs`

Source/OxyPlot/Axes/Rendering/AngleAxisRenderer.cs
```diff
@@ -112,9 +119,26 @@ public override void Render(Axis axis, int pass)
                     ha = HorizontalAlignment.Right;
                 }
 
+                maxTextLength = Math.Max(maxTextLength, this.RenderContext.MeasureText(text).Width);
```
VisualMelon, line 122: Needs to be `maxTextLength = Math.Max(maxTextLength, this.RenderContext.MeasureText(text, axis.ActualFont, axis.ActualFontSize, axis.ActualFontWeight).Width);`

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -274,6 +274,51 @@ protected virtual void RenderAxisTitle(Axis axis, double titlePosition)
...
+                    areaBottomLeft = new ScreenPoint(lpt.X + titleSize.Width * 0.5, lpt.Y);
+                    break;
+                case AxisPosition.Bottom:
+                    areaTopRight = new ScreenPoint(lpt.x - titleSize.Width * 0.5, lpt.Y);
+                    areaBottomLeft = new ScreenPoint(lpt.X + titleSize.Width * 0.5, lpt.Y - titleSize.Height);
+                    break;
+
+            }
+
+            // ensure minimum height/width of 5 for the title area , centered to the initial location
```
VisualMelon, line 303: Should lose the space before the comma

Source/OxyPlot/Rendering/ScreenPoint.cs
```diff
@@ -159,5 +159,16 @@ public bool Equals(ScreenPoint other)
...
+        /// <summary>
+        /// Moves the instance for the specified deltas.
+        /// </summary>
+        /// <param name="dx">Delta in x-direction</param>
+        /// <param name="dy">Delta in y-direction</param>
+        public void Move(double dx,double dy)
+        {
+            this.x += dx;
+            this.y += dy;
+        }
```
VisualMelon, line 172: There is discussion about making these structs readonly (mutable structs are a nightmare and all that): would prefer this returned an updated copy (obviously all usages will need to be modified). I'd also prefer the name `Offset`. I'm also going to suggest making this `internal`, because a solution built around `ScreenVector` would make me happier but is definitely out of scope here.

Should be a space after the comma.

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -296,6 +341,9 @@ protected virtual void RenderAxisTitle(Axis axis, double titlePosition)
         /// <param name="drawAxisLine">Draw the axis line if set to <c>true</c>.</param>
         protected virtual void RenderMajorItems(Axis axis, double axisPosition, double titlePosition, bool drawAxisLine)
         {
+            ScreenPoint pt0;
+            ScreenPoint pt1;
```
VisualMelon, line 345: Can we move these guys to just above their usage

Source/OxyPlot/Axes/Rendering/HorizontalAndVerticalAxisRenderer.cs
```diff
@@ -403,6 +455,9 @@ protected virtual void RenderMajorItems(Axis axis, double axisPosition, double t
                     SnapTo(plotAreaBottom, ref transformedValue);
                 }
 
+                string text = axis.FormatValue(value);
+                var labelSize = this.RenderContext.MeasureText(text, axis.ActualFont, axis.ActualFontSize, axis.ActualFontWeight);
```
VisualMelon, line 459: Using the overload that takes an angle and passing it `axis.Angle` (last param) means we can change the block below to the following, which is simpler and makes the code consistent with the automatic margin codes. (There is also a bug in the `Bottom` block at present: should be like `Top` but is like `Left` and `Right`). The behaviour of this still seems a bit funny, but because it matches the margin code it should be left alone in this PR. May be something for us to investigate further another time @Jonarw 
```
                    case AxisPosition.Left:
                        pt = new ScreenPoint(axisPosition + a1 - axis.AxisTickToLabelDistance, transformedValue);
                        this.GetRotatedAlignments(axis.Angle, -90, out ha, out va);
                        labelEndPosition = Math.Min(labelEndPosition, pt.x - labelSize.Width);
                        break;
                    case AxisPosition.Right:
                        pt = new ScreenPoint(axisPosition + a1 + axis.AxisTickToLabelDistance, transformedValue);
                        this.GetRotatedAlignments(axis.Angle, 90, out ha, out va);
                        labelEndPosition = Math.Max(labelEndPosition, pt.x + labelSize.Width);
                        break;

                    case AxisPosition.Top:
                        pt = new ScreenPoint(transformedValue, axisPosition + a1 - axis.AxisTickToLabelDistance);
                        this.GetRotatedAlignments(axis.Angle, 0, out ha, out va);
                        labelEndPosition = Math.Min(labelEndPosition, pt.y - labelSize.Height);
                        break;
                    case AxisPosition.Bottom:
                        pt = new ScreenPoint(transformedValue, axisPosition + a1 + axis.AxisTickToLabelDistance);
                        this.GetRotatedAlignments(axis.Angle, -180, out ha, out va);
                        labelEndPosition = Math.Max(labelEndPosition, pt.y + labelSize.Height);
                        break;
```

VisualMelon (changes requested):
Found a few bugs; fixes given in inline comments.

Missing implementation for `AngleAxisFullPlotArea` and `MagnitudeAxisFullPlotArea` in `MagnitudeAxisFullPlotAreaRenderer`. `AngleAxis` should be much the same, but `Magnitude Axis` would need to be a rect with a hole in it. We could create a general-purpose 'difference' shape, or a specific shape which is a rect with a rect cut out of it (somehow I suspect this doesn't have a familiar name). (I'd be happy to leave this to another PR to avoid another round of discussion here if it's OK with @Jonarw )

There are issues with `AngleAxis.Angle` property (30 in this example), but I don't think they are specific to this PR so shouldn't hold it up.
image.

If possible, could you rebase against `develop` so that the CI is happy to run again (should just be changes in the changelog)

Source/Examples/WPF/WpfExamples/App.xaml
```diff
@@ -1,7 +1,7 @@
 ﻿<Application x:Class="WpfExamples.App"
              xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
              xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-             StartupUri="MainWindow.xaml">
+             StartupUri="Examples/AxisAreaDemo/MainWindow.xaml">
```
VisualMelon, line 4: Looks like this file didn't make it into the commit

PR #2001, opened 2023-05-14
 Make consistent BaseValue and BaseLine across BarSeries, LinearBarSeries, and HistogramSeries

VisualMelon:
Having merged #1942 (from which all of this work is derived), this makes the BaseValue and BaseLine consistent across all the 'bar-type' series, and exposes the `ActualBaseLine` as a `{ public get; protected set; }` property.

### Checklist

### Changes
 - All 3 series have consistent support for negative values, determined by a `BaseValue` that defaults to zero.
 - Fixes issues with rendering against log axes by explicit detection and use of `YAxis.ClipMinimum` to clamp to end of plot
 - All 3 series have consistent support for a `BaseLine` which is included in the Y-axis: important for log-axes which extend to infinity that they have a sensisble 'reference' baseline. When `NaN`, actual value is derived for log-axes by dividing the smallest value by 10 (to give some padding to log plots: better ideas welcome (e.g. depend on range of values and base of axis?): computed values is exposed as `ActualBaseLine`

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Series/BarSeries/BarSeries.cs
+++ b/Source/OxyPlot/Series/BarSeries/BarSeries.cs
@@ -467,8 +467,8 @@ namespace OxyPlot.Series
                     this.Manager.SetCurrentBaseValue(stackIndex, categoryIndex, value < 0, topValue);
                 }
 
-                var clampBase = this.YAxis.IsLogarithmic() && !this.YAxis.IsValidValue(baseValue);
-                var p1 = this.Transform(clampBase ? this.YAxis.ClipMinimum : baseValue, categoryValue);
+                var clampBase = this.XAxis.IsLogarithmic() && !this.XAxis.IsValidValue(baseValue);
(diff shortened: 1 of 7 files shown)


Source/OxyPlot/Series/BarSeries/BarSeries.cs
```diff
@@ -295,14 +344,14 @@ protected override bool IsValid(BarItem item)
...
-        /// <param name="baseValue">The bar item base value.</param>
+        /// <param name="BaseValue">The bar item base value.</param>
         /// <param name="topValue">The bar item top value.</param>
         /// <param name="categoryValue">The bar item category value.</param>
         /// <param name="categoryEndValue">The bar item category end value.</param>
         protected void RenderLabel(
             IRenderContext rc,
             BarItem item,
-            double baseValue,
+            double BaseValue,
```
Jonarw: The capital 'B' is probably an accident?

VisualMelon: Fixed, and rebased on develop

PR #1920, opened 2022-08-16
Tidy up, modernise code and VS cleanup

BobLd:
### Checklist
- [x] I have included examples or tests (no need)
- [x] I have updated the change log
- [x] I am listed in the CONTRIBUTORS file
- [x] I have cleaned up the commit history (use rebase and squash)

### Changes proposed in this pull request:
- Tidy up, modernise code and VS cleanup (was previously in #1836)

@oxyplot/admins

PR #2016, opened 2023-07-19
Removed nesting levels in several places through block-scoped `using`s

Lehonti:
This is just reformatting for increased readability, so it should be quite safe

### DIFF
--- a/Source/Examples/ExampleGenerator/Program.cs
+++ b/Source/Examples/ExampleGenerator/Program.cs
@@ -90,11 +90,9 @@ namespace ExampleGenerator
             {
                 fileName = Path.ChangeExtension(fileName, ".pdf");
                 Console.WriteLine(fileName);
-                using (var stream = File.Create(fileName))
-                {
-                    var exporter = new PdfExporter { Width = 600d * 72 / 96, Height = 400d * 72 / 96 };
-                    exporter.Export(model, stream);
-                }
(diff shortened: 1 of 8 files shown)


Jonarw:
In general I think this is a good idea, and I am certainly not opposed to it. However I was planning to work on consolidating our code style (as a continuation of #1920) based on an updated .editorconfig in the near future, and I think it might be nice to have all style changes in one PR then.

Lehonti:
I understand. This is the corresponding rule:

https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0063

Not sure if it's a good idea to have the IDE enforce that rule, as in some cases the old syntax may be preferred.

VisualMelon:
I'm also not opposed to these specific changes, and I agree that in some cases the old style may be preferred (obviously you've identified situations where the scope is unchanged, which presumably the editor rule will not). I'd be happy to leave this unspecified in the editor config and for us to decide on a case-by-case basis, but I don't have any strong opinions about this.

Jonarw:
Yes, actually the automatic refactoring in VS based in the .editorconfig rule will only highlight and change `using`-blocks where this will not change the scope.

VisualMelon:
In that case, I agree we should hold this and do everything all at once with an agreeable editor config.

Issue #2017, opened 2023-07-21
Consolidating Code Style


Jonarw:
This issue is meant to be a follow-up to #1920. I think it is a good idea in general to consolidate our code base in terms of code style. Additionally, I think it would also be good to make it easier for future contributions to adhere to our style conventions - which mostly means that VS should automatically highlight places where these are violated.

Therefore, I would like to start the discussion by updating our .editorconfig file with more rules (will make a Draft PR soon).
This is obviously heavily influenced by my personal preferences, though I think must stuff is in line with the more or less unwritten conventions we had until now. But of course everything is up for debate!

After initial discussion on the .editorconfig, I would start by applying the rules to a portion of the codebase to get a feel on how it will look. Once we're happy, we can implement the changes on the whole code base.

A couple of comments on specific code style aspects:

- File-scoped namespaces
Generally I am a fan of these, as it saves one level of indentation pretty much everywhere. However, this change would result in changes to pretty much every line of code. Therefore I would propose to leave this as-is for now, and maybe make this into an isolated follow-up PR after the rest of the style changes are finished.

- `<inheritdoc />`
I would propose to change the documentation to `<inheritdoc />` where this is obviously possible (e.g. overridden functions). This requires a bit of manual work, but I think makes code more readable overall.

- New Language features
Do we want to upgrade to a newer Language version to be able to take advantage of some of the new features? Notable style-related things would be `new()`, pattern matching, `default` and probably some other stuff I am not thinking of right now. Adding PolySharp (as mentioned by HavenDV recently) would also enable us to use things like `Index`es and `ValueTuple`s, which would be nice I guess.

VisualMelon:
Thanks for writing this up properly. On those three points:

 - I'd agree we should leave the namespaces as-is for the time being.
 - I'm not convinced `<inheritdoc />` makes things more readable, but I do think it should be used when there is no additional information to be provided in an implementing/inheriting method. If this can't be automated reliably, then I think it should be considered low priority.
 - Based on the recent discussion, I'm happy for us to bump the C# version, but equally not fussed if it's deemed problematic at all. I'd oppose adding PolySharp, because it's another third-party dependency we force onto consumers... and without `ValueTuple`, I don't have to argue against the use of `ValueTuple` every time someone uses it in a way I don't like ;)

Jonarw:
[...]

PR #2018, opened 2023-07-21
Consolidating Code Style

Jonarw:
See Discussion in #2017.

### Checklist

- For now: Add an .editorconfig file as a discussion starter

Also adding a couple of screenshots for a more convenient display of the settings:
image
image
image

@oxyplot/admins


VisualMelon:
Initial thoughts:
 - We might want to make "require `this`" more than a suggestion, as @objorke was always pretty strict on this
 - I'm not wild about the `var` sugestions. In particular, I think we should disable the "For built-in types" and "Elsewhere" suggestions, so it's only a suggestion "where apparent".
 - As long as `ValueTuple` isn't available, we should disable "Prefer tuple swap"
 - I'd prefer we didn't have the "Use expression body for accessors" suggestion

I made some of these changes in the GUI editor, and it seems to have largely re-written the editor config, so not sure what's up with that, but pushed to https://github.com/VisualMelon/oxyplot/tree/meloneditorconfig if of any interest

Jonarw:
> We might want to make "require this" more than a suggestion, as @objorke was always pretty strict on this

The reason I would be reluctant to make any code-style rule more than a suggestion is that we have `TreatWarningsAsErrors` active on some projects. I wouldn't want style issues to prevent the code from compiling - I think it is a reasonable workflow to make code changes in any code style you are used to, test it, and apply automated code style fixes at the end.

> I'm not wild about the var sugestions. In particular, I think we should disable the "For built-in types" and "Elsewhere" suggestions, so it's only a suggestion "where apparent".

Fine with me. Personally I'm used to always using `var`, but I can see the advantages of your proposal, so let's do it like this.

> As long as ValueTuple isn't available, we should disable "Prefer tuple swap"

Agree, if we decide against PolySharp, we should disable this.

> I'd prefer we didn't have the "Use expression body for accessors" suggestion

I slightly prefer to have expression bodies for accessors, as it appears a bit less 'noisy' to me without the brackets. But I would also be fine with disabling this one.

VisualMelon:
Re. `this`, that's a good point: better leave it as-is.

Re. `var`, realistically I wouldn't mind most uses, but it's still something that I think warrants human judgement in all but the most trivial cases, and I'm generally of the opinion that example code should avoid `var` (though we could always relax the rules for the example library). I also really don't like `for (var i = ...` but that's probably just me.

To provide a little more explanation about expression bodied accessors, I don't mind `get` at all, but I'm not a big fan of using it with 'set', though it's not something I have particularly strong feelings about.

Jonarw:
Did the changes as discussed so far (and rebased, because apparently I started out not quite up-to-date).
I originally created the file in the Options dialog, now I edited it using the VS's settings editor - I assume that is why there seem to be a couple more changes in the file. I assume we will use the settings editor going forward, as this seems more complete.

Jonarw:
One advantage of enforcing expression-bodied accessors (where possible) would be that properties like this:
image
would be collapsed to this:
image
which I find much better in this case. Unfortunately it does not seem to be possible to just collapse the whitespace automatically.

Granted, this only affects cases where the property can not be converted to an auto property, as it is the case in this example. I'm not sure how many of these we have. But for me this would be a slight argument for expression-bodied accessors.

VisualMelon:
@Jonarw I'm perhaps not typical in not being bothered by this sort of verbosity, so I don't "see" the advantages, but the `set` notation does irk me. If you think it's worthwhile, then I'd say go with your instinct, because it's probably closer to the general consensus.

Jonarw:
After clicking through some code files, I did not find any other instances like this - usually properties like this are just converted to auto properties. So in the end I guess it does not matter too much which style we choose to use here.

However what I did realize is: I think we should avoid ambiguities in the style rules as far as possible. Firstly, I think it's good to be consistent, after all the purpose of this PR should be to make it easy for contributors to follow our style guidelines. 
Secondly, not having consistent code style rules kind of prevents the use of project-wide automated code style fixes in VS. For example, when you instruct VS to apply the rules for `var` usage project-wide, it will also apply the `var` rules with `silent` severity. So we should really only have `silent` on rules where we actually do not care, and therefore won't ever want to do project-wide automated style fixes.

So I think we have two areas where it would be good to remove ambiguities (i.e. change the severity to more than `silent`). 

- Expression-bodied accessors
My suggestion would be to tentatively enable them. Once the .editorconfig is fixed, I will apply it to the core library so we can see how it looks. If you still hate it then I'd also be happy to change it to block bodies :)
- `var` usage
As mentioned, personally I just always use `var`. However I can see the advantages of writing the explicit type in some instances.
  - Where type is apparent
  I think here we agree to use `var`
  - On built-in types
  I can see the advantages of using the explicit types here. I think it might be slightly verbose at times, but I think I'd be ok with not using `var` on built-in types
  - Elsewhere
  This is where it becomes a bit tricky. I can see how it would be beneficial in some cases to write the explicit type here, but in many cases I think this is needlessly verbose. Not saying that I couldn't be convinced otherwise, but I think I would prefer enforcing `var` here.

VisualMelon:
Re. Expression-bodied accessors, fine by me.

> For example, when you instruct VS to apply the rules for var usage project-wide, it will also apply the var rules with silent severity.

I can't reproduce this using the code-cleanup function in VS: it rewrites vars when I set them as suggestion, but not as disabled (silent). How are you running the code style fixes? If this true, then I'll unhappy agree with need to enforce something (to compensate for the tooling), but I'm hoping that we don't need to do so for built-in types and elsewhere.

Jonarw:
In this line we would have a violation of our `var` style rules with the new .editorconfig:
https://github.com/oxyplot/oxyplot/blob/816120cc256491fa16406e08a9594713d667408e/Source/OxyPlot/Axes/Rendering/AngleAxisFullPlotAreaRenderer.cs#L169

VS indicates this violation and offers an automatic fix with the little light bulb. When doing this fix and choosing to apply it to the whole project, VS will also apply the other `var` rules that are on `silent` for the whole project.
grafik

Independently of the tooling issue, I think there could still be some benefit of clear and unambiguous style rules, however we may define them. Sure, we would lose some leeway that might allow a human to write slightly more readable code by choosing when to use `var` and when not - but then again I guess this can be quite subjective.
Maybe I am overthinking this, and this is not that big of a deal - let me know if you think so :)
But consider this scenario: Assume we would merge the .editorconfig as it is right now, so allowing either `var` or concrete types in most cases. Then someone creates a new PR, adding some code that uses `var` everywhere. Someone else creates a PR that only uses concrete types (apart from the cases where VS thinks the type is apparent). Both PRs would be 'legal' according to our style rules. How would we proceed? Would we
- a) merge both PRs without requesting further changes?
This would lead to a somewhat inconsistent code style across the repo - kind of like the situation we have right now. Isn't this what we would like to fix with the .editorconfig rules?
- b) request changes from one or both authors regarding the usage of `var`?
But how do we determine which lines of code warrant changes? After all, this is somewhat subjective. Will we have additional written or unwritten rules for when to use `var` and when not?

Please don't get me wrong, I do think you have a valid point. I am just not quite sure how this will work out, but, as mentioned, I may be overthinking this. Maybe I'm also not quite typical in this regard, in that I'm not bothered by using `var` wherever possible.

VisualMelon:
> VS will also apply the other var rules that are on silent for the whole project.

That is really rather annoying...

> Maybe I am overthinking this, and this is not that big of a deal

That's pretty much my opinion about non-public-API things in general ;)

> How would we proceed

Assuming both are readable, I'd go with (a). Equally, both could be less readable in one place or another due to the use/lack of use of `var`, and I'd prefer to be free to dictate that on an instance-by-instace basis. I wouldn't want to be instructing either to refit the code blindly either way. `var` can certainly make certain code read better and avoid certain classes of bug, but I think it can also make code harder to follow (particularly example code, but again, we don't have to have the same convention in the example library).

> Isn't this what we would like to fix with the .editorconfig rules?

`var`/not-`var` just doesn't seem like something that warrants a convention to me. Frankly, most of this stuff I don't really care about, but code does look nice if the formatting is consistent, and having the editor enforce a convention stops people worrying about the conventions which makes contributing easier. `var`/not-`var` just seems fundamentally different to me than e.g. where to put white-space and braces.  That is to say, `var` doesn't for me fit in the box of things where no-one really cares what the convention is, just that there is a convention. Ane even when there is a convention, I'm not one who would rigidly enforce them (e.g. I'd be in favour of an upper-case parameter name if it was a standard mathematical symbol etc.).

Anyhow, I do think most people would agree more with you than me (there are certainly large .NET projects which suggest `var` everywhere), so I'm happy to go with suggesting `var` everywhere.

Jonarw:
To get an idea of what the most common consensus on this topic might be (if there is any) I checked the .editorconfigs of a couple C# repositories that came to my mind. This is in no way meant to be a representative selection or anything, just stuff that is somehow related to oxyplot or that I use at work. I also don't know how strictly these editorconfigs are adhered to, I just noted down what was specified.

- **ImageSharp**: Never use var, warning
- **SkiaSharp**: Always use var, suggestion
- **Avalonia**: Always use var, suggestion
- **Roslyn**: Always use var, suggestion
- **Scottplot**: No rules about var in editorconfig
- **TheAlgorithms**: Always use var, suggestion
- **Fody**: Always use var, error

From this arbitrary selection, there is some precedent for leaving `var` usage up to the individual (ScottPlot), however most repos seem to have some tendency either for or against `var` usage in general. Of course this proves nothing, just found it useful to get some idea of how other people handle this.

From my previous statements I guess it's clear that I wouldn't have a problem with suggesting `var` everywhere, but I wouldn't want to do it if you're not quite convinced, I could also live with leaving it unspecified.

VisualMelon:
Didn't know about TheAlgorithms: shall have to look into that (though I see some tuples as inputs in the public APIs...!). I would otherwise consider them representative of significant and respected C# projects (and a few more than I checked in my brief search).

I'm happy with whatever you think: I'm known for my extreme opinions on these things, and it seems there is more of a tendency to use `var` than not. We probably don't have to worry too much about the "apply to whole project" thing if we don't specify 'everywhere' because we can just reject PRs that make random changes to var in different places if we don't like them. I'd certainly be happy to not have any rules; I think we'd both object to enforcing not-`var` everywhere; and I'll live with `var` everywhere.

Jonarw:
I applied a round of automated style fixes to the core library in the commit I just pushed.
I decided to go for the always-`var` approach for now is it seems to be the most common as far as I have seen.

Now that we have some actual diffs, we can of course still further discuss the .editorconfig. In most cases it should be pretty easy to change the style rule and apply it to the project in case we want to change something.

Once we're happy, I will apply the style rules to the rest of the repo.

Source/OxyPlot/Annotations/FunctionAnnotation.cs
```diff
@@ -70,10 +70,10 @@ protected override IList<ScreenPoint> GetScreenPoints()
 
             if (fx != null)
             {
-                double x = this.ActualMinimumX;
+                var x = this.ActualMinimumX;
 
                 // todo: the step size should be adaptive
-                double dx = (this.ActualMaximumX - this.ActualMinimumX) / this.Resolution;
+                var dx = (this.ActualMaximumX - this.ActualMinimumX) / this.Resolution;
```
VisualMelon, line 76: Stuff like this a harrowing: at a glance, I have no idea if this is a real or integer division.

Jonarw:
Also worth mentioning:
I changed the `LangVersion` to 11 and added PolySharp after verifying that it does in fact not add any dependencies for package consumers. Everythings seems to work fine on .NET 4.6.2, including things like `Index`es.

Source/OxyPlot/PlotModel/PlotModel.cs
```diff
@@ -674,7 +637,7 @@ public OxyColor GetDefaultColor()
         /// <returns>The next default line style.</returns>
         public LineStyle GetDefaultLineStyle()
         {
-            return (LineStyle)((this.currentColorIndex / this.DefaultColors.Count) % (int)LineStyle.None);
+            return (LineStyle)(this.currentColorIndex / this.DefaultColors.Count % (int)LineStyle.None);
```
VisualMelon: I don't like this at all: it reads badly and almost no-one knows the precedence rules for `%`.

Source/OxyPlot/Rendering/OxyColor.cs
```diff
@@ -216,7 +192,7 @@ public static OxyColor FromHsv(double[] hsv)
         public static OxyColor FromHsv(double hue, double sat, double val)
         {
             double g, b;
-            double r = g = b = 0;
+            var r = g = b = 0;
```
VisualMelon: This is hideous, though not all `var`'s fault. Should change these to straight declarations, and change the switch further down to throw for unexpected values of `i`.

Source/OxyPlot/Rendering/OxyPen.cs
```diff
@@ -108,7 +102,7 @@ public override int GetHashCode()
         {
             unchecked
             {
-                int result = this.Color.GetHashCode();
+                var result = this.Color.GetHashCode();
```
VisualMelon, line 105: Not a problem, but we should consider using `HashCode.Combine` for this sort of thing.

Source/OxyPlot/Rendering/OxyRect.cs
```diff
@@ -16,12 +16,12 @@ namespace OxyPlot
     /// <summary>
     /// Describes the width, height, and point origin of a rectangle.
     /// </summary>
-    public struct OxyRect : IFormattable, IEquatable<OxyRect>
+    public readonly struct OxyRect : IFormattable, IEquatable<OxyRect>
```
VisualMelon, line 19: This may have perf implications, but should be fine.

Source/OxyPlot/Series/AreaSeries.cs
```diff
@@ -153,11 +125,12 @@ protected List<DataPoint> ActualPoints2
         public override TrackerHitResult GetNearestPoint(ScreenPoint point, bool interpolate)
         {
             var xy = this.InverseTransform(point);
-            var targetX = xy.X;
-            int startIdx = this.IsXMonotonic 
+
+            _ = xy.X;
```
VisualMelon: Can delete this line

Source/OxyPlot/Series/BarSeries/ErrorBarSeries.cs
```diff
@@ -66,13 +66,13 @@ protected internal override void UpdateMaxMin()
                 var labels = this.GetCategoryAxis().ActualLabels;
                 for (var i = 0; i < labels.Count; i++)
                 {
-                    int j = 0;
+                    var j = 0;
                     var items = this.ValidItems.Where(item => item.GetCategoryIndex(j++) == i).ToList();
```
VisualMelon, line 70: This is pretty confusing (not the `var`, the line below it)

Source/OxyPlot/Series/FinancialSeries/HighLowSeries.cs
```diff
@@ -97,13 +88,7 @@ public OxyColor ActualColor
...
         /// </summary>
         /// <value>The items.</value>
-        public List<HighLowItem> Items
-        {
-            get
-            {
-                return this.items;
-            }
-        }
+        public List<HighLowItem> Items { get; } = new List<HighLowItem>();
```
VisualMelon: Weird it didn't `new()` this; am I missing something?

Source/OxyPlot/Series/PieSeries.cs
```diff
@@ -258,51 +241,51 @@ public override void Render(IRenderContext rc)
...
-            this.total = this.slices.Sum(slice => slice.Value);
+            this.total = this.Slices.Sum(slice => slice.Value);
             if (Math.Abs(this.total) <= 0)
             {
                 return;
             }
 
             // todo: reduce available size due to the labels
-            double radius = Math.Min(this.PlotModel.PlotArea.Width, this.PlotModel.PlotArea.Height) / 2;
+            var radius = Math.Min(this.PlotModel.PlotArea.Width, this.PlotModel.PlotArea.Height) / 2;
```
VisualMelon, line 251: This whole block just more daunting without the types

VisualMelon (commented):
Just a random sample of comments (not all about the changes); not looked at everything. I'll probably eventually get used to `for (var i...)`, but there are a couple of places in here where I think `var` makes things worse, so can perhaps use those as a talking point. I'm also not wild about some of the parenthesis loss, though mostly this all looks fine to me.

Source/OxyPlot/Series/BarSeries/ErrorBarSeries.cs
Jonarw: I agree, but I deliberately wanted to avoid any code changes not directly related to .editorconfig rules in this PR

Source/OxyPlot/Series/FinancialSeries/HighLowSeries.cs
Jonarw: I seems that `new()` is not applied to property initializations, but only to field initializations. Seems a bit weird to me as well...

Source/OxyPlot/Rendering/OxyRect.cs
Jonarw: If anything it should improve performance as far as I know (because it potentially avoids defensive copies). But I doubt this will be noticeable.

Source/OxyPlot/Rendering/OxyPen.cs
Jonarw: Yes, fully agree. But this is again something I am not sure we should do as part of this PR, as it could potentially change behavior in some edge cases.

Source/OxyPlot/Rendering/OxyColor.cs
Jonarw: Agree about the hideousness, I will change that.
Arguably it could make more sense to check `hue` for valid range at the beginning of the function - but again, I would like to avoid behavior changes in this PR (even if it's fixing stuff).

Source/OxyPlot/PlotModel/PlotModel.cs
Jonarw: Should we set the parentheses preferences for arithmetic operators to 'Always for clarity'? I think I wouldn't mind that in most cases.

Source/OxyPlot/Annotations/FunctionAnnotation.cs
Jonarw: Counter-Point: Beginning the line with `double` does not guarantee that a floating-point division is performed. If `ActualMaximumX` and `ActualMinimumX` were `int`s, an integer division would be performed and the result promoted to `double`. I would argue that in this particular case the explicit type could even be misleading.

Source/OxyPlot/Series/PieSeries.cs
Jonarw: I can definitely see where you are coming from in this case.
Personally I don't mind it - I am just used to hovering over the `var` with the mouse pointer whenever I am in doubt about a type, and usually you can safely assume that blocks of math will just be `double` (or `float` if you are working in single precision).
But I can see how more explicitness may be beneficial here...

Source/OxyPlot/Annotations/FunctionAnnotation.cs
VisualMelon: No, you're right, `x` isn't actually used. I'm blaming `var` for something that isn't its fault.

Source/OxyPlot/PlotModel/PlotModel.cs
VisualMelon: Again, it's something where I'd prefer to leave it to the programmer to decide. I don't mind removing parens for `x * y + z` etc. because it's well known, but for less well known operators I'd prefer it to be discretionary.

Source/OxyPlot/Rendering/OxyRect.cs
VisualMelon: I've seen significant performance degradation with `readonly` structs in the past, but that was years ago (netcore 2.x era I think) and as you say it should hopefully go the other way around.

Source/OxyPlot/Series/PieSeries.cs
VisualMelon: Yeah, with tooling it's certainly easier to interrogate, but it's the sort of thing where when scanning it's easy to be suspicious of every line after the first few because the types are unclear and so it distracts from the code itself (maybe that's just me). I'll live with it.

Source/OxyPlot/Series/FinancialSeries/HighLowSeries.cs
Jonarw: I changed this to new(), and also some other places I found. Probably did not get all of them, but shouldn't matter too much I would think.

Source/OxyPlot/Series/AreaSeries.cs
Jonarw: Done

Source/OxyPlot/Rendering/OxyColor.cs
Jonarw: Changed these declarations for something more sensible

Source/OxyPlot/PlotModel/PlotModel.cs
Jonarw: We actually have the rule set up so it's up to the programmer (we have it on `silent`), so it will not give a suggestion when 'unnecessary' braces are added. This was again a case of the VS refactoring being somewhat eager

Jonarw:
I think I would be about done now. I applied the style rules to all projects except the Examples. I think we could leave them be for now.
Also made the indentation of project files consistent - we had a mix of 2-spaces and 4-spaces before.

I think the changes should be quite safe, I guess the only potential for breaking stuff should be around conditional compilations.

@VisualMelon Could you glance over the changes and check if there is anything you don't like at all?
I think there are actually not many instances of expression body syntax being applied to `set`ters, as often they are more than 1 line.
Regarding `var` - let me know how unhappy you are ;) If you would prefer, I would be fine with putting the `var` style rule on silent and reverting the `var`ification in select places.

VisualMelon:
@Jonarw I'll try to take a proper look at this on the weekend

Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
```diff
@@ -126,7 +126,7 @@ protected PlotViewBase()
         /// <summary>
         /// Gets the coordinates of the client area of the view.
         /// </summary>
-        public OxyRect ClientArea => new OxyRect(0, 0, this.ActualWidth, this.ActualHeight);
+        public OxyRect ClientArea => new(0, 0, this.ActualWidth, this.ActualHeight);
```
VisualMelon, line 129: I don't like this

VisualMelon (commented):
From my side, I think it's really a question of what the downsides of not forcing `var` are (again assuming no-one wants to force not-`var`). The more I think about it, the less I'm worried about leaving it alone, so I would prefer to leave it up to the dev, but having scrolled through most (not all) of this there is nothing (except `for (var i...)` that makes me really sad now.

I'm a bit unhappy about target-type new, which caught my eye in a number of places. Thinking about it for more than 10 seconds, I'd prefer that was unspecified as well for the time being if that isn't objectionable.

To try to avoid going in circles, can I suggest we leave `var` and target-type `new` unqualified for now (which should make this diff somewhat smaller and easier to work through), get this merged (so we can start to see the benefits where we agree), and then open another MR for `var` where I'm not allowed to complain endlessly?

Source/OxyPlot.Wpf.Shared/PlotViewBase.cs
Jonarw: For me, the upside of forcing `var` would be mainly that there is a clear rule that is easy to remember and follow. Leaving it up to the individual will lead to some parts of the code base where `var` is used extensively, and others where it isn't. Of course really this isn't a problem at all, just slightly triggers my obsessive-compulsive tendencies ;)

I'd be fine with leaving `var` and target-typed new unspecified for now. I'm not quite sure if I can revert just the changes to `var` and `new` and still keep the rest of the refactorings... In the worst case I'll have to revert everything and start from the beginning. I'll have to look into this, probably won't manage before next week.

Jonarw: Sorry, did not really find time to continue here. Also lacking a bit of motivation, as it does seem that I pretty much would have to start over - I did not find an easy way to revert just the changes relating to `var`.

I will be away for a couple of weeks starting this weekend, and probably will not be able to do anything during this time. If you, or someone else would like to pick this up this is perfectly fine by me, otherwise I will take another look in ~4 weeks.

Jonarw:
I am sorry, but after looking for some time, I did not find motivation to continue with this. As I didn't find a way to revert only the changes related to `var`, basically I'd have to revert everything and start over with a modified .editorconfig. Applying the automated code fixes doesn't sound so bad at first, but between making sure all the fixes are applied correctly for all target frameworks and taking into account conditional compilations this is somewhat tedious.
I would rather focus on other stuff and would like to get this out of my head, so I will just close this PR for now. If someone else would like to pick it up, feel free.

VisualMelon:
@Jonarw sorry it turned into such a nightmare (wherefore I'm wholly responsible)

PR #2020, opened 2023-08-03
private, non-virtual methods that don't access any instance members were made `static`

Lehonti:
### Checklist

- [ ] I have included examples or tests (N/A)

@oxyplot/admins

### DIFF
--- a/Source/Examples/WPF/ExampleBrowser/MainWindowViewModel.cs
+++ b/Source/Examples/WPF/ExampleBrowser/MainWindowViewModel.cs
@@ -100,7 +100,7 @@ namespace ExampleBrowser
             this.RaisePropertyChanged(nameof(this.CanvasModel));
         }
 
-        public static IEnumerable<Renderer> Renderers => Enum.GetValues(typeof(Renderer)).Cast<Renderer>();
+        public IEnumerable<Renderer> Renderers => Enum.GetValues(typeof(Renderer)).Cast<Renderer>();
 
         public ExampleInfo SelectedExample
         {
--- a/Source/Examples/WPF/ExampleBrowser/MainWindowViewModel.cs
+++ b/Source/Examples/WPF/ExampleBrowser/MainWindowViewModel.cs
@@ -100,7 +100,7 @@ namespace ExampleBrowser
             this.RaisePropertyChanged(nameof(this.CanvasModel));
         }
 
-        public IEnumerable<Renderer> Renderers => Enum.GetValues(typeof(Renderer)).Cast<Renderer>();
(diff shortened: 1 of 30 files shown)


Jonarw:
There appears to be some comments leftover by the automated refactoring. Also in some cases the access to the static methods needlessly states the class name.
In any case, I think making things `static` might warrant a bit more thought in some cases, so I would be hesitant to merge this even with these issues resolved.

VisualMelon:
I agree with @Jonarw ; just because the methods can be made static without changing the public API doesn't mean they should be. There a few methods in here that probably _should_ be static (and perhaps even moved somewhere more accessible, e.g. `MovingAverage`) but I'm of the opinion this should be considered on a case-by-case basis.

Lehonti:
@Jonarw you are right about the comments that were left over from the automated refactoring. I just removed them

@Jonarw  and @VisualMelon , in my opinion this is a bit like the `readonly` qualifier. It _ascertains_ at the language level something that is already there, and you can keep those guarantees in mind in order to code (and refactor) more safely. If you ever need to mutate a `readonly` field or need a `static` method to access instance members, you just remove those qualifiers. In any case, if the qualifiers aren't there in the first place, it's harder to know.

But what do you think?

Source/Examples/WPF/ExampleBrowser/MainWindowViewModel.cs
```diff
@@ -100,7 +100,7 @@ private void CoerceRenderer()
             this.RaisePropertyChanged(nameof(this.CanvasModel));
         }
 
-        public IEnumerable<Renderer> Renderers => Enum.GetValues(typeof(Renderer)).Cast<Renderer>();
+        public static IEnumerable<Renderer> Renderers => Enum.GetValues(typeof(Renderer)).Cast<Renderer>();
```
Jonarw: This change breaks the WPF ExampleBrowser - the property is used by a data binding, which does not work if the property is static. That is an example of why I think that making things `static` should be considered case by case, and not done with an automatic solution-wide refactoring.

Lehonti: That's a fair argument. I'm removing that one now. I thought the automated refactoring would only change methods, not properties

Lehonti: And yeah, automated refactoring could conflict with tools that rely on reflection, so I understand what you mean

Jonarw:
There are also other points to consider (e.g. making a `protected` method `static` will be a breaking change for derived classes). Some of the instances where this refactoring is completely safe should be covered in #2018 anyway, so I will close this.

PR #2039, opened 2023-09-12
close #2038 BarSeries hit test issue fixed

### LINKED ISSUE #2038: BarSeriesBase: HitTest does not deliver the correct BarItem, when there are invalid items in Data
### Steps to reproduce

1. Use the following PlotModel
```cs
var model = new PlotModel();
model.MouseDown += (s, e) =>
{
	if (e.ChangedButton == OxyMouseButton.Left)
	{
		foreach (var hit in model.HitTest(new HitTestArguments(e.Position, 0)))
		{
			if (hit.Item is BarItem barItem)
			{
				// change the color of the clicked Bar Item to
				barItem.Color = OxyColors.Red;
				model.InvalidatePlot(false);
			}
		}
	}
};
var s1 = new BarSeries { Title = "Series 1", IsStacked = true, StrokeColor = OxyColors.Black, StrokeThickness = 1, StackGroup = "3" };
int i = 0;
s1.Items.Add(new BarItem { Value = 25, CategoryIndex = i++ });
s1.Items.Add(new BarItem { Value = double.NaN, CategoryIndex = i++ }); // add one invalid item
s1.Items.Add(new BarItem { Value = 18, CategoryIndex = i++ });
s1.Items.Add(new BarItem { Value = 40, CategoryIndex = i++ });
model.Series.Add(s1);
```
2. Now Click on the top item, the `HistTest` returns the wrong item

Platform:
.NET version: All

### Expected behaviour

The HitTest should return the correct item. Thus the clicked BarItem should change color!

### Actual behaviour

The wrong item changes color.
image

### Problem description

I found the source of the misbehaviour in `BarSeriesBase.UpdateValidData` when the index in the reverse lookup dictionary is not correctly calculated.

I will provide a merge request to fix this bug.

belucha:
- add demonstration code
- fix the problem, by correctly setting the index of the ValidItemsIndexInversion Dictionary

Fixes #2038 

### Checklist

None.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 ## Unreleased
 
 ### Added
+- Example to demonstrate BarSeries HitTest bug (#2038)
 - Example to Show/Hide Legend (#1470)
 - Example of BarSeries stacked and with labels (#1979)
 - Example of issue with AreaSeries tracker (#1982)
@@ -29,6 +30,7 @@ All notable changes to this project will be documented in this file.
 - Unused LabelColor property from TornadoBarSeries, IntervalBarSeries, and RectangleBarSeries (#2030)
 
 ### Fixed
+- fixed issue with BarSeries, when the HitTest returns the wrong BarItem, when there are invalid items in the list. (#2038)
 - Placement of BarSeries labels when stacked (#1979)
 - SystemInvalidException in LineSeries when only Double.Nan values are added (#1991)
 - Issue with tracking AreaSeries with monotonic data points (#1982)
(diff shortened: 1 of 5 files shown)


VisualMelon:
Fix looks good to me!

Very boring, but bug examples should be in the `ExampleLibrary.Issues` class (and in this case marked closed because it is fixed): can you move it to `Issues` and add a `SubTitle` explaining the expected behaviour so that when we come back to in future we can be sure the issue hasn't returned.

belucha:
@VisualMelon thanks for the fast reply. I've made the changes as requested.

Regards Daniel

VisualMelon (approved):

VisualMelon:
All looks good to me; won't merge just yet so that the other maintainers have an opportunity to look it over.

PR #2086, opened 2024-06-24
Adding an example how to synchronize several different Y-Axis of the same PlotModel

flo-so:
Update CHANGELOG.md
Update CONTRIBUTORS
Create SynchronizedYAxisDemo Folder
Create MainWindow.xaml.cs
Create MainWindow.xaml

### Checklist

- Adding an example how to synchronize several different Y-Axis of the same PlotModel

@oxyplot/admins


Jonarw:
Nice, it certainly would be good to have an example like this!
Is there a specific reason why this can't be part of the ExampleLibrary?

PR #2089, opened 2024-07-20
fix #2061

### LINKED ISSUE #2061: DateTimeAxis.ToDateTime doesn't behave as intended in .NET 8
A `DateTimeAxis` with a step that required a reciprocating representation (e.g. 4hours = 1/6) results in unintended tick values.

image

I can't reproduce this with .NET Framework or .NET 6; observed it with .NET 7 also.

From the immediate window (across two debug sessions, one with .NET 7 and the other net framework 462):

```
DateTimeAxis.ToDateTime(45334.166666666664).ToString()
"12/02/2024 03:59:59"
DateTimeAxis.ToDateTime(45334.166666666664).ToString()
"12/02/2024 04:00:00"
```

I assume this is because of https://learn.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/7.0/datetime-add-precision where millisecond rounding was removed.

I don't have a suggestion for a fix this minute, but will try to work out a workaround soon for typical use-cases (probably just putting the implicit rounding back in).

Jonarw:
Fixes #2061 .

### Checklist

- Add `DateTimeAxis.DateTimePrecision` property, which is used to round `DateTime` values used by the axis. Default value is 1ms, so this should restore the behavior from .NET 6 and before.
- Add an example showcasing the issue
- Mark `DateTimeAxis.ToDateTime` as obsolete to make people aware of the issue

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot/Axes/DateTimeAxis.cs
+++ b/Source/OxyPlot/Axes/DateTimeAxis.cs
@@ -185,7 +185,7 @@ namespace OxyPlot.Axes
         }
 
         /// <summary>
-        /// Converts a numeric representation of the date (number of days after the time origin) to a DateTime structure, using the precision specified by DateTimePrecision.
+        /// Converts a numeric representation of the date (number of days after the time origin) to a DateTime structure, using the precision specified by <see cref="DateTimePrecision" />.
         /// </summary>
         /// <param name="value">The number of days after the time origin.</param>
         /// <returns>A <see cref="DateTime" /> structure. Ticks = 0 if the value is invalid.</returns>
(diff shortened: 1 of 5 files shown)


Source/OxyPlot/Axes/DateTimeAxis.cs
```diff
@@ -149,6 +184,16 @@ public static double ToDouble(DateTime value)
             return span.TotalDays + 1;
         }
 
+        /// <summary>
+        /// Converts a numeric representation of the date (number of days after the time origin) to a DateTime structure, using the precision specified by DateTimePrecision.
```
VisualMelon: `<see cref="DateTimePrecision" />`

VisualMelon (approved):
Thanks for picking this up!

VisualMelon (approved):

PR #2098, opened 2024-08-16
Refit ImageSharp again

VisualMelon:
(I need to rebase against develop and a few other things, so this is very much a WIP: don't bother reviewing it yet)

ImageSharp is out of beta, and the old libraries have listed vulnerabilities, so this bumps everything to the latest versions.

This requires targeting net6.0, so we lose netstandard and netframework support.

### Checklist

Just done a very quick refit so far, and doesn't look painful.

Outstanding issues:
 - Not tested properly yet, though the tests run and at least some of them produce recognisable figures
 - `DrawImage` doesn't work: may be possible to significantly simplify the method as there are some useful looking `DrawImage` overloads in ImageSharp now (though AFAIKT none that does the job in one)
    - Images render in the wrong place (e.g. PlotBackgroundInterpolated test)
    - I commented out the code that was responsible for nice edges on interpolated plots because the interface for changing individual pixels has gone away, so that's probably broken (maybe PlotBackgroundInterpolated again: dark edge suggests it's interpolating black: may be new graphics options to deal with this)
 - `DrawText` uses a hideous hard-coded path length because the old methods we used doesn't exist; may be that there is a nicer API for drawing text that doesn't involve laying out the glyphs etc. as is probably necessary for proper emoji support

@oxyplot/admins

### DIFF
--- a/Source/OxyPlot.ImageSharp.Tests/OxyPlot.ImageSharp.Tests.csproj
+++ b/Source/OxyPlot.ImageSharp.Tests/OxyPlot.ImageSharp.Tests.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
     <PropertyGroup>
-        <TargetFrameworks>net462;netcoreapp3.1</TargetFrameworks>
+        <TargetFrameworks>net6.0</TargetFrameworks>
         <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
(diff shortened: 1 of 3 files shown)


VisualMelon:
Looks like we don't need to go to 3 yet; can just bump the drawing package to 1.0.0. Opened https://github.com/oxyplot/oxyplot/pull/2099 ; won't scrub this branch because it might be useful later on

PR #2099, opened 2024-08-16
Bump ImageSharp.Drawing to 1.0.0

VisualMelon:
I was going to refit ImageSharp completely, but then I realised that ImageSharp.Drawing is fine with version 1.0.0 still, so this just bumps it so that it's not a beta version.

### Checklist

### Changes

 - Bump SixLabors.ImageSharp.Drawing dependencies to version 1.0.0
 - Add support for LineJoin in OxyPlot.ImageSharp/ImageRenderContext
 - Enable nullable in OxyPlot.ImageSharp

@oxyplot/admins


VisualMelon:
TODO:
 - import fixes from #2098 where applicable (zero thickness crash)
 - see if we can get line-join to work

VisualMelon:
TODO bits are done and tests all run; some examples are erroring but not sure if that's new.

VisualMelon:
Need to think about why `fontFamily` is nullable in `IRenderContext`.

VisualMelon:
Only errors that seem to be generated now are related to missing fonts

Source/OxyPlot.ImageSharp/ImageRenderContext.cs
```diff
@@ -254,9 +258,10 @@ public override void DrawText(ScreenPoint p, string text, OxyColor fill, string
         }
 
         /// <inheritdoc/>
-        public override OxySize MeasureText(string text, string fontFamily = null, double fontSize = 10, double fontWeight = 500)
+        public override OxySize MeasureText(string text, string? fontFamily = null, double fontSize = 10, double fontWeight = 500)
         {
-            return this.MeasureTextLoose(text, fontFamily, fontSize, fontWeight);
+            // TODO: should fontFamily be nullable?
```
Jonarw: Were you able to come to a conclusion on this?

Jonarw (commented):
Other than the one TODO, looks good to me!

Source/OxyPlot.ImageSharp/ImageRenderContext.cs
VisualMelon: The `RenderErrorMessage` logic leaves it `null`, which I presume is intentional to allow the render context to pick a simple font if there are rendering errors, or something like that. The example library is also full of `null` font families, so I don't think we can change it in a hurry. I did spot a bug, though, which I've lazily pushed as a small addition to this PR.

Jonarw (approved):

PR #2101, opened 2024-08-24
Update changelog for 2.2.0

VisualMelon:
Changes to the changelog for 2.2.0. Went with 2.2.0 rather than 2.1.3 because of all the dependency changes and the couple of breaking changes; nothing too much to be excited about in this release.

Once finalised, will create branch release/v2.2.0 and push to nuget.

Would be nice to include https://github.com/oxyplot/oxyplot/pull/2099 but can always do a point release.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
-## Unreleased
+## [2.2.0] - 2024-08-24
 
 ### Added
 - Example to demonstrate BarSeries HitTest bug (#2038)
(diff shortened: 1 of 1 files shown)


Jonarw (approved):

Jonarw:
Maybe we can try and get #2099 merged and then do the release?

VisualMelon:
That suits me. May be that we need to put out a point release to fix ImageSharp as it won't have been seriously tested, but that's OK with me.

VisualMelon:
@Jonarw if you're happy with this, I'll try to get the release done in the next few hours (running the tests can take a while these days)

VisualMelon:
Current develop pushed to https://www.myget.org/F/melonoxyplottesting as 2.2.0

VisualMelon:
Checked that the 2.2.0 WPF package works, and that 2.2.0 Core works with Avalonia 2.1.0-Avalonia11

Jonarw (approved):

VisualMelon:
Pushed

PR #2108, opened 2024-10-23
Add the colorblind friendly Okabe-Ito color palette

WimLeflere:
### Checklist

Add the colorblind friendly Okabe-Ito color palette.

Based on the paper: "How to make figures and presentations that are friendly to Colorblind people"
Masataka Okabe - Kei Ito (2008)
https://jfly.uni-koeln.de/color/

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## [2.3.0]
+
+### Added
+- Colorblind friendly Okabe-Ito color palette
+
 ## [2.2.0] - 2024-09-03
 
 ### Added
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -145,6 +145,7 @@ Udo Liess
 VisualMelon
 vhoehn <veit.hoehn@hte-company.de>
 Vsevolod Kukol <sevo@sevo.org>
+Wim Leflere <wim.leflere@gmail.com>
 Xavier <Xavier@xavier-PC.lsi>
 zur003 <Eric.Zurcher@csiro.au>
 Markus Ebner
--- a/Source/Examples/ExampleLibrary/Misc/PaletteExamples.cs
+++ b/Source/Examples/ExampleLibrary/Misc/PaletteExamples.cs
@@ -7,6 +7,7 @@
 namespace ExampleLibrary
 {
     using OxyPlot;
+    using OxyPlot.Axes;
 
     [Examples("Palettes")]
     public class PaletteExamples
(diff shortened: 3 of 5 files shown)


WimLeflere:
It would be nice to set this palette as the `DefaultColors` for `PlotModel`.
As the default colors (especially the first two) are quite hard to distinguish for colorblind people.

VisualMelon:
Thanks for opening this; I'll take a proper look at the weekend (ping me if I forget). We should also consider an alternative system of default styles (colours and markers and dashes etc.) to make it easier to make colorblind friendly plots (and black-and-white printable plots).

WimLeflere:
I started from this answer on StackOverflow for colorblind accessible colors in R. It has some interesting links to blogs and papers. https://stackoverflow.com/questions/57153428/r-plot-color-combinations-that-are-colorblind-accessible

Regarding the code change, should the property be renamed to `OkabeIto8` to indicate the number of colors (similar to other properties)? Or should it be changed to a factory function `public static OxyPalette OkabeIto(int numberOfColors)`?

VisualMelon:
The second is probably preferable, then may or may not add one with a pre-set number of colours. The Cividis provision is a good example to follow:

https://github.com/oxyplot/oxyplot/blob/989df42a75cfc1ef5751afe3f9eb859bc9e62ded/Source/OxyPlot/Rendering/OxyPalettes.MatPlotLib.cs#L1379

WimLeflere:
@VisualMelon have you had time to look at it?

VisualMelon:
Thanks for the reminder, will try to do so today.

Jonarw:
Thanks for this PR, I think this would be a very valuable addition!

As far as I understand, the Okabe-Ito color palette is meant to visualize distinct categories, rather than continuous data. In fact, for continuous data, we do have some color palettes which are supposed to be color-blind-friendly (Viridis is the default, and we also have Cividis).

I think the peaks-based examples are more suited for continuous color palettes. For a category-based color palette like this, maybe a BarSeries-based example would be more appropriate?

I am also not sure if Okabe-Ito would be the best color palette to use as the default palette. The first color of Okabe-Ito is just plain black, which I think might not be ideal whether you are color-blind or not.
Keep in mind that I am not an expert on this topic by any means, so I could be convinced otherwise.

VisualMelon:
@Jonarw 
> I am also not sure if Okabe-Ito would be the best color palette to use as the default palette

Completely agree, meant to suggest something a bit more ambitious (usual thing of me dumping ideas in posts as they occur to me rather than opening a new thread): it would be nice for there to be a way to change the default colours/markers for a plot more easily, e.g. so that you can swap out the existing defaults for printer/colour-blind friendly defaults. It looks like it's actually already easy to change the `DefaultColors`, but `PlotModel.GetDefaultLineStyle` is tied to the number of default colors, and there's no automatic assignment of markers. Would be nice if it was easy to get redundant coding (as described in https://jfly.uni-koeln.de/color/ linked by @WimLeflere) without having configure individual series.

@WimLeflere does it make sense for the colours to be interpolable? If so, should probably have an API more like Cividis/BlackWhiteRed (taking a number-of-colours parmeter); if not, I think it's fine as is, but I'd agree with @Jonarw that something with different colours between series would make more sense as an example, e.g. a set of line series with the `DefaultColors` on the PlotModel changed to those of the palette.

WimLeflere:
I've updated the example to use line series.
Because the default line series palette (especially the first two colors) is hard to distinguish for color-blind people.

Interpolation works OK, but I prefer the default behavior of the different line types.

PR #2123, opened 2025-01-24
Allow loading typefaces into SkiaRenderContext manually

### LINKED ISSUE #2118: Only monospaced fonts available in OxyPlot.SkiaSharp on WebAssembly
Platform: SkiaSharp / WebAssembly
.NET version: .NET 9

### Expected behaviour

Setting `DefaultFont` on `PlotModel` should have an effect. 

### Actual behaviour

The font never changes, it is always a monospaced font, which would seem to be a sort of fallback default in OxyPlot. For example, given this model:

    var model = new PlotModel
    {
        Title = "Cake Type Popularity",
        DefaultFont = "Arial"
    };
    ... code to add series etc...

this is the result:

image

I am using [this library](https://github.com/JensKrumsieck/OxyPlot.SkiaSharp.Blazor) in Blazor WebAssembly project, which I upgraded to .NET 9. Given [this code](https://github.com/oxyplot/oxyplot/blob/989df42a75cfc1ef5751afe3f9eb859bc9e62ded/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs#L871) inside of OxyPlot, I am wondering if this is a sort of "design choice"? 

Using different fonts should be possible, I am using the Blazor `SKCanvasView` (which is used here as well to display the plot) elsewhere also, and can load system fonts without issue. Maybe embedding a font could help, but that introduces all kinds of licensing questions I would rather avoid...

Alas, is there a reason and/or potential fix for this behaviour of OxyPlot.SkiaSharp in a WASM context?

[Edit] Actually, I am also only seeing the "Noto Mono" font in the other app where I use `SKCanvasView`.

mennowo:
Fixes #2118, by allowing loading fonts embedded in an arbitrary external library to be explicitly loaded into `SkiaRenderContext`.

### Checklist

- Add a new public method `AddTypeface` to `SkiaRenderContext` to allow manually loading fonts

### Remarks

There is no test, since I am unsure how to create one. I got started, but most of the logic/variables used is private to `SkiaRenderContext` and therefore not accessible from a test.

I rebased, but I am unsure how to clean up the commits that are in this PR now. I am happy to do that if someone can provide me with some pointer on how to proceed, or maybe it can just be squash-merged?

After this has been merged, I will also update [this repo](https://github.com/CodingConnected/OxyPlot.SkiaSharp.Blazor), so that OxyPlot is fully usable in a WebAssembly context. In case of interest, I am having great results with the altered version of OxyPlot in the browser; it looks nice, it's reactive, it's fast.

I'd be happy to merge that solution into oxyplot as well, but maybe it creates too much extra dependencies; let me know if there is interest in that regard. I might also create a nuget package for the OxyPlot.SkiaSharp.Blazor if I find the time.

@oxyplot/admins

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Example to demonstrate how to create vertical BarSeries
+- Add a public method `AddTypeface` to `SkiaRenderContext` to allow manually adding typefaces to the cache
 
 ## [2.2.0] - 2024-09-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
-## [2.2.1] - 2024-12-11
+## Unreleased
 
 ### Added
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Change Log
 All notable changes to this project will be documented in this file.
 
+## [2.2.1] - 2024-12-11
+
+### Added
+
+- Example to demonstrate how to create vertical BarSeries
+
 ## [2.2.0] - 2024-09-03
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 
 - Example to demonstrate how to create vertical BarSeries
 - Add a public method `AddTypeface` to `SkiaRenderContext` to allow manually adding typefaces to the cache
+- Add a virtual method `GetTypeface` to `SkiaRenderContext` to allow overriding retrieving typefaces
 
 ## [2.2.0] - 2024-09-03
 
--- a/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
+++ b/Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
@@ -476,6 +476,29 @@ namespace OxyPlot.SkiaSharp
         }
 
         /// <summary>
+        /// Manually add a typeface to the context typeface cache
+        /// This may be needed for example in a WebAssembly context, where system fonts cannot be accessed. 
+        /// In such a case, fonts must be embedded in the app or library. This method allows loading fonts
+        /// from an arbitrary external lib, and enables caching of loaded fonts locally, in case multiple 
+        /// instances of SkiaRenderContext are instatiated.
+        /// </summary>
+        /// <param name="fontFamily">The font family name</param>
+        /// <param name="fontWeight">The weight of the font (eg.: 400 for Regular, 700 for Bold; these are the only ones OxyPlot actually uses)</param>
+        /// <param name="typeface">Typeface instance to be added to cache, typically loaded from some embedded resource, eg.: <code>var typeface = SKTypeface.FromStream(stream);</code></param>
+        /// <returns>True if succesful, false if typeface with this family name and weight was already in cache</returns>
+        public bool AddTypeface(string fontFamily, double fontWeight, SKTypeface typeface)
+        {
+            var fontDescriptor = new FontDescriptor(fontFamily, fontWeight);
+            if (!this.typefaceCache.ContainsKey(fontDescriptor))
+            {
+                this.typefaceCache[fontDescriptor] = typeface;
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
         /// Disposes managed resources.
         /// </summary>
         /// <param name="disposing">A value indicating whether this method is called from the Dispose method.</param>
@@ -867,7 +890,7 @@ namespace OxyPlot.SkiaSharp
             if (!this.typefaceCache.TryGetValue(fontDescriptor, out var typeface))
             {
                 typeface = SKTypeface.FromFamilyName(fontFamily, new SKFontStyle((int)fontWeight, (int)SKFontStyleWidth.Normal, SKFontStyleSlant.Upright));
-#if NETSTANDARD2_0_OR_GREATER
+
                 if (typeface.FamilyName != fontFamily) // requested font not found or is WASM
                 {
                     try
@@ -890,7 +913,6 @@ namespace OxyPlot.SkiaSharp
                         Debug.WriteLine($"Requested Font {fontFamily} could not be found, falling back to {typeface.FamilyName}");
                     }
                 }
-#endif
                 this.typefaceCache.Add(fontDescriptor, typeface);
             }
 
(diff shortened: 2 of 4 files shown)


VisualMelon:
@Jonarw are you able to give this a bit of time? No serious precedent for the API but looks OK to me (PortableDocument.AddFont is the only thing that's obviously similar)

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -475,6 +475,29 @@ public void SetToolTip(string text)
...
+        /// This may be needed for example in a WebAssembly context, where system fonts cannot be accessed. 
+        /// In such a case, fonts must be embedded in the app or library. This method allows loading fonts
+        /// from an arbitrary external lib, and enables caching of loaded fonts locally, in case multiple 
+        /// instances of SkiaRenderContext are instatiated.
+        /// </summary>
+        /// <param name="fontFamily">The font family name</param>
+        /// <param name="fontWeight">The weight of the font (eg.: 400 for Regular, 700 for Bold; these are the only ones OxyPlot actually uses)</param>
+        /// <param name="typeface">Typeface instance to be added to cache, typically loaded from some embedded resource, eg.: <code>var typeface = SKTypeface.FromStream(stream);</code></param>
+        /// <returns>True if succesful, false if typeface with this family name and weight was already in cache</returns>
+        public bool AddTypeface(string fontFamily, double fontWeight, SKTypeface typeface)
```
Jonarw: This API does not allow replacing a previously added typeface with a new one. Is this a problem?

Maybe the function should always update the `typefaceCache` with the new `typeface`, and the return value should reflect if the typeface was added or if it already existed - what do you think?

Jonarw (commented):
I think this is a good idea, nice that there is a simple way to make this work on WASM. See my one comment regarding API.

Regarding squashing / rebasing: A `git rebase origin/develop` should usually do the trick. But in this case I think a squash-merge should also do the trick, so no need to do anything on your end.

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
VisualMelon: Short of adding add/set/remove methods or hiding this behind some wrapper collection type, I'm not sure what would make sense.

Looking at this more closely - and thinking about how we might want to do this in other platforms - I don't think working with the `typefaceCache` makes sense either. I'd suggest we move the logic for looking up a type face (in `GetTextPaint` at the moment) into a small virtual method so that it has an extension point, and consider adding a separate dictionary of 'override' lookups, which - where available - are used instead of querying `SKTypeface.FromFamilyName`.

Extension point is worth it because a user may want to support a range of custom fonts, and if e.g. WASM needed it's own weird lookup mechanism then it could be implemented once and forgotten about. Dictionary is worth it to cover symbol cases, and it can just be an `IDictionary<FontDescriptor, SkTypeface>` if we want to keep it simple and not think about the API too hard.

mennowo: Ok I'm happy to give this a go. This would mean making `FontDescriptor` public, which at the moment it is not.

One question: why would one need a seperate dictionary of 'override' lookups? To allow removing an added typeface later? Or some other use case? I would think typically one would load a font and expect it to be used to render a plot; wether or not it overrides something else is kind of irrelevant. Is there a use case for wanting the exact same `FontDescriptor` to later be related to a _different_ typeface? From a performance perspective, I would want only fonts I need to be loaded and only once (which is also why I cache them in WASM, where I have multiple simultaneous instances of the `PlotView` component, and therefor of `SkiaRenderContext`).

VisualMelon: It's an override because it will preclude looking up the font separately, but as you say that's not the main use-case (but we do have to define how it will behave if there is an existing font etc.).

Using the `typefaceCache` feels dodgy because it's intended as a cache, and currently disposes its elements when the `SkiaRenderContext` is lost: if OxyPlot didn't create the instance then we need to decide whether it should be disposing it (indeed, may be a problem for your use-case, but I don't know enough about SkiaSharp to really comment).

Personally I'd have no issue with making FontDescriptor public, but I could imagine that we might want a common FontDescriptor type for other platforms long term. That said, I'm not wedded to the idea of exposing an `IDictionary'2`: was really just throwing a simple idea out in hopes of some conversation and I'll generally defer to Jon on any Skia matters anyway.

mennowo: Ok, yes, the fact that `SkiaRenderContext` will call `Dispose()` on elements in the cache may present an issue. I will try to find some time the coming 2 weeks to have a go at this.

mennowo: @VisualMelon Looking into the now, I wonder why we need "a separate dictionary of 'override' lookups". Doesn't having an extenson point already solve most use cases? It allows writing a class that derives from `SkiaRenderContext` and then override `public virtual SKTypeface GetTypeface(FontDescriptor fontDescriptor)` (which I now implemented). Then, one can add any logic, caching, etc., to that derived class.

If there is a use case for having some sort of override lookup mechanism, I'm happy to look into it, but as it stands I don't really see the benefit, neither understand exactly what the requirements/specification for this would be.

mennowo: Shall I remove the previous method `AddTypeface`?

Jonarw: From my point of view it should be enough to have the virtual `SKTypeface GetTypeface(FontDescriptor fontDescriptor)`. As you say, derived classes can do their own caching as necessary.
I think the `bool AddTypeface(string fontFamily, double fontWeight, SKTypeface typeface)` should be removed then.

Source/OxyPlot.SkiaSharp/SkiaRenderContext.cs
```diff
@@ -11,6 +11,7 @@ namespace OxyPlot.SkiaSharp
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Drawing;
```
Jonarw: I don't think this is needed?

mennowo: Ok I just commited these changes. Could it be merged like this? It would be nice if this makes it into a new release so I can switch back to using the nuget package instead of custom built dll. However, there is no rush. Also, I'm happy to add additional feedback if more changes are needed.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ All notable changes to this project will be documented in this file.
 ### Added
 
 - Example to demonstrate how to create vertical BarSeries
+- Add a public method `AddTypeface` to `SkiaRenderContext` to allow manually adding typefaces to the cache
```
Jonarw, line 9: I think this line is obsolete now, otherwise this looks good to me now!

VisualMelon (approved):
I'm certainly happy with the changes as they are

CHANGELOG.md
mennowo: true, I can remove it later this week.

mennowo:
Ok nice, I can correct the change log later this week, and then it can be merged

PR #2129, opened 2025-02-15
Fix DateTimeAxis converting local time rather than UTC to TimeZone

carlreinke:
Fixes problem introduced by #2089 which caused the `DateTime` being converted to lose its `Kind`. 

### Checklist

- Fix regression introduced in v2.2.0

@oxyplot/admins


VisualMelon (approved):
Thanks for spotting and fixing this.

I'm not terribly familiar with the logic behind this, but behaviour looks good and changes make sense to me.

Jonarw (approved):

